PR #1, opened 2016-09-14
Make INSERT INTO a toplevel word and allow longer inline parentheses

ukupat:


src/core/InlineBlock.js
```diff
@@ -1,6 +1,6 @@
 import sqlTokenTypes from "./tokenTypes";
 
-const INLINE_MAX_LENGTH = 35;
+const INLINE_MAX_LENGTH = 50;
```
andrestaht: I think we should also have tests for values that are shorter than 50 will stay on single line and longer ones get split onto multiple lines.

ukupat: Thats a good idea, we already have them: `formats simple INSERT` query and `breaks long parenthized lists to multiple lines` in `behavesLikeSqlFormatter.js`

PR #8, opened 2016-09-20
Avoid usage of babel-polyfill

bsideup:
See https://babeljs.io/docs/usage/polyfill/

> If you are looking for something that won't modify globals to be used in a tool/library, checkout the transform-runtime plugin. This means you won't be able to use the instance methods mentioned above like Array.includes.


ukupat:
Thanks a lot! I'll deal with the failing build.

Issue #4, opened 2016-09-15
Replace Jasmine+Karma with Mocha+Chai


nene:
This whole setup for tests feels way too complicated:

- There's little point of using Karma when our tests don't need to run in Browser environment e.g. to interact with DOM.
- Jasmine too is written with browser-testing primarily in mind.
- This whole issue with only one `require "babel-polyfill"` is a nuisance.

With Mocha one would not need to write any complex configuration files or integrate with Webpack to support ES6. One could use a single command like:

```
$ mocha --compilers js:babel-register "test/**/*Test.js"
```

With Chain there's minor difference in the format of assertions, but that mostly means replacing `toEqual()` with `.to.equal()`.

PS. One thing I don't know about is integration with code coverage - but I suppose it should not be too difficult either.

bsideup:
I highly recommend you to take a look at Jest, i.e. check how we use it in https://github.com/bsideup/rx-connect .

It has built-in support for babel, coverage, watching, "test only changed", it's browserless, etc...

PR #9, opened 2016-09-20
#4 Replace Karma+Jasmine with Jest test framework

ukupat:
Related to issue #4


bsideup:
@ukupat Jest has really nice feature - snapshot testing. i.e. `expect(result).toMatchSnapshot()`. So you don't have to have this huge result strings in the test's code. Maybe we can use it as well?

PR #14, opened 2016-11-02
Placeholder replacement

ukupat:
I created two different placeholder token regexes (based on if placeholder has a name or not), but there is only one token type: PLACEHOLDER.

I haven't updated README.md yet, lets see what is your opinion on current solution.

Info about N1QL parameterized queries http://developer.couchbase.com/documentation/server/current/sdk/n1ql-query.html

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,29 @@ Currently just two SQL dialects are supported:
 - **sql** - [Standard SQL][]
 - **n1ql** - [Couchbase N1QL][]
 
+### Placeholders replacement
+
+```js
+// Named placeholders
+sqlFormatter.format("SELECT @foo FROM table1", {
+    params: {foo: "bar"}
+}));
+
+// Indexed placeholders
+sqlFormatter.format("SELECT ? FROM table1", {
+    params: ["bar"]
+}));
+```
+
+Both result in:
+
+```
+SELECT
+  bar
+FROM
+  table1
+```
+
 ## Usage without NPM
 
 If you don't use a module bundler, clone the repository, run `npm install` and grab a file from `/dist` directory to use inside a `<script>` tag.
--- a/README.md
+++ b/README.md
@@ -49,13 +49,13 @@ Currently just two SQL dialects are supported:
 
 ```js
 // Named placeholders
-sqlFormatter.format("SELECT @foo FROM table1", {
-    params: {foo: "bar"}
+sqlFormatter.format("SELECT * FROM tbl WHERE foo = @foo", {
+    params: {foo: "'bar'"}
 }));
 
 // Indexed placeholders
-sqlFormatter.format("SELECT ? FROM table1", {
-    params: ["bar"]
+sqlFormatter.format("SELECT * FROM tbl WHERE foo = ?", {
+    params: ["'bar'"]
 }));
 ```
 
@@ -63,9 +63,11 @@ Both result in:
 
 ```
 SELECT
-  bar
+  *
 FROM
-  table1
+  tbl
+WHERE
+  foo = 'bar'
 ```
 
 ## Usage without NPM
--- a/src/core/Params.js
+++ b/src/core/Params.js
@@ -1,0 +1,29 @@
+/**
+ * Handles placeholder replacement with given params.
+ */
+export default class Params {
+    /**
+     * @param {Object} params
+     */
+    constructor(params) {
+        this.params = params;
+        this.index = 0;
+    }
+
+    /**
+     * Returns param value that matches given placeholder with param key.
+     * @param {String} placeholder
+     * @return {String} param
+     */
+    get(placeholder) {
+        if (!this.params) {
+            return placeholder;
+        }
+        const matches = placeholder.match(/\w+(\s+\w+)*/);
+
+        if (matches) {
+            return this.params[matches[0]];
+        }
+        return this.params[this.index ++];
+    }
+}
--- a/src/core/Params.js
+++ b/src/core/Params.js
@@ -1,3 +1,5 @@
+import tokenTypes from "./tokenTypes";
+
 /**
  * Handles placeholder replacement with given params.
  */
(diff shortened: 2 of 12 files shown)


src/core/Tokenizer.js
```diff
@@ -29,8 +30,12 @@ export default class Tokenizer {
         this.OPEN_PAREN_REGEX = this.createParenRegex(cfg.openParens);
         this.CLOSE_PAREN_REGEX = this.createParenRegex(cfg.closeParens);
 
-        this.PLAIN_VARIABLE_REGEX = this.createVariableRegex(cfg.variableTypes, "[a-zA-Z0-9._$]+");
-        this.QUOTED_VARIABLE_REGEX = this.createVariableRegex(cfg.variableTypes, this.createStringPattern(cfg.stringTypes));
+        this.INDEXED_PLACEHOLDER_REGEX = this.createPlaceholderRegex(cfg.indexedPlaceholderTypes);
+        this.PLAIN_NAMED_PLACEHOLDER_REGEX = this.createPlaceholderRegex(cfg.namedPlaceholderTypes, "[a-zA-Z0-9._$]+");
+        this.STRING_NAMED_PLACEHOLDER_REGEX = this.createPlaceholderRegex(
```
nene: I don't really like the name `PLAIN_NAMED`. Perhaps better `IDENT_NAMED` to emphasize that the name is an identifier.

Speaking of identifiers though, the regex `[a-zA-Z0-9._$]+` isn't fully covered by tests - we don't test for the special characters `.`, `_` and `$`. I wonder, do we even really need to support `.` and `$`.

src/core/Tokenizer.js
```diff
@@ -165,23 +173,33 @@ export default class Tokenizer {
...
-    getQuotedVariableToken(input) {
+    getIndexedPlaceholderToken(input) {
         return this.getTokenOnFirstMatch({
             input,
-            type: sqlTokenTypes.VARIABLE,
-            regex: this.QUOTED_VARIABLE_REGEX
+            type: sqlTokenTypes.PLACEHOLDER,
+            regex: this.INDEXED_PLACEHOLDER_REGEX
         });
     }
```
nene: We could generalize these three methods - the only difference between them is the regex.

test/N1qlFormatterTest.js
```diff
@@ -132,4 +132,54 @@ describe("N1qlFormatter", function() {
...
+                2: "third"
+            }
+        });
+        expect(result).toBe(
+            "SELECT\n" +
+            "  second,\n" +
+            "  third,\n" +
+            "  first;\n"
+        );
+    });
```
nene: I think we should order other tests also similarly, so that all tests for `@variables` are grouped together like the above tests for `$variables`.

test/StandardSqlFormatterTest.js
```diff
@@ -60,4 +60,94 @@ describe("StandardSqlFormatter", function() {
...
+    });
+
+    it("replaces ? indexed placeholders with param values", function() {
+        const result = sqlFormatter.format("SELECT ?, ?, ?;", {
+            language: "sql",
+            params: {
+                0: "first",
+                1: "second",
+                2: "third"
+            }
```
nene: I think we should also have a test that uses plain array instead of object with numeric keys.

src/core/Params.js
```diff
@@ -0,0 +1,29 @@
...
+    /**
+     * Returns param value that matches given placeholder with param key.
+     * @param {String} placeholder
+     * @return {String} param
+     */
+    get(placeholder) {
+        if (!this.params) {
+            return placeholder;
+        }
+        const matches = placeholder.match(/\w+(\s+\w+)*/);
```
nene: I don't really understand why this regex is the way it is.

I understand that the goal is to detect whether it's a named placeholder. However it doesn't quite work - for example when I have a placeholder that does not consist of just spaces and word characters:

    SELECT @'$^&*'

It might be simpler to differentiate between named and indexed placeholders inside tokenizer.

Another thing to consider is escape sequences inside strings:

    SELECT @"hello \" world"

src/core/Tokenizer.js
ukupat: Not sure if we want to do it now when we have two different placeholder types. WDYT @nene ?

ukupat:
@nene covered your requested changes.

src/core/Params.js
```diff
@@ -12,17 +14,21 @@ export default class Params {
...
         }
-        const matches = placeholder.match(/\w+(\s+\w+)*/);
+        const key = type === tokenTypes.PLACEHOLDER_STRING ?
+            value.substring(2).slice(0, -1) :
+            value.substring(1);
 
-        if (matches) {
-            return this.params[matches[0]];
+        if (key) {
+            return this.params[key];
```
nene: We really have three placeholder token types: STRING, INDENT, INDEXED. Currently the INDEXED and INDENT are thrown together (although they're really quite different). By having three distinct types, we could express the three different choices in much clearer way:

```js
switch(type) {
    case INDEXED:
    case STRING:
    case IDENT:
}
```

Also, I think parsing the contents of the string and indent placeholders should really belong to Tokenizer. The Formatter should not care about the specific syntax that placeholders have. It's also common practice for Lexer to provide both a raw value and some computed value of a token like string. In that scenario we actually won't need 3 token types as STRING and INDENT can be represented with one.

Finally I still think that we should take at least some minimal care about escaping. As the Lexer allows escape sequences like `\"` in strings, the parameter replacement should also support them.

src/core/Tokenizer.js
nene: Added some other suggestion which might need a different generalization of the same thing.

ukupat:
I did my best to resolve @nene requested changes.

test/StandardSqlFormatterTest.js
ukupat: @nene added one, do we need more or does it satisfy us now?

src/core/Tokenizer.js
```diff
@@ -208,6 +208,10 @@ export default class Tokenizer {
         return token;
     }
 
+    getEscapedPlaceholderKey({key, stringChar}) {
```
nene: I think a better name would be: `quoteChar`

README.md
```diff
@@ -45,6 +45,29 @@ Currently just two SQL dialects are supported:
...
+// Named placeholders
+sqlFormatter.format("SELECT @foo FROM table1", {
+    params: {foo: "bar"}
+}));
+
+// Indexed placeholders
+sqlFormatter.format("SELECT ? FROM table1", {
+    params: ["bar"]
+}));
+```
```
nene: We should use a more real-life example. One normally uses placeholders in conditions of WHERE clause like:

```
SELECT * FROM tbl WHERE foo = ?
```

PR #18, opened 2017-02-24
Add TSQL (SQL Server) keywords and tests

mathsalmi:
This PR adds support for the following keywords: GO, CROSS JOIN, CROSS APPLY, and OUTER APPLY.


ukupat:
Nice work, @mathsalmi! 

At first me and @nene were thinking if TSQL needs to be separated from StandardSql logic. Fortunately we came to a conclusion that these new keywords should not cause any problems.

mathsalmi:
Nice! First I started in a different file but then when I was writing the unit tests I got to the conclusion I was doing the same work again. Anyway thanks! =D

ukupat:
@mathsalmi released new version with your changes https://github.com/zeroturnaround/sql-formatter/releases/tag/v1.2.0. Enjoy!

PR #21, opened 2017-05-15
Better formatting for MSSQL when using National Characters

mtxr:
formatting queries with `N'value here'`.

### DIFF
--- a/src/core/Formatter.js
+++ b/src/core/Formatter.js
@@ -73,6 +73,9 @@ export default class Formatter {
             else if (token.value === ":") {
                 formattedQuery = this.formatWithSpaceAfter(token, formattedQuery);
             }
+            else if (token.value === "N") {
+                formattedQuery = this.formatWithSpaceBefore(token, formattedQuery);
+            }
             else if (token.value === "." || token.value === ";") {
                 formattedQuery = this.formatWithoutSpaces(token, formattedQuery);
             }
@@ -164,8 +167,13 @@ export default class Formatter {
         }
     }
 
+    formatWithSpaceBefore(token, query) {
+        return _.trimEnd(query) + " " + token.value;
+
+    }
+
     formatWithSpaceAfter(token, query) {
-        return _.trimEnd(query) + token.value + " ";
+        return this.formatWithoutSpaces(token, _.trimEnd(query)) + " ";
     }
 
     formatWithoutSpaces(token, query) {
--- a/src/core/Formatter.js
+++ b/src/core/Formatter.js
@@ -165,7 +165,7 @@ export default class Formatter {
     }
 
     formatWithSpaceAfter(token, query) {
-        return this.formatWithoutSpaces(token, _.trimEnd(query)) + " ";
+        return _.trimEnd(query) + token.value + " ";
     }
 
     formatWithoutSpaces(token, query) {
--- a/src/core/Formatter.js
+++ b/src/core/Formatter.js
@@ -164,11 +164,6 @@ export default class Formatter {
         }
     }
 
-    formatWithSpaceBefore(token, query) {
-        return _.trimEnd(query) + " " + token.value;
-
-    }
-
     formatWithSpaceAfter(token, query) {
         return this.formatWithoutSpaces(token, _.trimEnd(query)) + " ";
     }
--- a/src/core/Formatter.js
+++ b/src/core/Formatter.js
@@ -73,9 +73,6 @@ export default class Formatter {
             else if (token.value === ":") {
                 formattedQuery = this.formatWithSpaceAfter(token, formattedQuery);
             }
-            else if (token.value === "N") {
-                formattedQuery = this.formatWithSpaceBefore(token, formattedQuery);
-            }
             else if (token.value === "." || token.value === ";") {
                 formattedQuery = this.formatWithoutSpaces(token, formattedQuery);
             }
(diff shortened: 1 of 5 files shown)


nene:
So, this handles the `N'string'`, though when you format a query like:

```
SELECT N'some string';
```

You'd expect a formatting like:

```
SELECT
  N'some string';
```

But instead the formatter now produces:

```
SELECT N'some string';
```

Also, when given a query like:

```
SELECT N, M FROM MY_TABLE;
```

You'd expect a result like:

```
SELECT
  N,
  M
FROM
  MY_TABLE;
```

But instead we get:

```
SELECT N,
  M
FROM
  MY_TABLE;
```

I think it would be better to handle this special string syntax inside Tokenizer, where all other strings are detected. Also, all other string syntaxes are configurable, so should be this. So we could configure strings like:

```
stringTypes: ["''", "N''"]
```

test/StandardSqlFormatterTest.js
```diff
@@ -202,6 +202,67 @@ describe("StandardSqlFormatter", function() {
...
 
+    it("formats simple SELECT", function() {
+        const result = sqlFormatter.format("SELECT 'value'", );
+        expect(result).toBe(
+            "SELECT\n" +
+            "  'value'\n"
+        );
+    });
+
+    it("formats simple SELECT", function() {
```
ukupat: It has a duplicate test description. Could we be more specific about why is this test different from the above one?

src/core/Tokenizer.js
```diff
@@ -54,10 +54,12 @@ export default class Tokenizer {
     // 2. square bracket quoted string (SQL Server) using ]] to escape
     // 3. double quoted string using "" or \" to escape
     // 4. single quoted string using '' or \' to escape
+    // 5. national character single quoted string using N'' or \' to escape
```
ukupat: Great comment! `constructor` `cfg.stringTypes` comment needs also some updating.

src/core/Formatter.js
```diff
@@ -165,7 +165,7 @@ export default class Formatter {
     }
 
     formatWithSpaceAfter(token, query) {
-        return _.trimEnd(query) + token.value + " ";
+        return this.formatWithoutSpaces(token, _.trimEnd(query)) + " ";
```
ukupat: Why did we change this? If we want to use `formatWithoutSpaces` to reduce duplication then we could use it like this: 

```
return this. formatWithoutSpaces(token, query) + " ";
```

 But I think it was easier to understand `formatWithSpaceAfter` logic when there was a bit of duplication in the code.

ukupat:
@mtxr – waiting for your changes.

mtxr:
Sorry. I missed this notification, I'll do this by the end of the day

test/StandardSqlFormatterTest.js
mtxr: Fixed on mtxr/sql-formatter@d6a294dc8b1c5e2a3c1b2e5f93488b265e9de78e

src/core/Tokenizer.js
mtxr: Thanks!

src/core/Formatter.js
mtxr: Fixed on mtxr/sql-formatter@d6a294dc8b1c5e2a3c1b2e5f93488b265e9de78e

test/behavesLikeSqlFormatter.js
```diff
@@ -8,7 +8,7 @@ export default function behavesLikeSqlFormatter(language) {
     it("uses given indent config for indention", function() {
         const result = sqlFormatter.format(
             "SELECT count(*),Column1 FROM Table1;",
-            {language, indent: "    "}
+            { language, indent: "    " }
```
ukupat: This change is not needed.

test/StandardSqlFormatterTest.js
```diff
@@ -202,6 +202,66 @@ describe("StandardSqlFormatter", function() {
         );
     });
 
+    it("formats simple SELECT", function() {
```
ukupat: This case is already covered by other tests, therefore we can remove it.

test/StandardSqlFormatterTest.js
```diff
@@ -202,6 +202,66 @@ describe("StandardSqlFormatter", function() {
...
+        const result = sqlFormatter.format(
+            "SELECT N'value';"
+        );
+        expect(result).toBe(
+            "SELECT\n" +
+            "  N'value';\n"
+        );
+    });
+
+    it("formats SELECT with complex WHERE", function() {
```
ukupat: This test file is getting quite huge. We could save some lines when merge this test and below one:

```
it("formats SELECT with complex WHERE", function() {
    const result = sqlFormatter.format(
        "SELECT * FROM foo WHERE Column1 = N'testing' AND Column2 = 'testing'" +
        "AND ( (Column3 = Column4 OR Column5 >= NOW()) );"
    );
    expect(result).toBe(
        "SELECT\n" +
        "  *\n" +
        "FROM\n" +
        "  foo\n" +
        "WHERE\n" +
        "  Column1 = N'testing'\n" +
        "  AND (\n" +
        "  Column2 = 'testing'\n" +
        "    (\n" +
        "      Column3 = Column4\n" +
        "      OR Column5 >= NOW()\n" +
        "    )\n" +
        "  );\n"
    );
});
```

src/core/Tokenizer.js
ukupat: I think you forgot to update `Tokenizer` `constructor` `cfg.stringTypes` comment.

mtxr: `5. national character quoted string using N'' or N\' to escape`

Is it ok?

ukupat: That is cool, but line 10 needs update as well :)

ukupat:
@mtxr – here you go https://github.com/zeroturnaround/sql-formatter/releases/tag/v1.2.2

PR #22, opened 2017-05-25
Uppercase reserved words = when passed option

whatyoubendoing:
- Allow option to uppercase reserved words

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,7 @@ You can also pass in configuration options:
 sqlFormatter.format("SELECT *", {
     language: "n1ql" // Defaults to "sql"
     indent: "    "   // Defaults to two spaces
+    uppercase: true  // Defaultd to false
 });
 ```
 
--- a/src/core/Formatter.js
+++ b/src/core/Formatter.js
@@ -102,12 +102,12 @@ export default class Formatter {
 
         this.indentation.increaseToplevel();
 
-        query += this.equalizeWhitespace(token.value);
+        query += this.equalizeWhitespace(this.formatReservedWords(token.value));
         return this.addNewline(query);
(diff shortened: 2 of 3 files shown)


danwinterwijntjes:
the most useful

nene:
Hello, thanks for the PR.

The problem with converting all reserved words to uppercase is that the list of reserved words that sql-formatter uses contains some words that are simply significant (like some names of built-in functions) but not really reserved. When we try to reduce this list to list of really reserved words, a question becomes: reserved words of which SQL dialect?

For example the following query is valid in MySQL:

```
SELECT id AS auto_increment FROM my_table
```

But `auto_increment` happens to be listed in `reservedWords` and would be therefore uppercased.

So, really the main problem in implementing this feature is compiling a solid set of reserved words that would really be reserved in all dialects (or compiling separate lists for specific dialects).

Therefore in its current form we can't accept this pull request.

whatyoubendoing:
@nene & @ukupat. would you be happy if i created a seperate reserved words list, just for capitalisation?

nene:
That's fine by me. We could split the current reserved words list to two:
one that can be capitalized, and other that can't. Just need to take care
the capitalizable list applies for all SQL dialects.


29.05.2017 5:48 AM kirjutas kuupäeval "Benjamin Bellamy" <
notifications@github.com>:

navxio:
guys, can we move this ahead? Can I help?

nene:
This enhancement is now part of larger PR: #112

I guess it solves the majority wish of having a simple upper-casing of keywords, although this solution is not fully safe.

nene:
Now released as part of 3.1.0

PR #24, opened 2017-06-13
Adding Postgres specific operators

sasha-alias:
Postgresql has several specific operators:
'::' - cast operator
'->' - get JSON element
'->>' - get JSON element as text
This commit adds them to the list of standard operators in order to not
break them when formatting.

https://github.com/zeroturnaround/sql-formatter/issues/23

### DIFF
--- a/test/behavesLikeSqlFormatter.js
+++ b/test/behavesLikeSqlFormatter.js
@@ -432,4 +432,10 @@ export default function behavesLikeSqlFormatter(language) {
         expect(format("'foo \\' JOIN bar'")).toBe("'foo \\' JOIN bar'\n");
         expect(format("`foo `` JOIN bar`")).toBe("`foo `` JOIN bar`\n");
     });
+
+    it("formats postgres specific operators", function() {
+        expect(format("column::int")).toBe("column :: int\n");
+        expect(format("v->2")).toBe("v -> 2\n");
+        expect(format("v->>2")).toBe( "v ->> 2\n");
+    });
 }
--- a/src/core/Tokenizer.js
+++ b/src/core/Tokenizer.js
@@ -19,7 +19,7 @@ export default class Tokenizer {
         this.LINE_COMMENT_REGEX = /^((?:#|--).*?(?:\n|$))/;
         this.BLOCK_COMMENT_REGEX = /^(\/\*[^]*?(?:\*\/|$))/;
         this.NUMBER_REGEX = /^((-\s*)?[0-9]+(\.[0-9]+)?|0x[0-9a-fA-F]+|0b[01]+)\b/;
-        this.OPERATOR_REGEX = /^(!=|<>|==|<=|>=|!<|!>|\|\||.)/;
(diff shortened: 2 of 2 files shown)


test/PostgresTests.js
```diff
@@ -0,0 +1,29 @@
+import sqlFormatter from "./../src/sqlFormatter";
+import behavesLikeSqlFormatter from "./behavesLikeSqlFormatter";
+
+describe("StandardSqlFormatter", function() {
```
ukupat: You only added some extra logic to Tokenizer logic that is used by all of the dialects, therefore we can move these tests to behavesLikeSqlFormatter.js.

test/PostgresTests.js
```diff
@@ -0,0 +1,29 @@
...
+    behavesLikeSqlFormatter();
+
+    it("formats postgres cast operator", function() {
+        const result = sqlFormatter.format("SELECT 1::int");
+        expect(result).toBe(
+            "SELECT\n" +
+            "  1 :: int\n"
```
ukupat: Is this valid Postgres when there is spaces between `1`, `::` and `int`?

test/PostgresTests.js
```diff
@@ -0,0 +1,29 @@
...
+        );
+    });
+
+    it("formats postgres JSON '->' operator", function() { const result = sqlFormatter.format("SELECT v->2");
+        expect(result).toBe(
```
ukupat: We can simplify these tests by just giving `v->2` into the formatter.

test/PostgresTests.js
sasha-alias: Yes it is valid:
``` sql
postgres=# select 1 :: int;
 int4
------
    1
(1 row)
```

sasha-alias:
thank you for quick review, is it better now?

test/behavesLikeSqlFormatter.js
```diff
@@ -432,4 +432,25 @@ export default function behavesLikeSqlFormatter(language) {
         expect(format("'foo \\' JOIN bar'")).toBe("'foo \\' JOIN bar'\n");
         expect(format("`foo `` JOIN bar`")).toBe("`foo `` JOIN bar`\n");
     });
+
+    it("formats postgres cast operator", function() {
+        const result = sqlFormatter.format("column::int");
+        expect(result).toBe(
+            "column :: int\n"
+        );
```
ukupat: I think we can simplify this even further:
```javascript
expect(format("column::int")).toBe("column :: int\n");
```

ukupat: And we can move all of these new tests to 1 `it` block with description `formats PostgreSQL specific operators`.

sean185:
Nice! You guys are quick!

Just wondering, does it make sense to have language specific operators to be specified in the constructor?
As a future upgrade of sorts perhaps?

ukupat:
We are planning to split up the Standard SQL formatter configuration and introduce more dialects (right now we only have Standard SQL and N1QL dialect supported). This way we can go full nuts with PostgreSQL, MSSQL, DB2... specific formatting. This will probably happen when we release 2.0.

So I think that it is fine for now to keep it in the constructor.

PR #25, opened 2017-06-17
IBM DB2 support

ukupat:
I used ftp://ftp.software.ibm.com/ps/products/db2/info/vr105/pdf/en_US/DB2SQLRefVol1-db2s1e1051.pdf for learning and setting up DB2 configuration.


src/core/Tokenizer.js
```diff
@@ -57,11 +63,11 @@ export default class Tokenizer {
     // 5. national character quoted string using N'' or N\' to escape
     createStringPattern(stringTypes) {
         const patterns = {
+            "``": "((`[^`]*($|`))+)",
```
ukupat: Reordered them to match documentation order.

src/languages/Db2Formatter.js
```diff
@@ -0,0 +1,104 @@
...
+    "TRANSLATE", "TRANSLATION", "TREAT", "TRIGGER", "TRIM", "TRUE", "TRUNCATE", "TYPE",
+    "UESCAPE", "UNDO", "UNIQUE", "UNKNOWN", "UNNEST", "UNTIL", "UPPER", "USAGE", "USER", "USING",
+    "VALIDPROC", "VALUE", "VARCHAR", "VARIABLE", "VARIANT", "VARYING", "VAR_POP", "VAR_SAMP", "VCAT", "VERSION", "VIEW",
+    "VOLATILE", "VOLUMES", "WHEN", "WHENEVER", "WHILE", "WIDTH_BUCKET", "WINDOW", "WITH", "WITHIN", "WITHOUT", "WLM", "WRITE",
+    "XMLELEMENT", "XMLEXISTS", "XMLNAMESPACES",
+    "YEAR", "YEARS"
+];
+
+const reservedToplevelWords = [
+    "ADD", "AFTER", "ALTER COLUMN", "ALTER TABLE", "DELETE FROM", "EXCEPT", "FETCH FIRST", "FROM", "GROUP BY", "GO", "HAVING",
```
ukupat: Only thing that was added is "FETCH FIRST"

src/core/Tokenizer.js
```diff
@@ -12,15 +12,17 @@ export default class Tokenizer {
      *  @param {String[]} cfg.closeParens Closing parentheses to enable, like ), ]
      *  @param {String[]} cfg.indexedPlaceholderTypes Prefixes for indexed placeholders, like ?
      *  @param {String[]} cfg.namedPlaceholderTypes Prefixes for named placeholders, like @ and :
+     *  @param {String[]} cfg.lineCommentTypes Line comments to enable, like # and --
      */
     constructor(cfg) {
-        this.WORD_REGEX = /^(\w+)/;
+        this.WORD_REGEX = /^([\w|#|@]+)/;
```
ukupat: I tried playing around with word and operators tokenizing order (match operators before words) but that was no-go. Currently I went with the easiest solution that shouldn't break anything...

src/languages/Db2Formatter.js
nene: So to this group has "FETCH FIRST" compared to StandardSqlFormatter.

Might be worth to document it. Another thing: would be nice if the lines aligned with StandardSqlFormatter, so I could see differences when I just run `diff` for these files.

Of course a best diff would be one word per line. Perhaps it's actually worth doing that, although it creates pretty long files, it should be easier to maintain. Like currently I see that there are lots of differences in the `reservedWords` list, but it's really hard to tell what are the exact differences.

src/core/Tokenizer.js
nene: We should also make this WORD_REGEX configurable. We could have a config option like `specialWordChars: ["@", "#"]`.

Other SQL dialects don't allow `#` and `@` inside identifiers, which currently results in a scenario where formatting the following query with StandardSqlFormatter:

```sql
SELECT a#comment, here
```

results in:

```sql
SELECT
    a#comment,
    here
```

while correct formatting would be:

```sql
SELECT
    a #comment, here
```

test/Db2FormatterTest.js
```diff
@@ -20,6 +20,22 @@ describe("Db2Formatter", function() {
...
+            {language: "db2"}
+        );
+        expect(result).toBe(
+            "SELECT\n" +
+            "  col#1\n" +
+            "FROM\n" +
+            "  -- This is a comment\n" +
+            "  MyTable;\n"
+        );
+    });
```
nene: We should also test the formatting of `@` and `#` inside identifiers. This test partially covers the latter, but it would be better to have a separate test with clear description, explaining the `@` and `#` are part of identifiers (e.g. not treated as operators).

src/languages/Db2Formatter.js
ukupat: I tried making toplevel and newline words easier to compare because these are the most important ones.

I think one word per line is overkill. If you really want to compare differenct dialects config then you can use online tools or write a quick script with Lodash (that's what I did).

src/core/Tokenizer.js
```diff
@@ -49,6 +50,10 @@ export default class Tokenizer {
         return new RegExp(`^(${reservedWordsPattern})\\b`, "i");
     }
 
+    createWordRegex(specialChars = []) {
+        return new RegExp(`^([\\w|${specialChars.join("|")}]+)`);
```
nene: This will result in regex like: `^([\w|@|#]+)`

As the `|` characters are part of character class `[ ... ]` this will match stuff like `foo|bar` as a single word, as `|` has no special meaning inside square brackets.

test/Db2FormatterTest.js
```diff
@@ -38,7 +38,8 @@ describe("Db2Formatter", function() {
 
     it("recognizes @ and # as identifiers", function() {
```
nene: It would be more correct to say: "as **part** of identifiers"

PR #27, opened 2017-07-10
Remove addition of newline to end of output.

rahelest:


ukupat:
Hmmm looks good. We have to bump major version for this release because the output is different.

PR #36, opened 2018-02-08
Adding more Postgres specific operators

Offir:
Postgresql has several specific operators(addition to [#24](https://github.com/zeroturnaround/sql-formatter/pull/24))
more info can be found: https://www.postgresql.org/docs/8.4/static/functions-matching.html
`~~` - represent LIKE
`~~*` - represent ILIKE
`!~~` - represent NOT LIKE
`!~~*` represent NOT ILIKE
Regular Expression Match Operators:
`~*`,
`!~`,
`!~*`,
This commit adds them to the list of standard operators in order to not
break them when formatting.


kensaggy:
+1 for this.

nene:
Thanks!

PR #39, opened 2018-02-27
Add CASE…WHEN formatting for standard SQL

damon-davison:
This makes CASE…WHEN statements tidier, as follows:

```sql
CASE
  expression
  WHEN 'foo' THEN 1
  WHEN 'bar' THEN 2
  WHEN 'baz' THEN 3
  ELSE 4
END;
```

```sql
CASE
  WHEN 'foo' THEN 1
  WHEN 'bar' THEN 2
  WHEN 'baz' THEN 3
  ELSE 4
END;
```


nene:
Thanks for the contribution @damon-davison and sorry for the long feedback-delay.

Your changes work well for formatting a stand-alone CASE-WHEN statement, but what about using it inside another SQL expression, like:

```
SELECT foo, bar, CASE baz WHEN 'one' THEN 1 WHEN 'two' THEN 2 ELSE 3 END FROM table
```

I would expect this to be formatted as:

```
SELECT
  foo,
  bar,
  CASE baz
    WHEN 'one' THEN 1
    WHEN 'two' THEN 2
    ELSE 3
  END
FROM
  table
```

but instead I get this:

```
SELECT
  foo,
  bar,
CASE baz
  WHEN 'one' THEN 1
  WHEN 'two' THEN 2
  ELSE 3
END
FROM
  table
```

I fiddled with this for a while and concluded that the current system doesn't really allow us to format such nested constructs properly. The CASE...END block needs to work like the parentheses do... and I can kinda make it work by defining:

```
                openParens: ["(", "CASE"],
                closeParens: [")", "END"],
```

but the quirk is that currently open-parenthesis always starts a new line, so the result looks a bit off:

```
SELECT
  foo,
  bar,
  CASE
    baz
    WHEN 'one' THEN 1
    WHEN 'two' THEN 2
    ELSE 3
  END
FROM
  table
```

@damon-davison would you be interested in polishing this further?

damon-davison:
@nene Yeah, I agree it could use some polishing. I'm afraid I don't have the time to work on it further at the moment. Maybe this small improvement is good enough for now and the next phase can live as an issue until someone finds the time to refine it. 😄

nene:
Implemented the formatting of CASE...WHEN using my suggested approach of treating them similarly to pair of parenthesis.

Released in 2.2.0

PR #47, opened 2018-08-18
No transpilation for nodejs usage

apla:
To me, it is not a good idea to install `babel-runtime` just for es module `import`. So, I replaced babel with [esm](https://github.com/standard-things/esm). Unfortunately, `jest` have no idea of es modules and [needs](https://github.com/facebook/jest/issues/4842) [pretty heavy patches](https://github.com/kenotron/esm-jest/), so I've used `mocha` and node's `assert`.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,5 +1,11 @@
 {
-    "extends": "zt/base",
+    "parserOptions": {
+        "ecmaVersion": 7,
+        "sourceType": "module",
+        "ecmaFeatures": {
+            "modules": true
+        }
+    },
     "env": {
         "jest": true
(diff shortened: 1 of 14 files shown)


nene:
Thanks for the PR. It's quite a major change you're proposing though.

I took a look at this and it seems that `babel-runtime` is actually not needed at all. So I simply dropped it. Not sure why it was added in the first place. Possibly Babel has improved since.

Also planning to upgrade Babel itself. But I'm planning to keep it. In addition to modules I'd also like to use the ES2018 unicode-property-escapes feature to fix some bugs in current regexes.

PR #52, opened 2018-10-25
Formatter adds semicolon onto last line inside of line comment

jylertones:
When inputting a line comment on the line above the semicolon, the semicolon is formatted onto that line. When the query is executed, it is missing a semicolon and may be considered incomplete SQL.

Query:

```
SELECT * FROM user
-- no where clause here
;
```
Expected:

```
SELECT
  *
FROM
  user 
-- no where clause here
;
```
Actual:

```
SELECT
  *
FROM
  user -- no where clause here;
```

### DIFF
--- a/src/core/Formatter.js
+++ b/src/core/Formatter.js
@@ -74,7 +74,13 @@ export default class Formatter {
                 formattedQuery = this.formatWithSpaceAfter(token, formattedQuery);
             }
             else if (token.value === "." || token.value === ";") {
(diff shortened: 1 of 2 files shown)


nene:
Hey, thanks for the bug report and patch!

Turned out this issue goes a bit deeper and doesn't only effect semicolons, but also several other cases. Covered the other cases by myself.

nene:
Released the fix in 2.3.2

PR #54, opened 2018-11-03
Add option 'reservedWordConverter' and then can covert sql reserved w…

wangzhengbo:
…ords to upper case (when reservedWordConverter set to 'toUpperCase') or lower case (when reservedWordConverter set to 'toLowerCase') or any other value (when reservedWordConverter set to a function).

### DIFF
--- a/src/core/Formatter.js
+++ b/src/core/Formatter.js
@@ -35,6 +35,18 @@ export default class Formatter {
         return formattedQuery.trim();
     }
 
+    convertReservedWord(token) {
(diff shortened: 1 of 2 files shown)


nene:
Hi @wangzhengbo, thanks for a pull request.

Have you looked into the already existing pull request on the same topic: https://github.com/zeroturnaround/sql-formatter/pull/22

How is your solution different? How does your code solve the problems pointed out in there?

nene:
Implemented differently with simple `uppercase: true|false` option.

PR #57, opened 2018-12-10
Update README.md

jfoclpf:
--save is deprecated in `npm i`

I suppose using `require` is cleaner than import as it is standard in node

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,13 +11,13 @@ It supports [Standard SQL][], [Couchbase N1QL][], [IBM DB2][] and [Oracle PL/SQL
 Get the latest version from NPM:
 
 ```
-npm install --save sql-formatter
+npm install --save-dev sql-formatter
 ```
 
 ## Usage
 
 ```js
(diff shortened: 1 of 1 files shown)


nene:
Indeed, `--save` is deprecated, but the new replacement is not `--save-dev` but `--save-prod`. The latter though is the default, so better to just write:

```
npm install sql-formatter
```

I'd prefer keeping the use of `import` syntax, as it's is more future proof. Node might not support it yet, but it soon will. And one can also use this library on client side.

jfoclpf:
I think this library is to be used, if used with npm, merely within
development mode. Why would you force the final user to download this
library? It's not just for the purpose of nice pretty log? Maybe I'm
missing something.

Yes indeed with import you can use client side

jfoclpf:
Ah, I see the point.
Yes indeed then just npm i sql-formatter should be used then.

nene:
Dropped the `--save` flag from README.

PR #58, opened 2018-12-20
feat: support multiple statements

rafaelrpinto:
This PR attempts to improve formatting of multiple statement SQLs like it was discussed in here: https://github.com/zeroturnaround/sql-formatter/issues/50

Basically we keep the semicolon formatting as is unless its followed by one or more line breaks and have another statement after it. In that case we preserve one line break between the statements.

I think this allows some formatting of multiple statements without interfering with the existing behaviour.

### DIFF
--- a/test/StandardSqlFormatterTest.js
+++ b/test/StandardSqlFormatterTest.js
@@ -358,7 +358,8 @@ describe("StandardSqlFormatter", function() {
             "SELECT\n" +
             "  a\n" +
             "FROM\n" +
-            "  b --comment;"
+            "  b --comment\n" +
+            ";"
         );
     });
 
--- a/test/StandardSqlFormatterTest.js
+++ b/test/StandardSqlFormatterTest.js
@@ -358,8 +358,7 @@ describe("StandardSqlFormatter", function() {
             "SELECT\n" +
             "  a\n" +
             "FROM\n" +
-            "  b --comment\n" +
-            ";"
+            "  b --comment;"
         );
     });
 
--- a/src/core/Formatter.js
+++ b/src/core/Formatter.js
@@ -189,7 +189,7 @@ export default class Formatter {
     }
 
     formatQuerySeparator(token, query) {
-        return trimEnd(query) + token.value + "\n";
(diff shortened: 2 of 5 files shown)


nene:
Hi @rafaelrpinto, thanks for a patch! Proper multiple-query support is something we definitely need.

It's nice that you have thought about potential backwards-compatibility issues. But I think we're better off not preserving the current (buggy) formatting of semicolons. Unless the whitespace in input SQL is somehow significant, it should not effect the output. Breaking up a chain of SQL statements on a single line is something one would expect a formatter to do.

Would you be interested in changing this pull request so that END_STATEMENT = `;` ?

Also, I suggest we rename `END_STATEMENT` token to `QUERY_SEPARATOR`.

nene:
PS. Also, it looks that your git config is missing author name. I see:

```
Author: yourname <raprp@posteo.de>
```

rafaelrpinto:
@nene 
Thanks for the heads-up on the git config :+1: 

Since we will simply consider the semicolon as line break always I rolled back most of the changes and took a simplified approach without creating any token types. 

Please let me know if that was what you had in mind.

Thanks.

test/StandardSqlFormatterTest.js
```diff
@@ -358,8 +358,7 @@ describe("StandardSqlFormatter", function() {
             "SELECT\n" +
             "  a\n" +
             "FROM\n" +
-            "  b --comment\n" +
-            ";"
+            "  b --comment;"
```
nene: This is a bug. Semicolon after a comment should not be moved to the end of the comment.

You should be able to fix this by calling `trimTrailingWhitespace` instead of `trimEnd` inside the new `formatQuerySeparator` function.

rafaelrpinto: @nene 
Thanks for the review. I made the adjustments you asked and the test now is behaving like it was before.

nene:
Thanks. Look great! I'll merge this and hopefully will also release soonish (I'd like to also eliminate a security alert I'm getting because of an outdated dev-dependency).

osv:
Bump for release :)

PR #61, opened 2019-01-16
Remove # as Standard SQL comment

michalkrcik-swi:


nene:
I agree that `#` is not a standard SQL comment. Only `--` is.

However I can't accept this PR as is. We still need have a support `#` comments somewhere.

How would you propose solving this?

PR #64, opened 2019-03-04
comment line maybe crlf

sally2015:
My project has the problem that when users copy the sql from windows platform, formatter cannot match the sql of comment，instead of thinking them as the operators.

before
image

after
image

so I add the reg `\r\n`

### DIFF
--- a/src/core/Tokenizer.js
+++ b/src/core/Tokenizer.js
@@ -43,7 +43,7 @@ export default class Tokenizer {
     }
 
     createLineCommentRegex(lineCommentTypes) {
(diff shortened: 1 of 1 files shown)


nikylin:
but I installed version ^2.3.2 by npm,it still has this problem.and to solve it ,i use an display-none textarea.every time before i try to format the text ,i assign the text to textarea and then get value from it.

nene:
It's not really a proper fix, but for now a similar partial fix is included in #112

nene:
Now part of 3.1.0 release

PR #68, opened 2019-03-20
docs: updated to reference the correct package name

nick-woodward:


nick-woodward:
Looks like `@teamsqlio` actually just forked this repository.

PR #84, opened 2019-09-19
use `lodash/includes` instead of `Array.prototype.includes`

nicoder:
in order to be able to use the library in IE11 without needing a polyfill


nene:
Thanks!

PR #86, opened 2019-10-19
Fix for unicode characters

### LINKED ISSUE #46: Not support Cyrillic symbols
Example:
Original:
```sql
SELECT test, тест FROM table;
```
Result:
```sql
SELECT
  test,
  т е с т
FROM
  table;
```

kufii:
Fixes #46


nene:
Now part of #112

PR #87, opened 2019-10-29
reset indentation for new queries

kufii:
```sql
select * from test;
CREATE TABLE TEST(id NUMBER NOT NULL, col1 VARCHAR2(20), col2 VARCHAR2(20));
```
was getting formatted like

```sql
SELECT
  *
FROM
  test;
CREATE TABLE TEST(
    id NUMBER NOT NULL,
    col1 VARCHAR2(20),
    col2 VARCHAR2(20)
  );
```
this fixes it

### DIFF
--- a/src/core/Formatter.js
+++ b/src/core/Formatter.js
@@ -190,6 +190,7 @@ export default class Formatter {
     }
 
     formatQuerySeparator(token, query) {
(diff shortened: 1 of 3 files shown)


nene:
Now part of larger PR: #112

PR #95, opened 2020-01-21
Add support for PG bitwise shift, square / cube root, and absolute value operators.

iancamp:
Added support for the following operators:
* @ absolute value (wasn't really a problem, but added it to the regex anyway)
* |/ (square root)
* ||/ (cube root)
* << (bitwise shift left)
* \>> (bitwise shift right)


nene:
Thanks. This is now included to 3.1.0 release.

PR #114, opened 2021-01-22
Supports setting specialwordChars with the cfg parameter

yliu199x:
In some projects, Freemarker will be used in the SQL file. SQL - Formatter Formatting SQL corrupts Freemarker statements. For example: "<#if stat_type = =" 0 ">" after formatting will become "< #if stat_type = =" 0 “>”(a space is added between the symbols "<" and "#"). Such a statement cannot be executed normally. Whether more configurable items can be opened in the cfg parameter.? Through the cfg parameter, improve the flexibility of the component configuration.


nene:
Sorry, but I won't accept a broken pull request.

PR #130, opened 2021-03-15
support square brackets in Spark

forestfang-stripe:
Support square brackets for SparkSQL which is used to access data in complex data types such as array and structs.

I couldn't find an explicit documentation of this support in SparkSQL documentation. However, it can be inferred based on the Hive compatibility doc and Hive documentation itself:
- https://docs.databricks.com/spark/latest/spark-sql/compatibility/hive.html
- https://cwiki.apache.org/confluence/display/Hive/LanguageManual+UDF#LanguageManualUDF-OperatorsonComplexTypes

FWIW this syntax is commonly used in SparkSQL examples: https://stackoverflow.com/questions/28332494/querying-spark-sql-dataframe-with-complex-types


nene:
Thanks for the PR.

This issue is now fixed in SQL Formatter 6.0

PR #134, opened 2021-03-20
Add option to indent query separator ";" on a final new line

romainr:
A new conflig flag might be overkill or this could be yet another non opiniated option, but sharing in case there someone knows a smarter way to propose the same functionalities with as less friction as possible.

Flag:

--indent-query-separator
Indent query separator ";" on a final new line

e.g.

If set to true:

      SELECT * FROM foo;

the ending ';' will be indented on its own line:

      SELECT
        *
      FROM
        foo
      ;

Instead of:

      SELECT
        *
      FROM
        foo;


nene:
This feature has now been incorporated to sql-formatter. There's now a new config option named `newlineBeforeSemicolon`.

Thanks for the PR.

PR #136, opened 2021-03-28
Add support for Salesforce SOQL

paustint:
This PR adds support for SOQL - https://developer.salesforce.com/docs/atlas.en-us.soql_sosl.meta/soql_sosl/sforce_api_calls_soql_select.htm

There are a couple of issues that I could use some help on resolving:

1. SOQL supports [date literals](https://developer.salesforce.com/docs/atlas.en-us.soql_sosl.meta/soql_sosl/sforce_api_calls_soql_select_dateformats.htm) with variables, for example `WHERE CreatedDate = LAST_N_MONTHS:12`. the issue is that the formatter either puts a space before or after the `:` depending on if `:` is included in the `specialWordChars` array. This is not desired, there should not be a space at all. **Any thoughts on how to solve this?**
```diff
+   CloseDate < NEXT_N_FISCAL_YEARS:3 --> desired outcome
-   CloseDate < NEXT_N_FISCAL_YEARS :3 --> specialWordChars: [':']
-  CloseDate < NEXT_N_FISCAL_YEARS: 3 --> specialWordChars: [ ]
```
2. SOQL supports a [TYPEOF](https://developer.salesforce.com/docs/atlas.en-us.soql_sosl.meta/soql_sosl/sforce_api_calls_soql_select_typeof.htm) query, which is similar to a CASE statement and is used to determine which fields to show in which cases.
    1.  The `TYPEOF` keyword is followed by the name of the table, so it would be desirable to include the name of the table on the same line as the `TYPEOF` clause instead of a new line, which looks weird.
    2. The  `THEN` and `ELSE` clauses typically have more than one field specified and they are formatted on a new line( which is fine) but they are not indented, and it would look a lot better if they were. Or even if the list of fields after would not be indented at all, that would be better as well. **I can live with this as-is, as TYPEOF is not extremely common, but it would be nice to have better formatting**

`SELECT TYPEOF What WHEN Account THEN Phone, NumberOfEmployees WHEN Opportunity THEN Amount, CloseDate END FROM Event`

```diff
+TYPEOF What
+  WHEN Account THEN Phone,
+   NumberOfEmployees
+  WHEN Opportunity THEN Amount,
+   CloseDate
+  ELSE Name,
+   Email
+END

-TYPEOF
-  What
-  WHEN Account THEN Phone,
-  NumberOfEmployees
-  WHEN Opportunity THEN Amount,
-  CloseDate
-  ELSE Name,
-  Email
-END
```

This syntax commonly has a list of fields in the

**Summary of Changes:**
- Added support for SOQL
- Added unit tests specific to the language
- Fixed typo with exception error message
- Updated readme with changes

cc @AllanOricil


AllanOricil:
@paustint add a test for formatting a SOQL with Dates https://developer.salesforce.com/docs/atlas.en-us.soql_sosl.meta/soql_sosl/sforce_api_calls_soql_select_dateformats.htm


The hack I did to be able to get away with that problem with " : " was to run this small code after formatting the query


````js
const formatDateTime = (input) => {
  const inputChars = input.split('')
  for (let i = input.length - 1; i > 0; i--) {
    if (
      ['-', ':', '+'].includes(input[i]) &&
      input[i - 1] &&
      input[i - 1] === ' '
    ) {
      inputChars.splice(i - 1, 1)
    }
  }
  return inputChars.join('')
}
````

AllanOricil:
This is the line that causes the problem for SOQL    https://github.com/zeroturnaround/sql-formatter/blob/master/src/core/Formatter.js#L95

AllanOricil:
If we have a new token type called, perhaps, REGEX_EXPRESSION, then we can use regex for Date, datetime, and date literal keywords.

paustint:
closing since the PR has not been addressed by maintainers.

PR #138, opened 2021-04-08
Valid and explicit uppercase configuration example

romainr:
Kind of a nit but more consistent and easier to quick test.


nene:
Thanks for the PR. By now the uppercase option has been renamed, but still, thanks.

Issue #129, opened 2021-03-11
Type error while formatting


Hwan-seok:
## Wanna format
```
CREATE TABLE IF NOT EXISTS `user_associated_capsule` (`user_uid` VARCHAR(255) , `capsule_uid` CHAR(36) BINARY , `did_open_request` TINYINT(1) DEFAULT false, `is_opened` TINYINT(1) DEFAULT false, PRIMARY KEY (`user_uid`, `capsule_uid`), FOREIGN KEY (`user_uid`) REFERENCES `user` (`uid`) ON DELETE CASCADE ON UPDATE CASCADE, FOREIGN KEY (`capsule_uid`) REFERENCES `capsule` (`uid`) ON DELETE CASCADE ON UPDATE CASCADE) ENGINE=InnoDB;
```

## Expected result
```
CREATE TABLE IF NOT EXISTS `user_associated_capsule` (
    `user_uid` VARCHAR(255),
    `capsule_uid` CHAR(36)BINARY,
    `did_open_request` TINYINT(1) DEFAULT FALSE,
    `is_opened` TINYINT(1) DEFAULT FALSE,
    PRIMARY KEY (`user_uid` , `capsule_uid`),
    FOREIGN KEY (`user_uid`)
        REFERENCES `user` (`uid`)
        ON DELETE CASCADE ON UPDATE CASCADE,
    FOREIGN KEY (`capsule_uid`)
        REFERENCES `capsule` (`uid`)
        ON DELETE CASCADE ON UPDATE CASCADE
)  ENGINE=INNODB;
```

## Actual 
```
TypeError: Cannot read property 'slice' of undefined at parseKey(\sql-formatter\lib\core\Tokenizer.js:191:20)
 at Tokenizer.getPlaceholderTokenWithKey(\sql-formatter\lib\core\Tokenizer.js:235:21)
...

```

#### Dialect
- mysql

psacawa:
I encounter this with a simpler case:

*command.sql*

```
SeleCt * from mytable where true  and false ;
```

When running `cat command.sql | sql-formatter`, I get:

```
/home/psacawa/.nvm/versions/node/v14.15.0/lib/node_modules/sql-formatter/lib/core/Tokenizer.js:191
          return v.slice(1);
                   ^

TypeError: Cannot read property 'slice' of undefined
[...]

PR #142, opened 2021-04-26
:bug: fix format error while sql includes false

xiaokk06:
while default namedPlaceholderTypes is empty array,  this.IDENT_NAMED_PLACEHOLDER_REGEX is false.
```javascript
    this.IDENT_NAMED_PLACEHOLDER_REGEX = regexFactory.createPlaceholderRegex(
      cfg.namedPlaceholderTypes,
      '[a-zA-Z0-9._$]+'
    );
```
then you got { type: 'palceholder', value: undefined } while formatter sql which includes false.
```javascript
  getPlaceholderTokenWithKey({ input, regex, parseKey }) {
    const token = this.getTokenOnFirstMatch({ input, regex, type: tokenTypes.PLACEHOLDER });
    if (token) {
      token.key = parseKey(token.value);
    }
    return token;
  }
```
and then error occurs while call this function in getIndexedPlaceholderToken.
```
  getIndexedPlaceholderToken(input) {
    return this.getPlaceholderTokenWithKey({
      input,
      regex: this.INDEXED_PLACEHOLDER_REGEX,
      parseKey: (v) => v.slice(1), // v is undefined
    });
  }
```
so it will show the error below in console:
image

the merge request resolve this porblem which also refered in #129

### DIFF
--- a/src/core/Tokenizer.js
+++ b/src/core/Tokenizer.js
@@ -272,6 +272,9 @@ export default class Tokenizer {
   }
 
   getTokenOnFirstMatch({ input, type, regex }) {
+    if (!regex) {
+      return undefined;
+    }
     const matches = input.match(regex);
 
     return matches ? { type, value: matches[1] } : undefined;


iwata:
@nene Any progress?

nene:
Thanks, this bug has been fixed now with different code, but thanks for pointing out the issue.

PR #153, opened 2021-06-14
Fix tokenizer number regex to include scentific notation

xstevenyung:
This PR fixes the number regex for scientific notation with a "+" on `Tokenizer` (e.g. 2E+3)

I don't really know where to create an associated test for the change. If you can give me some pointers, I would gladly create a test for this.

Related issue:
https://github.com/zeroturnaround/sql-formatter/issues/152

Thanks for the work you provide 🙏


nene:
Thanks. This fix has now been incorporated.

PR #154, opened 2021-06-24
feat: add `module` field into package.json for esm

### LINKED ISSUE #105: build es6 version of the sql-formatter
In order to use sql-formatter in an es6 project I had to import it in this format `import sqlFormatter from 'sql-formatter/src/sqlFormatter.js';` and then download all the dependencies for sql-formatter which is currently only "lodash" in its es6 version.

I would like to use an es6 version of the sql-formatter at runtime without any bundling/packaging step so that we can import it in a proper es6 format import `import sqlFormatter from "sql-formatter";`

JounQin:
close #105

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "description": "Format whitespace in a SQL query to make it more readable",
   "license": "MIT",
   "main": "lib/sqlFormatter.js",
+  "module": "src/sqlFormatter.js",
   "types": "src/sqlFormatter.d.ts",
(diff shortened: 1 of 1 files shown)


nene:
This patch no more works now that the whole thing has been converted to TypeScript.

JounQin:
@nene Ideally, it can be released as a dual package with commonjs + ESM at the same time?

https://web.dev/commonjs-larger-bundles/

PR #163, opened 2021-08-24
Add --lines-after-output option

yowcow:
In environments like NeoVim, currently appended trailing newline is unnecessary.

Similar to `--lines-between-queries` option, `--lines-after-output` will provide control over trailing newlines.


nene:
Sorry, but won't do.

It's standard practice for command-line tools to end their output with a newline. If you really need to avoid it, you can pipe this to some other tool or use sql-formatter as a library function, which won't add a newline to the end of its return value.

PR #165, opened 2021-08-28
postgresql update, better formatting for create function

ray-x:
Update keyword, better formatting for `create function`


nene:
Sorry, not accepting this. Too many different changes in this pull request and no tests or examples of what exactly it fixes.

PR #172, opened 2021-09-16
fix: Skip tokenization if regex is false

### LINKED ISSUE #147: When there is "false" in SQL, the format will  fail
input 
```
CREATE TABLE run_stat(
  carId int, avg_speed double
  ) WITH (
  'connector.type' = 'jdbc',
  'connector.url' = 'jdbc:mysql://199.1XX.1XX.XXX:3306/aaa?useSSL=false',
  'connector.table' = 'mysql_run_log',
  'connector.username' = 'aaa',
  'connector.password' = 'aaa');
```
There will be errors during the formatting process, and the test will fail in this[online site](https://zeroturnaround.github.io/sql-formatter/). The current version is 4.0.2, but it is normal to use 2.3.3.

czgu:
Closes #147

Skip the regex match if regex is falsey. Also added unit tests for prevent future cases

### DIFF
--- a/src/core/Tokenizer.js
+++ b/src/core/Tokenizer.js
@@ -20,7 +20,8 @@ export default class Tokenizer {
    */
   constructor(cfg) {
     this.WHITESPACE_REGEX = /^(\s+)/u;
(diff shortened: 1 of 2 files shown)


nene:
This has been fixed now, but with different code. Thanks for the PR.

PR #182, opened 2021-10-26
fix #181

### LINKED ISSUE #181: Don't add break line for  `CHARACTER SET` in MariaDbFormatter
Hi, seems like with current options, there is no way to archive that.
I think this is because `SET` is included in `reservedTopLevelWords`. 
Perhaps,we need an exception for `CHARACTER SET`
Current result:

    ALTER TABLE
       `test`.`t4` CHARACTER
    SET
       = dec8,
       COLLATE = dec8_swedish_ci,
       DROP COLUMN `c`,
       CHANGE COLUMN `b` `b` MEDIUMINT(11) UNSIGNED ZEROFILL NOT NULL; 

Expected result:

    ALTER TABLE `test`.`t4` 
    CHARACTER SET = dec8 , COLLATE = dec8_swedish_ci ,
    DROP COLUMN `c`,
    CHANGE COLUMN `b` `b` MEDIUMINT(11) UNSIGNED ZEROFILL NOT NULL ;

onmotion:
Add CHARACTER SET keyword support for mariaDB & MySQL

```sql
ALTER TABLE t MODIFY col1 VARCHAR(50) CHARACTER SET greek
```

compiled to

```sql
ALTER TABLE
  t MODIFY col1 VARCHAR(50) CHARACTER SET greek
```

instead of

```sql
ALTER TABLE
  t MODIFY col1 VARCHAR(50) CHARACTER
SET
  greek
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,13169 @@
 {
   "name": "sql-formatter",
   "version": "4.0.2",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
+      "name": "sql-formatter",
+      "version": "4.0.2",
+      "license": "MIT",
+      "dependencies": {
+        "argparse": "^2.0.1"
+      },
+      "bin": {
+        "sql-formatter": "bin/sqlfmt.js"
+      },
+      "devDependencies": {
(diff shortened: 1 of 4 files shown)


nene:
Thanks for the fix. This is now finally merged (manually).

PR #190, opened 2021-12-23
Fix dependency vulnerabilities and migrate to Webpack 5

jacobobq:
This solves various npm reported vulnerabilities. In order to do this, it was necessary to migrate the project to Webpack 5 and update the version of babel-related packages.

No changes were necessary in the webpack configuration itself as no deprecated features were in use.


nene:
Thanks for the PR. These dependencies have been upgraded now.

PR #193, opened 2022-01-24
Add string types to standard sql

hokaccha:
This PR supports backquotes, raw strings, and raw bytes literals for Standard SQL.

I referred to the bigquery documentation to check the specifics of Standard SQL.
https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical

## Input

```sql
SELECT `col` FROM `tbl` WHERE col = r'val'
```

## Current behavior 

```sql
SELECT
  ` col `
FROM
  ` tbl `
WHERE
  col = r 'val'
```

## after

```sql
SELECT
  `col`
FROM
  `tbl`
WHERE
  col = r'val'
```


nene:
The backticks are now part of standard sql formatter.

Not sure about the `r` and `b` prefixed strings though. Probably should add support for them to bigquery formatter. Created issue: #205

PR #194, opened 2022-02-02
CLI: Make stdout output match written file content

teejae:
Change sql-formatter CLI to output formatted content without extra newline, to match written file content.
- `console.log` -> `process.stdout.write` to not use redundant `\n`
  - Newline is already appended to the `formattedQuery`, so no need for redundant `\n`
- Can be used to diff against original file content


nene:
Thanks for the PR. I've incorporated this change. Though had to do it manually as the master branch has diverged significantly by now.

PR #195, opened 2022-02-02
CLI: Add --check-diff option to check for diffs in output

teejae:


nene:
Thanks for the PR, but not accepting this feature.

This job is better left for a separate command line tool (which could generically compare input & output of any command).

PR #196, opened 2022-02-10
Add RETURNING to postgres top-level words

SteveDeWald:
In the official docs\[1\] `RETURNING` is given the same formatting treatment as `WHERE` and other top-level words.

\[1\]: https://www.postgresql.org/docs/14/dml-returning.html


nene:
Thanks for the PR.

The RETURNING keyword is now properly recognized in Postgres, MariaDB, N1QL and PL/SQL.

PR #197, opened 2022-03-02
Wrap parentheses based on entire line length

johanholmerin:
This probably needs further work but I at least wanted to get it started.

## The issue

The formatter currently wraps parentheses if the content is above a certain length(50). It does not however account for the length of the line before the parentheses. This PR passes the current line length to the wrap check so that it can be used as the initial value, wrapping lines earlier. Lines that don't contain parentheses can still become very long.

I also updated the line length check to account for whitespace, but maybe not doing so was per design.

### Input

```sql
CREATE TABLE some_long_table_name (a INT PRIMARY KEY, b TEXT, c INT NOT NULL, d INT NOT NULL);
```

### Output before

```sql
CREATE TABLE some_long_table_name (a INT PRIMARY KEY, b TEXT, c INT NOT NULL, d INT NOT NULL);
```

### Output after

```sql
CREATE TABLE some_long_table_name (
  a INT PRIMARY KEY,
  b TEXT,
  c INT NOT NULL,
  d INT NOT NULL
);
```

## Related issues:

- https://github.com/zeroturnaround/sql-formatter/issues/40
- https://github.com/zeroturnaround/sql-formatter/issues/5

### DIFF
--- a/src/core/Formatter.js
+++ b/src/core/Formatter.js
@@ -84,7 +84,8 @@ export default class Formatter {
         formattedQuery = this.formatWithSpaces(token, formattedQuery);
         this.previousReservedToken = token;
       } else if (token.type === tokenTypes.OPEN_PAREN) {
-        formattedQuery = this.formatOpeningParentheses(token, formattedQuery);
+        const prevLength = formattedQuery.split('\n').pop().length;
+        formattedQuery = this.formatOpeningParentheses(token, formattedQuery, prevLength);
       } else if (token.type === tokenTypes.CLOSE_PAREN) {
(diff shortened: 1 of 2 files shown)


johanholmerin:
Partially fixed by 6436772887c748c8464051e6af9a5ac5d3251ce0. The issue with whitespace not being included in the length calculation remains

nene:
Yeah. The whitespace issue is definitely not by design. Hoping to alleviate the whole situation by a larger redesign of the whole formatter.

Thanks for your PR though. It's definitely has the right idea functionality-wise.

PR #207, opened 2022-05-04
Set exec permission to cli

htaketani:
It seems exec permission of CLI command were lost when filename renamed.
I set exec permission to CLI command.


nene:
Thanks for noticing and sending a patch!

PR #208, opened 2022-05-04
Allow no arguments when cli command is piped

htaketani:
The cli command shows help when no arguments passed.

```
# show help
sql-formatter
```

However, help is also displayed when piped or redirected, as in the following example.

```
# pipe
echo 'select * from tbl where id = 3' | sql-formatter

# redirect
sql-formatter < input.sql
```

I have modified the cli command to not show help when stdin is not tty.


nene:
Thanks for the patch. Much appreciated.

PR #220, opened 2022-05-31
Update typescript imports to use absolute paths

inferrinizzard:
- resolved misc merge conflicts
- bumped typescript version
- added config in tsconfig.json and webpack.config.js to allow for absolute paths
- updated all imports to use absolute paths for clarity
- also added partial type imports where applicable


nene:
Merged this in by cherry-picking the import path-related commits.

PR #238, opened 2022-06-14
Attempt to optimize tokenizer

nene:
Instead of chewing off bits from the input string, move the regex matching start position forward.

Measured the performance from command line and in browsers using https://jsben.ch site using the following code:

```js
sqlFormatter.format("select supplier_name,city from (select * from suppliers join addresses on suppliers.address_id=addresses.id) as suppliers where supplier_id>500 order by supplier_name asc,city desc; ALTER TABLE `Album` ADD CONSTRAINT `FK_AlbumArtistId` FOREIGN KEY (`ArtistId`) REFERENCES `Artist` (`ArtistId`) ON DELETE NO ACTION ON UPDATE NO ACTION; CREATE TABLE `Customer` ( `CustomerId` INT NOT NULL AUTO_INCREMENT, `FirstName` NVARCHAR(40) NOT NULL, `LastName` NVARCHAR(20) NOT NULL, `Company` NVARCHAR(80), `Address` NVARCHAR(70), `City` NVARCHAR(40), `State` NVARCHAR(40), `Country` NVARCHAR(40), `PostalCode` NVARCHAR(10), `Phone` NVARCHAR(24), `Fax` NVARCHAR(24), `Email` NVARCHAR(60) NOT NULL, `SupportRepId` INT, CONSTRAINT `PK_Customer` PRIMARY KEY  (`CustomerId`)); INSERT INTO `Track` (`Name`, `AlbumId`, `MediaTypeId`, `GenreId`, `Composer`, `Milliseconds`, `Bytes`, `UnitPrice`) VALUES (N'Jump Around', 258, 1, 17, N'E. Schrody/L. Muggerud', 217835, 8715653, 0.99); INSERT INTO `Track` (`Name`, `AlbumId`, `MediaTypeId`, `GenreId`, `Composer`, `Milliseconds`, `Bytes`, `UnitPrice`) VALUES (N'Salutations', 258, 1, 17, N'E. Schrody/L. Dimant', 69120, 2767047, 0.99); INSERT INTO `Track` (`Name`, `AlbumId`, `MediaTypeId`, `GenreId`, `Composer`, `Milliseconds`, `Bytes`, `UnitPrice`) VALUES (N'Put Your Head Out', 258, 1, 17, N'E. Schrody/L. Freese/L. Muggerud', 182230, 7291473, 0.99); INSERT INTO `Track` (`Name`, `AlbumId`, `MediaTypeId`, `GenreId`, `Composer`, `Milliseconds`, `Bytes`, `UnitPrice`) VALUES (N'Top O'' The Morning To Ya', 258, 1, 17, N'E. Schrody/L. Dimant', 216633, 8667599, 0.99); INSERT INTO `Invoice` (`CustomerId`, `InvoiceDate`, `BillingAddress`, `BillingCity`, `BillingState`, `BillingCountry`, `BillingPostalCode`, `Total`) VALUES (48, '2013/8/2', N'Lijnbaansgracht 120bg', N'Amsterdam', N'VV', N'Netherlands', N'1016', 1.98);", {
  language: "sql",
});
```

Also did another measurement with 10x longer input string. 

- NodeJS: no difference (7 x faster with longer string)
- Chrome: 3 x faster (15 x faster with longer string)
- Firefox: 2 x faster (10 x faster with longer string)
- Safari: 3 x faster (25 x faster with longer string)

For some reason though I'm failing to see any performance difference when using sql-formatter program from command line with a large input file. However when using it as a library I do see a difference in NodeJS.

### DIFF
--- a/src/core/formatCommaPositions.ts
+++ b/src/core/formatCommaPositions.ts
@@ -1,7 +1,7 @@
 import type { CommaPosition } from 'src/types';
 import { maxLength } from 'src/utils';
 
-import { WHITESPACE_REGEX } from './Tokenizer';
+const WHITESPACE_REGEX = /^(\s+)/u;
 
 /**
  * Handles comma placement - either before, after or tabulated
--- a/src/core/Tokenizer.ts
+++ b/src/core/Tokenizer.ts
@@ -3,8 +3,8 @@ import { equalizeWhitespace, escapeRegExp, id } from 'src/utils';
 import * as regexFactory from './regexFactory';
 import { type Token, TokenType } from './token';
 
-export const WHITESPACE_REGEX = /^(\s+)/u;
-const NULL_REGEX = /(?!)/; // zero-width negative lookahead, matches nothing
+const WHITESPACE_REGEX = /(\s+)/uy;
+const NULL_REGEX = /(?!)/uy; // zero-width negative lookahead, matches nothing
 
 const toCanonicalKeyword = (text: string) => equalizeWhitespace(text.toUpperCase());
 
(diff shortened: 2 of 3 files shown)


src/core/formatCommaPositions.ts
```diff
@@ -1,7 +1,7 @@
 import type { CommaPosition } from 'src/types';
 import { maxLength } from 'src/utils';
 
-import { WHITESPACE_REGEX } from './Tokenizer';
+const WHITESPACE_REGEX = /^(\s+)/u;
```
inferrinizzard, line 4: does this pattern also need the sticky flag ?

nene: Nope. This one I explicitly replaced to use a separate non-sticky regex, not import the sticky one from Tokenizer.

PR #246, opened 2022-06-21
Parenthesis parser

nene:
An initial attempt at building some sort of syntax tree so that the formatter could work with a bit higher level info than pure tokens.

At this first iteration I'm just parsing out groups of parenthesis. For example the following SQL:

```sql
SELECT (price + (price * tax)) FROM table;
```

produces the following AST:

```js
[
  {
    type: 'statement',
    children: [
      { type: 'token', token: { type: 'RESERVED_COMMAND', value: 'SELECT' } },
      { type: 'parenthesis', openParen: '(', closeParen: ')', children: [
        { type: 'token', token: { type: 'IDENT', value: 'price' } },
        { type: 'token', token: { type: 'OPERATOR', value: '+' } },
        { type: 'parenthesis', openParen: '(', closeParen: ')', children: [
          { type: 'token', token: { type: 'IDENT', value: 'price' } },
          { type: 'token', token: { type: 'OPERATOR', value: '*' } },
          { type: 'token', token: { type: 'IDENT', value: 'tax' } },
        ] },
      ] },
      { type: 'token', token: { type: 'OPERATOR', value: ';' } },
    ],
  },
]
```

As you see, I'm wrapping all plain tokens inside a special `type: 'token'` node. This is to avoid mixing Token types with AST node types. (Tried that approach initially and it created tons of problems, especially regarding types).

For now this is all very much work in progress...

### DIFF
--- a/src/core/ExpressionFormatter.ts
+++ b/src/core/ExpressionFormatter.ts
@@ -6,7 +6,7 @@ import InlineBlock from './InlineBlock';
 import Params from './Params';
 import { isReserved, isCommand, isToken, type Token, TokenType, EOF_TOKEN } from './token';
 import toTabularFormat from './tabularStyle';
-import { AstNode, isTokenNode, Parenthesis } from './Parser';
+import { AstNode, isTokenNode, Parenthesis } from './ast';
 import { indentString, isTabularStyle } from './config';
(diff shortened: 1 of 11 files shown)


nene:
As this developed into a larger rewrite of the whole parsing approach, closing this PR in favor of #250

PR #250, opened 2022-06-24
SQL Formatter 8.0

nene:
A new approach to formatting that relies more heavily on Parser rather than working on plain tokens.

To make it easier to work on this new approach I dropped some features (at least for now):

- [ ] newlineBeforeOpenParen / newlineBeforeCloseParen
- [ ] multilineLists
- [ ] indentStyle

All the lookAhead/Behind logic is now removed from the formatter logic and moved to Parser, which at the moment detects the following nodes:

- Statement - `;`-separated chunks of code.
- Clause - like `SELECT ... `, `WHERE ...` or `CREATE TABLE ...`
- BinaryClause - like `UNION` or `INTERSECT ALL`
- LimitClause - like `LIMIT 25, 5`
- FunctionCall - like `func(arg1, arg2)`
- ArraySubscript - like `my_array[10]`
- Parenthesis - matching pair of parenthesis (`(..)`, `[..]`, `{..}`) and everything between them
- BetweenPredicate - like `BETWEEN 5 AND 10`
- AllColumnsAsterisk - the `*` operator when used in `SELECT *`
- TokenNode - a wrapper for plain tokens. This is to avoid mixing Token types with AST node types. (Tried that approach initially and it created tons of problems in type system).

### DIFF
--- a/src/core/ExpressionFormatter.ts
+++ b/src/core/ExpressionFormatter.ts
@@ -5,7 +5,14 @@ import Indentation from './Indentation';
 import InlineBlock from './InlineBlock';
 import Params from './Params';
 import { isReserved, type Token, TokenType, EOF_TOKEN } from './token';
-import { AstNode, BetweenPredicate, isTokenNode, LimitClause, Parenthesis } from './ast';
+import {
+  AllColumnsAsterisk,
+  AstNode,
+  BetweenPredicate,
+  isTokenNode,
+  LimitClause,
+  Parenthesis,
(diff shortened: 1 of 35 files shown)


nene:
Yeah, I played around with Nearley. There's still the option to adopt it eventually.

At the moment I wanted to test how this whole parsing approach would play out.

Also, can't really use Nearly before switching the lexer to Moo or rewriting it to be Moo-compatible.

PR #252, opened 2022-06-28
8.0: Parsing of parenthesis

nene:
Just the parenthesis parsing portion, which works with all current config options.

Managed to get all the tests working, except for one.

PR #253, opened 2022-06-28
SQL Formatter 8.0

nene:
Main integration branch for the large parser changes in upcoming 8.0 version.

Splitted up to multiple sub PRs:

- [x] #252 
- [x] #254
- [x] #255 
- [x] #256
- [x] #257
- [x] #258
- [ ] #259
- [x] #260
- [x] #261
- [x] #262
- [ ] #263
- [x] #264
- [x] #270 
- [ ] #273


src/core/ExpressionFormatter.ts
```diff
@@ -0,0 +1,383 @@
...
+    this.params = params;
+    this.layout = layout;
+  }
+
+  public format(nodes: AstNode[]): Layout {
+    this.nodes = nodes;
+
+    for (this.index = 0; this.index < this.nodes.length; this.index++) {
+      const node = this.nodes[this.index];
+      switch (node.type) {
```
inferrinizzard, line 53: you can use the Enum in switch case with String()
ie. `switch (String(node.type)) { case: Enum.foo ... }`

src/core/ExpressionFormatter.ts
```diff
@@ -0,0 +1,383 @@
...
+        this.show(node.countToken),
+        WS.SPACE
+      );
+    } else {
+      this.layout.add(WS.NEWLINE, WS.INDENT, this.show(node.countToken), WS.SPACE);
+    }
+    this.layout.indentation.decreaseTopLevel();
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
inferrinizzard, line 177: can we remove this comment and rename the param to `_node` for now

src/core/ExpressionFormatter.ts
```diff
@@ -0,0 +1,383 @@
...
+
+  /** Formats a block comment onto query */
+  private formatBlockComment(token: Token) {
+    this.splitBlockComment(token.value).forEach(line => {
+      this.layout.add(WS.NEWLINE, WS.INDENT, line);
+    });
+    this.layout.add(WS.NEWLINE, WS.INDENT);
+  }
+
+  private splitBlockComment(comment: string): string[] {
```
inferrinizzard, line 245: add comment for this function

src/core/Formatter.ts
```diff
@@ -53,10 +56,27 @@ export default class Formatter {
 
   private formatAst(statements: Statement[]): string {
     return statements
-      .map(stat => new StatementFormatter(this.cfg, this.params).format(stat))
+      .map(stat => this.formatStatement(stat))
```
inferrinizzard, line 59: can be `.map(this.formatStatement)`

src/core/InlineBlock.ts
```diff
@@ -8,85 +9,68 @@ import { isToken, type Token, TokenType } from './token';
...
-    for (let i = index; i < tokens.length; i++) {
-      const token = tokens[i];
-      length += token.value.length;
+  private inlineWidth(parenthesis: Parenthesis): number {
+    let length = 2; // two parenthesis
 
-      if (this.isForbiddenToken(token)) {
-        return false;
+    for (const node of parenthesis.children) {
+      switch (node.type) {
```
inferrinizzard, line 26: same note as Enum above

src/core/InlineBlock.ts
```diff
@@ -8,85 +9,68 @@ import { isToken, type Token, TokenType } from './token';
...
       }
     }
-    return false;
+    return length;
+  }
+
+  private betweenWidth(node: BetweenPredicate): number {
+    return [node.betweenToken, node.expr1, node.andToken, node.expr2]
+      .map(token => token.value.length)
+      .reduce((a, b) => a + b);
```
inferrinizzard, line 65: replace with `.join('').length`

src/core/Layout.ts
```diff
@@ -5,24 +5,25 @@ import Indentation from './Indentation';
 /** Whitespace modifiers to be used with add() method */
 export enum WS {
   SPACE = 1, // Adds single space
```
inferrinizzard, line 7: numeric literals here are not required, already implicitly ints

inferrinizzard (approved):
Everything looks good so far, just some nitpicks for adding an Enum for the parser token types

src/core/Formatter.ts
nene: Won't work. `this` context is needed inside `formatStatement`. I could do:

```js
.map(this.formatStatement, this)
```

but that's more of a question with style. I would have preferred the separate `this` argument before arrow functions arrived to JavaScript, but nowadays IMHO the arrow-function is preferable as it will also work in functions that don't accapt a separate `this` parameter.

src/core/InlineBlock.ts
nene: I'd rather avoid string-concatenation for simply computing the length.

Opted to instead extract a `sum()` function to eliminate the reduce.

src/core/ExpressionFormatter.ts
nene: Not necessary, this works just fine:

```js
switch (node.type) {
    case: Enum.foo:
```

src/core/ExpressionFormatter.ts
nene: Good point. This can be confusing to understand. Added comment.

PR #254, opened 2022-06-28
8.0: Refactor StatementFormatter to ExpressionFormatter and add parser tests

nene:
Some cleanup tasks.

### DIFF
--- a/src/core/ast.ts
+++ b/src/core/ast.ts
@@ -1,0 +1,24 @@
+import { Token } from './token';
+
+export type Statement = {
+  type: 'statement';
+  children: AstNode[];
+};
+
+// Wrapper for plain nodes inside AST
+export type TokenNode = {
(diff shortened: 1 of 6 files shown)


src/core/ast.ts
```diff
@@ -0,0 +1,24 @@
+import { Token } from './token';
+
```
inferrinizzard, line 2: - [ ] rename this file as a typedef file

inferrinizzard (commented):
👍 ♻️

src/core/ast.ts
nene: I'd rather avoid changing file extensions based on the content. I also can't really see any benefit from having some files with `.d.ts` extension.

Currently there happens to be also a function inside here, so it can't become a `.d.ts` file.

PR #258, opened 2022-06-28
8.0: Detect SELECT * in Parser

nene:
Detection of the `*` operator when used in `SELECT *`. Now detected inside Parser instead having to do a lookbehind while formatting.


inferrinizzard:
does this account for languages like bigquery where you can use * after selecting additional columns?
ie. `SELECT foo, *, bar`

nene:
No, it does not. Also it doesn't account for languages which allow `SELECT tbl.*` syntax. But similarly the old version did not.

For now this is primarily moving the logic from formatter to parser without really improving it a whole lot.

PR #262, opened 2022-06-28
8.0: Refactor WhitespaceBuilder to avoid string-manipulation

nene:
Instead of using `WhitespaceBuilder.toString()` method to get the result of formatting a sub-expression (e.g. a Parenthesis block) and then doing string-manipulation to append it to current formatting state, the `WhitespaceBuilder` now exposes two new methods:

- `toLayout()` returns the internal representation of formatted SQL,
- `addLayout()` takes the output from the above method and appends it to current formatting context, adding indentation as needed to each line.

Additionally the `Statement` node no more contains semicolon token inside its children, but instead has a `hasSemicolon` field.

### DIFF
--- a/src/core/ExpressionFormatter.ts
+++ b/src/core/ExpressionFormatter.ts
@@ -17,7 +17,7 @@ import {
   Parenthesis,
 } from './ast';
 import { indentString } from './config';
-import WhitespaceBuilder, { WS } from './WhitespaceBuilder';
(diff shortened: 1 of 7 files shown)


nene:
Note: The main part of this refactor was pretty much reverted by #263 

These `toLayout()` and `addLayout()` methods are now gone. We do still avoid string manipulation by simply sharing the `WhitespaceBuilder` instance.

PR #263, opened 2022-06-28
8.0: Restore indentStyle support

nene:
Currently having some trouble restoring the original style.

Previously everything was just a stream of tokens, so you could indent things however you pleased, but now the following SQL:

```sql
SELECT *
FROM foo JOIN bar
```

gets parsed into structure like:

```js
[
  {type: "clause", name: "SELECT", children: ["*"]},
  {type: "clause", name: "FROM", children: ["foo", "JOIN", "bar"]},
]
```

and each array of children gets formatted in isolation, assuming that they are all indented 1 step more than the preceding `SELECT` or `FROM`. But that doesn't hold true for the formatting of `JOIN` in tabular style - there we would want to indent it at the same level as `SELECT`.

I see a few ways here:

- change the Parser to output JOIN blocks to the same level as clauses. Don't really like that. The AST shouldn't depend  on our formatting needs.
- change the formatting logic to eliminate this limit. Like instead of re-indenting the formatted sub-expression, passing the current indentation to the sub-expression formatter.

### DIFF
--- a/src/core/ExpressionFormatter.ts
+++ b/src/core/ExpressionFormatter.ts
@@ -253,7 +253,14 @@ export default class ExpressionFormatter {
   }
 
   private formatJoin(token: Token) {
-    this.query.add(WS.NEWLINE, WS.INDENT, this.show(token), WS.SPACE);
+    if (isTabularStyle(this.cfg)) {
+      // in tabular style JOINs are at the same level as clauses
+      this.query.indentation.decreaseTopLevel();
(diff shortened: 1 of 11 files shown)


nene:
Reorganized the use of `WhitespaceBuilder` so that now a single instance is shared throughout formatting of a single SQL statement. This allows changing the indentation as much as needed. Basically switching back to the original way we did formatting.

This change also brought back some of the quirks, like how comma gets placed after a line-comment. But those were more accidents of implementation in either way.

`indentStyle` now works just as it does in 7.0.

PR #265, opened 2022-06-29
Deprecate problematic config-options

nene:
Documented the problems with these three config options:

- multilineLists
- newlineBeforeOpenParen
- newlineBeforeCloseParen

Because of these issues, the confusion they create in users and the extra complexity they bring to the formatter (as I mentioned in #253) I propose deprecating these options and dropping them in 8.0 release.

A large part of the problem with these options is how they interact with `expressionWidth` option, which itself is problematic. What we really would like to have is a `lineWidth` option.

Also, I think it's more important to concentrate on fixing all the issues we have with our default format. All the extra options, especially the buggy options, make this a lot harder.


inferrinizzard (approved):
Yeah, I agree that these would be easier to re-implement with a proper parsing system so we can drop for now

PR #268, opened 2022-06-30
[VSCode] Drop prettier-sql from yarn.lock

nene:
Just ran `yarn install` and this was the result. Should not be needed any more.


inferrinizzard (approved):
👍

Issue #267, opened 2022-06-30
[FORMATTING] Issues with create view and joins - bigquery


m-ahmadkhan:
**Input data**
```sql
create or replace view `my_dataset.my_view` as (
  select t1.col1, t1.col2 from `my_dataset.my_table` as t1
  inner join `my_dataset.my_other_table` as t2 on t1.col1 = t2.col3 and t1.col2 = t2.col4
  where t1.col3 = "my_val" and t1.col4 = "my_val2"
);
```

Config used:
```json
{
  "language": "bigquery",
  "tabWidth": 2,
  "useTabs": false,
  "keywordCase": "upper",
  "indentStyle": "standard",
  "multilineLists": "always",
  "logicalOperatorNewline": "before",
  "aliasAs": "always",
  "tabulateAlias": false,
  "commaPosition": "after",
  "newlineBeforeOpenParen": false,
  "newlineBeforeCloseParen": true,
  "expressionWidth": 50,
  "linesBetweenQueries": 2,
  "denseOperators": false,
  "newlineBeforeSemicolon": false
}
```

**Expected Output**

```sql
CREATE OR REPLACE VIEW `my_dataset.my_view` AS (
  SELECT
    t1.col1,
    t1.col2
  FROM
    `my_dataset.my_table` AS t1
  INNER JOIN
    `my_dataset.my_other_table` AS t2
  ON
    t1.col1 = t2.col3
    AND t1.col2 = t2.col4
  WHERE
    t1.col3 = "my_val"
    AND t1.col4 = "my_val2"
);
```

**Actual Output**

```sql
create
OR REPLACE view `my_dataset.my_view` AS (
  SELECT
    t1.col1,
    t1.col2
  FROM
    `my_dataset.my_table` AS t1
    INNER JOIN `my_dataset.my_other_table` AS t2 ON t1.col1 = t2.col3
    AND t1.col2 = t2.col4
  WHERE
    t1.col3 = "my_val"
    AND t1.col4 = "my_val2"
);
```

There are a few issues with this output:

1. CREATE OR REPLACE VIEW:
   - It didn't convert `create` and `view` to uppercase, despite having the config `keywordCase` set to `upper`
   - `OR` in `CREATE OR REPLACE VIEW` should not be considered a logical operator, and be kept in a single line
2. INNER JOIN:
   - INNER JOIN keyword should have one less indent level
   - Table name for the join should go to a separate line (just like it handles the FROM clause)
[...]

PR #269, opened 2022-06-30
Support create table/view optional arguments

m-ahmadkhan:
Issue: https://github.com/sql-formatter-org/sql-formatter/issues/267

Create table syntax: https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#syntax_2
Create view syntax: https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#syntax_7


nene:
Thanks! That's a great start.

Will need to generalize this for all supported SQL dialects. There's likely quite a bit of variation as to what's supported.

If you're up for it, go ahead and add support for other dialects as well. If not, no problem. I'll take it over and do it by myself.

nene (approved):

nene:
This is now released in 7.0.4

PR #271, opened 2022-07-01
Upgrade dependencies

nene:
All seems to work. Encountered no problems.

Tested the new release-it with `--dry-run`. Had no problems. Also, looking at the [changelog](https://github.com/release-it/release-it/blob/master/CHANGELOG.md) none of the breaking changes in 15.0 should effect us.


inferrinizzard (approved):
👍

PR #272, opened 2022-07-01
More documentation for SQL syntax

nene:
More details on syntax differences.

### DIFF
--- a/sql/create-table.md
+++ b/sql/create-table.md
@@ -1,0 +1,76 @@
+# CREATE TABLE statement
+
+[SQL standard][] specifies the following CREATE TABLE syntax:
+
+    CREATE [{GLOBAL | LOCAL} TEMPORARY] TABLE
+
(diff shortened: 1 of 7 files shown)


nene:
> Can we also add CREATE DATABASE and CREATE SCHEMA to this PR too ?

For start my main goal is to cover the most basic and most-used things:

- SELECT
- INSERT
- DELETE
- UPDATE
- CREATE TABLE

Started with CREATE TABLE and CREATE VIEW mainly because of this PR: #269

- CREATE DATABASE is not in SQL standard and definitely not supported by many DB-s.
- CREATE SCHEMA is in standard but also not supported everywhere.

For start I'd rather concentrate on the most common things.

inferrinizzard:
oh yes, I thought this PR was going to be focused on CREATE statements but general docs for the most common clauses is good

inferrinizzard (approved):

PR #283, opened 2022-07-05
Refactor Tokenizer, update TokenTypes

inferrinizzard:
- split Tokenizer into `tokenizer.ts` main file and `tokenizerEngine.ts` where all of the regex matching logic is contained
- renamed `preprocess` to `postProcess`
- moved regex logic into lexer/ dir, split regex util and regex types off into separate files
- replace PARAMETER token type with NAMED_PARAMETER, QUOTED_PARAMETER, INDEXED_PARAMETER, POSITIONAL_PARAMETER
- renamed IDENT to IDENTIFIER
- added QUOTED_IDENTIFIER token type
- added COMMA token type
- add babel plugin to import sql files during jest tests


nene:
In general I don't really like this "repurposing" of an old branch. Especially half-a-year old branch. Like the first commit in here is from January, and it's completely irrelevant by now. Instead it would have been nicer to just discard the old branch and start a new, cherry-picking the changes that you want to keep, or even just copying the files over.

Like, I happen to know why this branch starts with Moo installation, but somebody else looking at this PR a year from now will have no idea what's going on.

Also as a result of this long history there are currently bunch of accidental changes in here that have nothing to do with this tokenizer refactor, but just came along from the Moo integration.

src/core/ExpressionFormatter.ts
```diff
@@ -208,11 +208,18 @@ export default class ExpressionFormatter {
...
         return this.formatCaseEnd(token);
-      case TokenType.PARAMETER:
+      case TokenType.NAMED_PARAMETER:
+      case TokenType.QUOTED_PARAMETER:
+      case TokenType.INDEXED_PARAMETER:
+      case TokenType.POSITIONAL_PARAMETER:
         return this.formatParameter(token);
+      case TokenType.COMMA:
+        return this.formatComma(token);
+      case TokenType.DELIMITER:
```
nene, line 218: This case never happens. All DELIMITER tokens gets consumed by parser and never reach ExpressionFormatter.

src/core/formatCommaPositions.ts
```diff
@@ -1,7 +1,7 @@
 import type { CommaPosition } from 'src/types';
 import { maxLength } from 'src/utils';
 
-const WHITESPACE_REGEX = /^(\s+)/u;
+import { WHITESPACE_REGEX } from 'src/lexer/regexUtil';
```
nene, line 4: The new regex is not the same as the old one.

- old: `/^(\s+)/u`
- new: `/(\s+)/uy`

test/sql/test.sql
```diff
@@ -0,0 +1,19 @@
...
+	FROM europe.greek.zeta
+	GROUP BY $eta, iota HAVING chi, psi, phi
+	UNION
+	SELECT MIN(rho)
+	FROM greek.omega
+)
+JOIN `greek_dev.sigma` ON tau = theta
+WHERE kappa AND lambda OR mu
+;
+SELECT upsilon AS y, omicron AS o FROM xi;
```
nene, line 19: What's the purpose of this file?

package.json
```diff
@@ -113,6 +113,7 @@
     "babel-jest": "^28.1.2",
     "babel-loader": "^8.1.0",
     "babel-plugin-add-module-exports": "^1.0.2",
+    "babel-plugin-inline-import": "^3.0.0",
```
nene, line 116: Seems unused.

src/languages/n1ql.formatter.ts
```diff
@@ -516,7 +516,7 @@ export default class N1qlFormatter extends Formatter {
       reservedJoins,
       reservedDependentClauses,
       reservedLogicalOperators: ['AND', 'OR', 'XOR'],
-      reservedKeywords: dedupe([...reservedKeywords, ...reservedFunctions]),
+      reservedKeywords: dedupe(reservedKeywords.concat(reservedFunctions)),
```
nene, line 519: As I commented in another PR, I see no need to replace spread operator with concat() calls.

src/core/Formatter.ts
```diff
@@ -45,7 +45,7 @@ export default class Formatter {
    * @return {string} The formatter query
    */
   public format(query: string): string {
-    const tokens = this.cachedTokenizer().tokenize(query);
+    const tokens = this.tokenizer().tokenize(query);
```
nene, line 48: Why not use the cached version?

nene (commented):
Found bunch of things that seem like leftovers from old changes in this branch.

nene (commented):
Found a few more issues.

src/core/ExpressionFormatter.ts
inferrinizzard: That what I thought but I left it in as a just it case to make sure it doesn't trigger the default case

src/languages/n1ql.formatter.ts
inferrinizzard: the concat is faster than spread for large lists, such as the keyword lists

src/utils.ts
```diff
@@ -26,6 +26,12 @@ export const equalizeWhitespace = (s: string) => s.replace(/\s+/gu, ' ');
 // identity function
 export const id = <T>(x: T) => x;
 
+// object deep comparison
+export const objectEquals = (x: any, y: any): boolean =>
```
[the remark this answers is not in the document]
inferrinizzard: was previously used for checking token output diff between moo/custom Tokenizer, left it in as a general util function

src/core/ExpressionFormatter.ts
nene: I think here we do want to trigger the default case in case any semicolon slips through, so we'd get a clear error as soon as possible instead of failing silently in some harder to diagnose way.

src/languages/n1ql.formatter.ts
nene: As in any case with performance, the only way to tell is do actually perform a measurement and see if there's a real difference.

src/utils.ts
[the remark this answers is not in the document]
nene: No reason to keep it then. Can always restore it from Git history if we do happen to need it again.

src/lexer/tokenizerEngine.ts
```diff
@@ -0,0 +1,160 @@
...
+      this.matchToken(TokenType.RESERVED_JOIN_CONDITION)
+    );
+  }
+
+  // Shorthand for `match` that looks up regex from REGEX_MAP
+  private matchToken(tokenType: TokenType): Token | undefined {
+    return this.match({
+      type: tokenType,
+      regex: this.REGEX_MAP[tokenType]!.regex,
+      transform: this.REGEX_MAP[tokenType]!.value,
```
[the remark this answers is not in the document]
inferrinizzard: typescript doesn't provide proper type narrowing for variable keys:
https://stackoverflow.com/questions/64616994/typescript-type-narrowing-not-working-for-in-when-key-is-stored-in-a-variable/64618261#64618261

src/languages/n1ql.formatter.ts
inferrinizzard: https://www.measurethat.net/Benchmarks/Show/579/1/arrayprototypeconcat-vs-spread-operator

src/lexer/tokenizerEngine.ts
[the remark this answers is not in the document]
nene: This is not really the same problem. Here we don't know what the value of `tokenType` is.

The root of the issue is that `REGEX_MAP` is defined as `Partial<>`, so it might contain all TokenTypes or it might contain none.

Even I, when reading this code, can't really tell for sure.

I see two solutions:

- Replace the use of `Partial<>` with the actual list of keys that are allowed.
- Check for existence of a key, like we're doing in `matchPlaceholderToken`.

inferrinizzard:
> Found bunch of things that seem like leftovers from old changes in this branch.

Most of the changes I would have done, they just happened to come up while the moo/tokenizer refactor was happening

src/lexer/tokenizerEngine.ts
```diff
@@ -0,0 +1,159 @@
...
+
+export default class TokenizerEngine {
+  private REGEX_MAP: Partial<Record<TokenType, TokenRule>>;
+
+  // The input SQL string to process
+  private input = '';
+  // Current position in string
+  private index = 0;
+
+  constructor(tokenizerOptions: Partial<Record<TokenType, TokenRule>>) {
```
nene, line 18: Should rename this parameter.

src/languages/n1ql.formatter.ts
inferrinizzard: sure I'll use spread

nene (approved):

inferrinizzard:
> Overall I think it's a good restructuring of the lexer logic. Better to merge this long-winded branch in before it manages to gather any more commits.

Agreed, will make any parser prep changes in another branch

PR #285, opened 2022-07-05
Split core/ directory into smaller sub directories

inferrinizzard:
broke up core/ into core/, formatter/ and parser/


nene:
I approve the reorganization. Makes sense.

But I'm not in favor of the file naming. I'd prefer following the principle of naming the file the same as the main class or function it exports. If it exports multitude of things, then sure, the name could be different.

Like currently you have renamed `Formatter.ts` to `formatter.ts`. (Although imports still mistakenly reference `Formatter.ts`). I'd prefer keeping the name as `Formatter.ts`.

inferrinizzard:
sure, I'll rename modules back to PascalCase instead

nene (approved):

PR #292, opened 2022-07-07
Remove core/ dir

nene:
As a continuation of recent code-reorganizations...

Now that we have `lexer/`, `formatter/` and `parser/` dirs there's not much of a reason for having a `core/` dir.


inferrinizzard:
I also considered this, had a few ideas regarding:
- Params, configs could be in a options/ dir but might be a big awkward with only 2 files and no plans of expansion there
- core _could_ be used to hold the other top level files but that isn't as clear

I think what you have here also makes sense, the only suggestion I would make is to move token.ts to top level since it's used in every context

inferrinizzard (approved):

PR #294, opened 2022-07-07
Remove !assertions and other small refactors

nene:
Eliminated use of typescript non-null assertions and added ESLint rule to avoid them in to future. Or at least to make it easier to spot these cases, as the `!` character is such a small thing to notice when reviewing code.

Plus a few additional little changes.

### DIFF
--- a/src/lexer/TokenizerEngine.ts
+++ b/src/lexer/TokenizerEngine.ts
@@ -8,15 +8,15 @@ export interface TokenRule {
 }
 
 export default class TokenizerEngine {
-  private REGEX_MAP: Partial<Record<TokenType, TokenRule>>;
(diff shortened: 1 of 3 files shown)


src/lexer/TokenizerEngine.ts
```diff
@@ -119,15 +119,16 @@ export default class TokenizerEngine {
     );
   }
 
-  // Shorthand for `match` that looks up regex from REGEX_MAP
+  // Shorthand for `match` that looks up regex from rules
   private matchToken(tokenType: TokenType): Token | undefined {
-    if (!(tokenType in this.REGEX_MAP)) {
```
inferrinizzard, line 124: unfortunate that the `in` operator doesn't provide type narrowing with variables

PR #126, opened 2021-03-05
Add named argument operators to posgresql

briandunn:
Prevent munging of [postgresql named argument operators](https://www.postgresql.org/docs/12/sql-syntax-calling-funcs.html#SQL-SYNTAX-CALLING-FUNCS-NAMED).

Resolves #125.

PR #296, opened 2022-07-08
Update list of operators for all languages

### LINKED ISSUE #291: Parse error when using Postgres JSON syntax
**Describe the bug**
A `Parsing Error` occurs for the following query (language `PostgreSQL`):
```sql
SELECT id, name FROM tabe WHERE jsoncol @? 'strict $.** ? (@.id == 2 )'
```

**Expected behavior**
Should format the query

**Actual behavior**
```
Parse error: Unexpected "? 'strict $.** ? (@.id == 2 )'"

Stack Trace:

Error: Parse error: Unexpected "? 'strict $.** ? (@.id == 2 )'"
    at E.value (https://unpkg.com/sql-formatter@latest/dist/sql-formatter.min.js:1:37270)
    at N.value (https://unpkg.com/sql-formatter@latest/dist/sql-formatter.min.js:1:29851)
    at Module.OS (https://unpkg.com/sql-formatter@latest/dist/sql-formatter.min.js:1:208492)
    at HTMLTextAreaElement.format (https://sql-formatter-org.github.io/sql-formatter/index.js:5
```

**Usage**

  - this can be reproduced in the [online-formatter](https://sql-formatter-org.github.io/sql-formatter/)
  - set the for language to `PostgreSQL`
  -  8.0.2

inferrinizzard:
- resolves #291 
- updated operators for all languages if not already comprehensive
- added missing operators for Postgres

### DIFF
--- a/src/languages/postgresql.formatter.ts
+++ b/src/languages/postgresql.formatter.ts
@@ -1649,6 +1649,7 @@ const binaryOperators = [
   // Geometric operators
   '@-@',
   '@@',
+  '#',
   '##',
   '<->',
   '&&',
@@ -1674,6 +1675,7 @@ const binaryOperators = [
   // Text Search Operators
   '@@@',
   // JSON Operators
+  '?',
   '@?',
   '?&',
   '->',
@@ -1683,10 +1685,9 @@ const binaryOperators = [
   '#-',
   // Other Operators
   ':=',
-  // Unknown
   '::',
   '=>',
-  '=#',
+  '-|-',
 ];
 
 // https://www.postgresql.org/docs/14/index.html
--- a/src/languages/postgresql.formatter.ts
+++ b/src/languages/postgresql.formatter.ts
@@ -1618,26 +1618,35 @@ const reservedJoins = [
 const reservedDependentClauses = ['WHEN', 'ELSE'];
 
 const binaryOperators = [
+  // Math Operators
   '<<',
   '>>',
-  '||/',
   '|/',
-  '::',
-  ':=',
-  '->>',
-  '->',
-  '#>>',
-  '#>',
-  '=>',
-  '~~*',
-  '~~',
-  '!~~*',
-  '!~~',
-  '~*',
-  '!~*',
-  '!~',
+  '||/',
   '!!',
+  // String Operators
   '||',
+  // Pattern Matching Operators
+  '~~',
+  '~~*',
+  '!~~',
+  '!~~*',
+  // POSIX RegExp operators
+  '~',
+  '~*',
+  '!~',
+  '!~*',
+  // Similarity Operators
+  '<%',
+  '<<%',
+  '%>',
+  '%>>',
+  // Byte Comparison Operators
+  '~>~',
+  '~<~',
+  '~>=~',
+  '~<=~',
+  // Geometric operators
   '@-@',
   '@@',
   '##',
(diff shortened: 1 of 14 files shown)


inferrinizzard:
the `::` operator is some languages is odd since it's not used with the same spacing, ie. it's usually foo::bar instead of foo :: bar

nene:
`::` is a type cast operator. We definitely shouldn't make it part of an identifier. Just like with the `.` operator we can simply format it differently from other operators.

nene:
In Transact-SQL `::` is a scope-resolution operator: https://docs.microsoft.com/en-us/sql/t-sql/language-elements/scope-resolution-operator-transact-sql?view=sql-server-ver16

src/languages/postgresql.formatter.ts
```diff
@@ -1618,26 +1618,35 @@ const reservedJoins = [
...
+  // String Operators
+  '||',
+  // Pattern Matching Operators
   '~~',
-  '!~~*',
+  '~~*',
   '!~~',
+  '!~~*',
+  // POSIX RegExp operators
+  '~',
```
nene, line 1635: Should move `~` to Tokenizer where all other single-letter operators are defined:

```ts
[TokenType.OPERATOR]: {
  regex: regex.operator('+-/*%&|^><=.[]{}`:$@', [
```

Though, looking at this list we should definitely remove the `` ` `` from it, which is never an operator on its own, but instead part of quotes.

Likely should also drop `[]{}` as these are covered by parenthesis patterns. Though perhaps better to leave them for now, as these serve as a fallback in case we have missed to include them to `open/closeParens`. I attempted to [document this syntax](https://github.com/sql-formatter-org/sql-formatter/blob/master/sql/arrays-and-maps.md), but soon found out that I had missed Postgres and BigQuery. Likely there are more. Need to do further investigation.

There are also some omissions in this list of single-letter operators:

- `#` - Postgres bitwise exclusive OR (also a geometric operator)
- `?` - Postgres jsonb operator
- `!` - Don't know of such an operator, but as it can be part of an operator, it's a good fallback to have.

src/languages/postgresql.formatter.ts
```diff
@@ -1659,9 +1668,25 @@ const binaryOperators = [
...
+  '->>',
+  '#>',
+  '#>>',
+  '#-',
+  // Other Operators
+  ':=',
+  // Unknown
+  '::',
+  '=>',
+  '=#',
```
nene, line 1689: So, what's up with these, and what's the difference between "Other" and "Unknown"?

- `:=` is [assignment](https://www.postgresql.org/docs/current/plpgsql-statements.html#PLPGSQL-STATEMENTS-ASSIGNMENT) or for named arguments
- `::` is a type cast
- `=>` is for named function arguments #126  
- `=#` well, you just added it, so you better know what it's for :)

src/languages/tsql.formatter.ts
```diff
@@ -1227,7 +1227,20 @@ const reservedDependentClauses = ['WHEN', 'ELSE'];
...
+    '!>',
+    '+=',
+    '-=',
+    '*=',
+    '/=',
+    '%=',
+    '|=',
+    '&=',
+    '^=',
+    // '::'
```
nene, line 1242: There sure is such an operator: https://docs.microsoft.com/en-us/sql/t-sql/language-elements/scope-resolution-operator-transact-sql?view=sql-server-ver16

nene (approved):

PR #299, opened 2022-07-09
Document SQL query syntax

nene:
Most of the basic stuff should now be covered.


nene:
No code to review here, only changing Markdown files, so merging this straight in.

PR #300, opened 2022-07-10
Fix formatting of ORDER BY for MariaDB

### LINKED ISSUE #298: [FORMATTING] MariaDB formats sample query on demo page differently
**Input data**
```sql
select supplier_name,city from
(select * from suppliers join addresses on suppliers.address_id=addresses.id)
as suppliers
where supplier_id>500
order by supplier_name asc,city desc;
```

with the default settings on the website.

**Expected Output**
Every other formatter formats it like this:

```sql
select
  supplier_name,
  city
from
  (
    select
      *
    from
      suppliers
      join addresses on suppliers.address_id = addresses.id
  ) as suppliers
where
  supplier_id > 500
order by
  supplier_name asc,
  city desc;
```

**Actual Output**

The MariaDB formatter formats it like this:

```sql
select
  supplier_name,
  city
from
  (
    select
      *
    from
      suppliers
      join addresses on suppliers.address_id = addresses.id
  ) as suppliers
where
  supplier_id > 500
order by
  supplier_name asc,
  city
desc
;
```

**Usage**

https://sql-formatter-org.github.io/sql-formatter/
MariaDB
`latest`

This happens because in MariaDB, `DESC` is short for `DESCRIBE` so it appears in `reservedCommands` instead of `reservedKeywords` 

https://github.com/sql-formatter-org/sql-formatter/blob/76fb498acbc0e479eaeb3fb299045eaa8b619d3d/src/languages/mariadb.formatter.ts#L939

nene:
Also add a test for ORDER BY to avoid such regression in the future.

Fixes #298


nene:
A trivial fix. Merging straight in.

PR #304, opened 2022-07-11
Fix parsing and formatting of LIMIT clause

### LINKED ISSUE #301: [FORMATTING] Comments after LIMIT keyword
**Input data**
options = { language: 'postgresql' };

**Expected Output**

```sql
LIMIT
  --limit special in this case as there is sensitivity on max
  LEAST(100, COALESCE('__limit', 20));
```

**Actual Output**

```sql
LIMIT
  --limit special in these case as there is sensitivity on max LEAST(100, COALESCE('__limit', 20));
```

**Usage**

calling inside a Jest snapshot-serializer
postgresql
The issue appeared after upgrading to v8.0.2 (from v7.0.4)

### LINKED ISSUE #303: Feature Request: Remove space after LEAST and GREATEST
Hi there, the latest version 8 is much improved thank you.
(I raised 2 issues about LIMIT as they are different bugs.)

One more tiny thing... v8 started adding a space between `LEAST` and `(` in Postgres fetches.
eg:
`LEAST(` becomes `LEAST (` 

This also affects...
`GREATEST(` becomes `GREATEST (`

Not sure this was intentional?

Thanks for considering.

nene:
This serves as a case study of how easy it is to get things wrong when parsing even very simple SQL expressions.

Fixes #301
Fixes #303

### DIFF
--- a/src/parser/Parser.ts
+++ b/src/parser/Parser.ts
@@ -162,20 +162,25 @@ export default class Parser {
   }
 
   private limitClause(): LimitClause | undefined {
-    if (isToken.LIMIT(this.look()) && this.look(2).type === TokenType.COMMA) {
-      return {
-        type: NodeType.limit_clause,
-        limitToken: this.next(),
-        offset: [this.nextTokenNode()],
-        count: this.next() && [this.nextTokenNode()], // Discard comma token
-      };
-    }
     if (isToken.LIMIT(this.look())) {
-      return {
-        type: NodeType.limit_clause,
-        limitToken: this.next(),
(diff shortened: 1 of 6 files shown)


src/parser/Parser.ts
```diff
@@ -188,6 +175,23 @@ export default class Parser {
     return undefined;
   }
 
+  private expressionsUntilClauseEnd(
```
inferrinizzard, line 178: add function comments here

inferrinizzard (approved):

Issue #243, opened 2022-06-20
[FORMATTING] Inconsistent placement of offset index in select - bigquery


m-ahmadkhan:
**Input data**

I'm trying to format the following BigQuery query:
```sql
SELECT CAST(SPLIT(
        key1.path, ',')[OFFSET(1)] AS INT64) AS col1,
    CAST(SPLIT(key2.path, ',')[OFFSET(1)] AS INT64) AS col2,
    key3_looooooooooooooong_name.id AS col3
FROM `my_dataset.my_table`;
```

With the formatting options:
```
{
  "language": "bigquery",
  "tabWidth": 2,
  "useTabs": false,
  "keywordCase": "upper",
  "indentStyle": "standard",
  "multilineLists": 4,
  "logicalOperatorNewline": "before",
  "aliasAs": "always",
  "tabulateAlias": false,
  "commaPosition": "after",
  "newlineBeforeOpenParen": true,
  "newlineBeforeCloseParen": true,
  "expressionWidth": 50,
  "linesBetweenQueries": 2,
  "denseOperators": false,
  "newlineBeforeSemicolon": false
}
```

**Actual Output**

At the first format, it generates the query (notice the `(1)` on newlines with no spaces before them):
```sql
SELECT CAST(
    SPLIT(key1.path, ',')[
      OFFSET
(1)
    ] AS INT64
  ) AS col1,
  CAST(
    SPLIT(key2.path, ',')[
      OFFSET
(1)
    ] AS INT64
  ) AS col2,
  key3_looooooooooooooong_name.id AS col3
FROM `my_dataset.my_table`;
```

And on formatting again, it adds spaces before the `(1)`:
```sql
SELECT CAST(
    SPLIT(key1.path, ',')[
      OFFSET
        (1)
    ] AS INT64
  ) AS col1,
  CAST(
    SPLIT(key2.path, ',')[
      OFFSET
        (1)
    ] AS INT64
  ) AS col2,
  key3_looooooooooooooong_name.id AS col3
FROM `my_dataset.my_table`;
```

**Small changes to the query**

If I just make a little change in the query to move `key3_looooooooooooooong_name.id AS col3` at the first place in the select statement. It would generate a different result.

```sql
SELECT key3_looooooooooooooong_name.id AS col3,
    CAST(SPLIT(
        key1.path, ',')[OFFSET(1)] AS INT64) AS col1,
[...]

Issue #140, opened 2021-04-19
The results of the two operations are different when i use standarSqlFormat，the indent before '(' are not the same


a82448186:
I have a sql

```
insert into user(name, age) values('text', 11)
```
the first time i format it, i get the result
```
insert into
  user(name, age)
values
('text', 11)
```
Based on the last result，Continue to format，i get the result
```
insert into
  user(name, age)
values
  ('text', 11)
```

i expect the same sql get the same result just like the second
the version i used was 4.0.2

nene:
Problem still present in latest master. Likely related to `values(` looking like a function and getting formatted differently from `values (`.

nene:
This is now finally fixed in 9.0.0-beta1

Issue #237, opened 2022-06-13
Allow uppercasing of function names separately from keywords


nene:
Currently we have a single option `keywordCase` which specifies the case for both keywords and function names.

It would be nice to allow more fine-grained control by having a separate option `functionCase` to control whether function names should be uppercased or lowercased. It's quite common practice when writing SQL to have different case-convention for the keywords and function names. Most commonly uppercasing keywords and lowercasing function names, like:

```sql
SELECT
  round(price * abs(tax)) AS final_price
FROM
  inventory;
```

This feature is also widespread in many other SQL formatting tools:

- [SQLFormat](https://sqlformat.org/) has options for keyword and identifier case
- [Instant SQL Formatter](https://www.dpriver.com/pp/sqlformat.htm) has 9 options for various language elements.
- [Freeformatter.com](https://www.freeformatter.com/sql-formatter.html) has options for keyword and identifier case
- [SQL Complete](https://sql-format.com/) has options for separate casing of alias, function, data-type, identifier, keyword, variable.

nene:
The initial most obvious solution I had in my head was to distinguish them solely by name, but this won't work as some function name are also keywords used elsewhere in SQL. For example:

- `SET()` function and `UPDATE table SET ...`
- `CAST(1 AS INT)` and `RETURNS INT CAST FROM FLOAT`

Luckily these are fairly rare cases. But still.

Another option would be to look whether function name is followed by open-parenthesis `(`. This seems more promising, though one has to keep in mind that not every name followed by parenthesis is a function.

inferrinizzard:
There is also the reverse case: some functions do not require parentheses: ie. `CURRENT_DATE, TIMESTAMP, etc.`

jacobsvante:
`functionCase` would be a great addition to this already awesome project! It just looks strange to me too see PostgreSQL functions in uppercase.

karlhorky:
Ah this is also closed by my PR for `dataTypeCase` + `functionCase`:

- #673

PR #307, opened 2022-07-12
New approach to function call matching

nene:
Previously the approach to detecting function calls was as follows:

- Inside Tokenizer: save preceding whitespace of each token to `whitespaceBefore` field
- Inside Parser: check if `RESERVED_KEYWORD` or `IDENTIFIER` is followed by `(` and there's no whitespace between them.

This has been a great hack to format function calls without really having a knowledge of which words are function names. But this approach has the fundamental downside of it relying on the whitespace of original code to make formatting decisions, which leads to surprising behavior (see #140 and #243) where re-formatting an already formatted SQL produces a different result.

The new approach introduced in this PR goes as follows:

- Inside Tokenizer: label all function names as `type: RESERVED_FUNCTION_NAME` tokens.
- Inside Parser: check if `RESERVED_FUNCTION_NAME` is followed by `(`.

This is mostly how one would expect this to work. However there are various caveats:

- A common convention is to format data types similarly to function calls e.g. `DECIMAL(5, 2)`. For now this is solved by also including data types (that can take parameters) also to the functions list. In the future should completely separate data types from functions.
- `CAST` is used in various forms, but as its main use is in `CAST(expr AS type)`, it's for now listed alongside functions.
- `ANY` and `SOME` can be used as aggregate functions, in which case they should be formatted as `HAVING ANY(condition)`. But a way more common usage is as an operator: `WHERE x = ANY (1, 2, 3)`. So I've opted to remove these from list of functions.
- `NULLIF` and `COALESCE` have sometimes been listed as keywords, at other times as function names. Changed them to be function names in every dialect. That really hints of a general problem that these keyword/function-name lists need a proper review.

The major downside of this new approach is, that:

- calls of a custom function like `foo(1, 2)` are now formatted as `foo (1, 2)`.

But there's also a major benefit from parsing side:

- this change should greatly simplify switching to a parser-generator, as the parser won't have to look at whitespace (e.g. Nearley or Jison can't really distinguish between a token with and without preceding whitespace. Unless we go and introduce separate whitespace tokens, it's a major blocker for adopting a parser generator).

So, in the end it's about tradeoffs. IMHO the sacrifices are worth the gains.


test/features/case.ts
```diff
@@ -20,12 +20,12 @@ export default function supportsCase(format: FormatFn) {
 
   it('formats CASE ... WHEN with an expression', () => {
     const result = format(
-      "CASE toString(getNumber()) WHEN 'one' THEN 1 WHEN 'two' THEN 2 WHEN 'three' THEN 3 ELSE 4 END;"
+      "CASE trim(sqrt(2)) WHEN 'one' THEN 1 WHEN 'two' THEN 2 WHEN 'three' THEN 3 ELSE 4 END;"
```
nene, line 23: Here and in several other tests I've switched over to using the some widely adopted function names, so the tests would work across all dialects.

test/behavesLikeSqlFormatter.ts
```diff
@@ -195,18 +195,18 @@ export default function behavesLikeSqlFormatter(format: FormatFn) {
...
     `);
   });
 
   it('breaks long parenthesized lists to multiple lines', () => {
     const result = format(`
       INSERT INTO some_table (id_product, id_shop, id_currency, id_country, id_registration) (
-      SELECT IF(dq.id_discounter_shopping = 2, dq.value, dq.value / 100),
-      IF (dq.id_discounter_shopping = 2, 'amount', 'percentage') FROM foo);
+      SELECT COALESCE(dq.id_discounter_shopping = 2, dq.value, dq.value / 100),
+      COALESCE (dq.id_discounter_shopping = 2, 'amount', 'percentage') FROM foo);
```
nene, line 209: Switched from `IF` to `COALESCE` which is supported in all dialects.

test/plsql.test.ts
```diff
@@ -164,7 +165,7 @@ describe('PlSqlFormatter', () => {
     `);
     expect(result).toBe(dedent`
       WITH
-        t1(id, parent_id) AS (
+        t1 (id, parent_id) AS (
```
nene, line 168: Not really sure which way is better here. It's not a function call.

However it should be possible to improve this with special parsing of WITH clause in the future.

test/bigquery.test.ts
```diff
@@ -114,10 +114,11 @@ describe('BigQueryFormatter', () => {
     `);
   });
 
+  // TODO: Possibly incorrect formatting of STRUCT<>() and ARRAY<>()
```
nene, line 117: This is currently an unfortunate change in formatting.

This whole typed arrays/structs parsing needs a complete rewrite though. The current solution is a big hack. It should really be parsed inside Parser not hacked together inside postProcess() of tokenizer.

src/languages/sqlite.formatter.ts
```diff
@@ -52,6 +52,89 @@ const reservedFunctions = {
...
+    'JSON_SET',
+    'JSON_TYPE',
+    'JSON_TYPE',
+    'JSON_VALID',
+    'JSON_QUOTE',
+    'JSON_GROUP_ARRAY',
+    'JSON_GROUP_OBJECT',
+    'JSON_EACH',
+    'JSON_TREE',
+  ],
```
nene, line 123: Found lots of SQLite functions that we were missing.

nene:
This change also paves the way for implementing separate uppercasing option for function names: #237

test/behavesLikeSqlFormatter.ts
inferrinizzard: interesting that IF isn't as universal as COALESCE

inferrinizzard (approved):
I think this overall makes sense but agree that the keyword lists need to be reviewed 👍

PR #315, opened 2022-07-14
Reorganize languages/ directory

nene:
Moved each `*.formatter.ts` file to separate directory and moved keywords and functions lists to separate files.

Moving these huge lists out should make it simpler dealing with these formatter files. Also I've often been scrolling up'n'down in an attempt of finding the keywords and functions lists, sometimes thinking I'm looking at functions, when in fact it's the keywords list. Should be better with less scrolling now.


nene:
Well, it doesn't actually conflict the addition of a new dialect into the old directory structure. Plus I'd rather merge this sooner than later, as I'd like to do some further changes on top of it.

Issue #247, opened 2022-06-23
Feature Request: Add "OR" to logicalOperatorNewline


rigogsilva:
**Describe the Feature**
Is there a region so not add OR to the logicalOperatorNewline? Or maybe is there a different way to control that? 

**Why do you want this feature?**
Normally in sql the ON is not on the same line as the join itself. For example: 

```sql
SELECT
  *
FROM
  foo
  JOIN foo 
    ON foo.account_id = bar.account_id
    AND foo.something = bar.something
  AND a = b
```

This looks a bit weird: 

```sql
SELECT
  *
FROM
  foo
  JOIN foo ON foo.account_id = bar.account_id
    AND foo.something = bar.something
  AND a = b
```

And it is nice to look at the ON and AND clause without having to move your eye to the right. Helps readability.

nene:
With the latest version (7.0.2) I actually get this result:

```sql
SELECT
  *
FROM
  foo
  JOIN foo ON foo.account_id = bar.account_id
  AND foo.something = bar.something
  AND a = b
```

But I agree that the current formatting of `ON` expression leaves it looking quite confusing. Ideally both of these `AND`-operator lines should be indented. I would consider both of these to be acceptable:

```sql
FROM
  foo
  JOIN foo ON foo.account_id = bar.account_id
    AND foo.something = bar.something
    AND a = b
-- or
FROM
  foo
  JOIN foo ON
    foo.account_id = bar.account_id
    AND foo.something = bar.something
    AND a = b
```

Though in my experience joining over multiple fields is not something you do often. You can get a much better result from the formatter if you use parenthesis:

```sql
SELECT
  *
FROM
  foo
  JOIN foo ON (
    foo.account_id = bar.account_id
    AND foo.something = bar.something
    AND a = b
  )
```

With the current architecture this is pretty difficult to fix. More likely after a larger rewrite of the parser that's currently in progress.

rigogsilva:
Yeah I was thinking something like this also: 

```sql
FROM
  foo
  JOIN foo 
    ON foo.account_id = bar.account_id
    AND foo.something = bar.something
    AND a = b
```

Which is closer to what I have seen in sql formatting. 

Not a problem. Thanks for considering. I was able to handle it with post processing witch is not ideal but works: 

```javascript
  const onFormattedQuery = (
    sqlFormatterQuery.toString().split("\n")
    .map((line: string) => {
      if (line.includes(" ON ")) {
[...]

PR #317, opened 2022-07-14
Remove RESERVED_JOIN_CONDITION token type

nene:
The ordering of this was pretty confusing in the TokenizerEngine:

```ts
      this.matchToken(TokenType.RESERVED_KEYWORD) ||
      this.matchToken(TokenType.RESERVED_LOGICAL_OPERATOR) ||
      this.matchToken(TokenType.RESERVED_JOIN_CONDITION)
```

with `ON` and `USING` coming after all other keywords for some reason. The reason being that keywords actually contains some not-so-plain keywords like `ON UPDATE` and `ON DELETE` which we want to detect before a plain `ON` token.

There also wasn't any special logic for formatting of these tokens.

I separated these special `ON UPDATE` and `ON DELETE` from other "normal" keywords. Really planning to move them out of the keywords lists entirely.

But for now the order of matching RESERVED_* tokens should make more sense, with the most generic keywords being detected as the last resort.

### DIFF
--- a/src/lexer/TokenizerEngine.ts
+++ b/src/lexer/TokenizerEngine.ts
@@ -110,8 +110,8 @@ export default class TokenizerEngine {
       this.matchToken(TokenType.RESERVED_BINARY_COMMAND) ||
       this.matchToken(TokenType.RESERVED_DEPENDENT_CLAUSE) ||
       this.matchToken(TokenType.RESERVED_JOIN) ||
-      this.matchToken(TokenType.RESERVED_FUNCTION_NAME) ||
       this.matchToken(TokenType.RESERVED_LOGICAL_OPERATOR) ||
+      this.matchToken(TokenType.RESERVED_FUNCTION_NAME) ||
       this.matchToken(TokenType.RESERVED_KEYWORD)
     );
   }


inferrinizzard:
With this revert, would the default behaviour be to inline the ON and USING keywords for JOINs ?
ie.
```sql
SELECT *
FROM tbl1
JOIN tbl2 ON tbl1.foo = tbl2.foo
```

Not sure if I agree with that being the default.
I've always seen the ON broken to a newline and sometimes indented, which is why I initial placed it as a `RESERVED_DEPENDENT_CLAUSE` instead of a standard keyword to produce output along the lines of
```sql
SELECT *
FROM tbl1
JOIN tbl2
ON tbl1.foo = tbl2.foo
```
or 
```sql
SELECT *
FROM tbl1
JOIN tbl2
    ON tbl1.foo = tbl2.foo
```

nene:
> With this revert, would the default behaviour be to inline the ON and USING keywords for JOINs ?

This has always been the default in SQL Formatter. This PR does not change that behavior - as can be seen from there being no changes to tests.

You indeed changed that behavior in Prettier SQL and during the big merge I reverted it back to how it was before to restore backwards compatibility with the old formatting.

There's a feature request for different formatting of `ON` keyword in `JOIN` clauses: #247. I suggest to discuss that further under that issue.

The problem is more complex than simply the placement of `ON`. Rather one needs to think how to format `JOIN table ON condition AND condition AND condition`.

inferrinizzard (approved):
Approved, DEPENDENT_CLAUSE formatting discussion to come later

PR #324, opened 2022-07-15
Don't test for JOIN ... USING () support in dialects not supporting this syntax

nene:
I was looking into cleaning up reserved keyword lists and discovered that we're running tests for JOIN ... USING syntax even for dialects not supporting it: DB2, Hive, N1QL and Transact-SQL.


test/features/join.ts
```diff
@@ -2,9 +2,16 @@ import dedent from 'dedent-js';
 
 import { FormatFn } from 'src/sqlFormatter';
 
-type Options = { without?: string[]; additionally?: string[] };
+type Options = {
```
inferrinizzard, line 5: this should be interface

inferrinizzard (approved):

PR #326, opened 2022-07-16
Document Trino syntax for SELECT clauses

nene:
Document the missing parts of query syntax:

- WITH
- GROUP BY
- ORDER BY
- OFFSET..LIMIT & OFFSET..FETCH

SELECT and FROM were already covered.

WHERE and HAVING clauses are the same as in all dialects.


nene:
Merging straight in as these are only internal documentation changes.

PR #327, opened 2022-07-16
Utility for auto-generation of syntax variants

nene:
Listing all these syntax variations (like currently done for BigQuery) has grown a bit overboard:

```js
  'CREATE FUNCTION',
  'CREATE OR REPLACE FUNCTION',
  'CREATE FUNCTION IF NOT EXISTS',
  'CREATE TEMP FUNCTION',
  'CREATE OR REPLACE TEMP FUNCTION',
  'CREATE TEMP FUNCTION IF NOT EXISTS',
  'CREATE TEMPORARY FUNCTION',
  'CREATE OR REPLACE TEMPORARY FUNCTION',
  'CREATE TEMPORARY FUNCTION IF NOT EXISTS',
```

Wrote a helper function to generate all the above variants automatically from syntax like this:

```js
  'CREATE [OR REPLACE] [TEMP|TEMPORARY] FUNCTION [IF NOT EXISTS]',
```

It's a very simplistic utility, but at least it cuts the amount of code down drastically. The resulting list is also more readable with all this repetitive noise eliminated.

Note: the above expression also produces some invalid variants like `CREATE OR REPLACE FUNCTION IF NOT EXISTS`, but IMHO that's a very small overhead not worth worrying about. The gains in readability should win here by a lot.


src/expandPhrases.ts
```diff
@@ -0,0 +1,67 @@
...
+type PhrasePart = string[];
+
+const REQUIRED_PART = /[^[]*/y;
+const OPTIONAL_PART = /\[.*?\]/y;
+
+const parsePhrase = (text: string): PhrasePart[] => {
+  let index = 0;
+  const result: PhrasePart[] = [];
+  while (index < text.length) {
+    REQUIRED_PART.lastIndex = index;
```
inferrinizzard, line 39: add comments here for required part / optional part logic

src/expandPhrases.ts
```diff
@@ -0,0 +1,67 @@
...
+    OPTIONAL_PART.lastIndex = index;
+    const optionalMatch = OPTIONAL_PART.exec(text);
+    if (optionalMatch) {
+      const choices = optionalMatch[0]
+        .slice(1, -1)
+        .split('|')
+        .map(s => s.trim());
+      result.push(['', ...choices]);
+      index += optionalMatch[0].length;
+    }
```
inferrinizzard, line 55: can we also throw an error here in case neither regex matches ?

inferrinizzard (approved):

src/expandPhrases.ts
```diff
@@ -0,0 +1,67 @@
...
+  }
+  return result;
+};
+
+const buildCombinations = ([first, ...rest]: PhrasePart[]): string[] => {
```
inferrinizzard, line 60: very clever permutation exploder

src/expandPhrases.ts
nene: Good point. One might forget to close a parenthesis and then we'd enter an infinite loop.

src/languages/bigquery/bigquery.formatter.ts
```diff
@@ -20,122 +21,58 @@ const reservedCommands = [
...
-  'CREATE FUNCTION',
-  'CREATE OR REPLACE FUNCTION',
-  'CREATE FUNCTION IF NOT EXISTS',
-  'CREATE TEMP FUNCTION',
-  'CREATE OR REPLACE TEMP FUNCTION',
-  'CREATE TEMP FUNCTION IF NOT EXISTS',
-  'CREATE TEMPORARY FUNCTION',
-  'CREATE OR REPLACE TEMPORARY FUNCTION',
-  'CREATE TEMPORARY FUNCTION IF NOT EXISTS',
+  'CREATE [OR REPLACE] [TEMP|TEMPORARY|TABLE] FUNCTION [IF NOT EXISTS]',
```
m-ahmadkhan, line 40: This will also generate `CREATE OR REPLACE FUNCTION IF NOT EXISTS`, but `OR REPLACE` and `IF NOT EXISTS` can't co-exist.

nene: Yep, I know. As mentioned in PR description.

It doesn't really matter a lot that the formatter is capable of formatting some incorrect syntax. I'm more concerned about memory overhead from generating these extra impossible variants. But I think that's a minuscule loss compared to the win in readability.

PR #328, opened 2022-07-17
Move SQL docs to wiki

nene:
Moved all these SQL syntax docs to [Wiki](https://github.com/sql-formatter-org/sql-formatter/wiki).

Simpler to edit and browse the documentation there.


inferrinizzard:
is there a way to store the wiki source as part of the repo somehow?

I think it would still be good to have it track changes to the wiki over time

nene:
The wiki is itself a git repo. You can just:

```
git clone git@github.com:sql-formatter-org/sql-formatter.wiki.git
```

inferrinizzard (approved):

PR #332, opened 2022-07-18
Double check String types for formatter languages

### LINKED ISSUE #331: Double quotes (") in SQL fails formatting for Spark dialect
Using this module in SQL formatter in VSCode plug-in. However if a double-quote is found, the tokenizer errors Tokenizer.tokenize call with the following error
`Error: Parse error: Unexpected ""`

Formatter works when double quotes are replaced with single quotes.

inferrinizzard:
Resolves #331

### DIFF
--- a/src/languages/db2/db2.formatter.ts
+++ b/src/languages/db2/db2.formatter.ts
@@ -177,7 +177,7 @@ export default class Db2Formatter extends Formatter {
       reservedDependentClauses: ['WHEN', 'ELSE', 'ELSEIF'],
       reservedKeywords: keywords,
       reservedFunctionNames: functions,
-      stringTypes: [{ quote: "''", prefixes: ['X', 'G', 'N', 'GX', 'UX', 'U&'] }],
+      stringTypes: [{ quote: "''", prefixes: ['G', 'N', 'X', 'BX', 'GX', 'UX', 'U&'] }],
       identTypes: [`""`],
       positionalParams: true,
       namedParamTypes: [':'],
--- a/src/languages/mariadb/mariadb.formatter.ts
+++ b/src/languages/mariadb/mariadb.formatter.ts
@@ -269,7 +269,10 @@ export default class MariaDbFormatter extends Formatter {
       reservedKeywords: keywords,
       reservedFunctionNames: functions,
       // TODO: support _ char set prefixes such as _utf8, _latin1, _binary, _utf8mb4, etc.
-      stringTypes: ['""', { quote: "''", prefixes: ['B', 'X'] }],
+      stringTypes: [
+        { quote: "''", prefixes: ['B', 'X'] },
+        { quote: '""', prefixes: ['B', 'X'] },
+      ],
       identTypes: ['``'],
       identChars: { first: '$', rest: '$' },
       variableTypes: [
--- a/src/languages/mariadb/mariadb.formatter.ts
+++ b/src/languages/mariadb/mariadb.formatter.ts
@@ -268,7 +268,8 @@ export default class MariaDbFormatter extends Formatter {
       reservedLogicalOperators: ['AND', 'OR', 'XOR'],
       reservedKeywords: keywords,
       reservedFunctionNames: functions,
-      stringTypes: ['""', { quote: "''", prefixes: ['X'] }],
+      // TODO: support _ char set prefixes such as _utf8, _latin1, _binary, _utf8mb4, etc.
+      stringTypes: ['""', { quote: "''", prefixes: ['B', 'X'] }],
       identTypes: ['``'],
       identChars: { first: '$', rest: '$' },
       variableTypes: [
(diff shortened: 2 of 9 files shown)


src/languages/spark/spark.formatter.ts
```diff
@@ -144,7 +144,10 @@ export default class SparkFormatter extends Formatter {
       reservedFunctionNames: functions,
       openParens: ['(', '['],
       closeParens: [')', ']'],
-      stringTypes: [{ quote: "''", prefixes: ['X'] }],
+      stringTypes: [
+        { quote: "''", prefixes: ['r'] },
+        { quote: '""', prefixes: ['r'] },
+      ],
```
nene, line 150: I see you removed the hex-strings `X. These are still supported by Spark.

I updated the docs here as well: https://github.com/sql-formatter-org/sql-formatter/wiki/Strings

inferrinizzard: missed those in the binary section

src/languages/mariadb/mariadb.formatter.ts
```diff
@@ -268,7 +268,8 @@ export default class MariaDbFormatter extends Formatter {
       reservedLogicalOperators: ['AND', 'OR', 'XOR'],
       reservedKeywords: keywords,
       reservedFunctionNames: functions,
-      stringTypes: ['""', { quote: "''", prefixes: ['X'] }],
+      // TODO: support _ char set prefixes such as _utf8, _latin1, _binary, _utf8mb4, etc.
+      stringTypes: ['""', { quote: "''", prefixes: ['B', 'X'] }],
```
nene, line 272: Are we sure that MariaDB does not support `B""` and `X""` strings?

Should also document MariaDB `B`-strings in wiki.

src/languages/spark/spark.formatter.ts
```diff
@@ -144,7 +144,10 @@ export default class SparkFormatter extends Formatter {
       reservedFunctionNames: functions,
       openParens: ['(', '['],
       closeParens: [')', ']'],
-      stringTypes: [{ quote: "''", prefixes: ['X'] }],
+      stringTypes: [
+        { quote: "''", prefixes: ['r', 'X'] },
+        { quote: '""', prefixes: ['r', 'X'] },
```
nene, line 149: Should use uppercase `R` prefix instead of `r` for consistency with all other prefixes configs.

src/languages/sql/sql.formatter.ts
```diff
@@ -70,7 +70,7 @@ export default class SqlFormatter extends Formatter {
       reservedDependentClauses: ['WHEN', 'ELSE'],
       reservedKeywords: keywords,
       reservedFunctionNames: functions,
-      stringTypes: [{ quote: "''", prefixes: ['X'] }],
+      stringTypes: [{ quote: "''", prefixes: ['X', 'N'] }],
```
nene, line 73: Looks like the standard also supports `U&''` strings: https://jakewheat.github.io/sql-overview/sql-2008-foundation-grammar.html#Unicode-character-string-literal

Should also document these all in the wiki.

src/languages/sqlite/sqlite.formatter.ts
```diff
@@ -73,7 +73,10 @@ export default class SqliteFormatter extends Formatter {
       reservedDependentClauses: ['WHEN', 'ELSE'],
       reservedKeywords: keywords,
       reservedFunctionNames: functions,
-      stringTypes: [{ quote: "''", prefixes: ['X'] }],
+      stringTypes: [
+        { quote: "''", prefixes: ['X'] },
+        { quote: '""', prefixes: ['X'] },
```
nene, line 78: Should not add `""` strings to SQLite - that will conflict with `""`-quoted identifiers.

Looks like SQLite has some strange rules where it sometimes treats double-quoted text as an identifier and sometimes as a string. And vice versa with single quotes. See: https://www.sqlite.org/lang_keywords.html

nene:
Additionally, would be nice to have all these additional string types be covered by tests.

inferrinizzard:
will update wiki _after_ code changes are merged, tests to come

src/languages/db2/db2.formatter.ts
```diff
@@ -177,7 +177,7 @@ export default class Db2Formatter extends Formatter {
       reservedDependentClauses: ['WHEN', 'ELSE', 'ELSEIF'],
       reservedKeywords: keywords,
       reservedFunctionNames: functions,
-      stringTypes: [{ quote: "''", prefixes: ['X', 'G', 'N', 'GX', 'UX', 'U&'] }],
+      stringTypes: [{ quote: "''", prefixes: ['G', 'N', 'X', 'BX', 'GX', 'UX', 'U&'] }],
```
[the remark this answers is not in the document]
inferrinizzard: https://www.ibm.com/docs/en/db2-for-zos/11?topic=constants-binary-string

src/languages/mariadb/mariadb.formatter.ts
inferrinizzard: all the examples use single quotes but double quote strings are mentioned in 1 line in the MariaDB docs
so presumably they are compatible, will update

PR #333, opened 2022-07-19
Improve Trino row pattern formatting

verhovsky:
and also remove `` ` `` quotes, because those were only in the grammar definition so that the parser can tell users in an error message that Trino doesn't support them.


verhovsky:
Similarly to me incorrectly including `` ` `` quotes, https://github.com/sql-formatter-org/sql-formatter/wiki/identifiers should actually say "(no differences)" for Trino. Identifiers starting with a number are only [in the grammar](https://github.com/trinodb/trino/blob/0beb0e0ab93315976b4d5ef49fe555a3139cc12a/core/trino-parser/src/main/antlr4/io/trino/sql/parser/SqlBase.g4#L1179-L1181) so that the parser can raise a better error message.

src/languages/trino/trino.formatter.ts
```diff
@@ -136,10 +137,14 @@ export default class TrinoFormatter extends Formatter {
       reservedDependentClauses: ['WHEN', 'ELSE'],
       reservedKeywords: keywords,
       reservedFunctionNames: functions,
-      openParens: ['(', '['],
-      closeParens: [')', ']'],
+      // https://trino.io/docs/current/sql/match-recognize.html#row-pattern-syntax
+      openParens: ['(', '[', '{', '{-'],
+      closeParens: [')', ']', '}', '-}'],
```
nene, line 142: Wow... Looks like Trino has some pretty complex pattern matching syntax.

I would very much appreciate some tests for these patterns.

nene:
Updated the Wiki page for Trino identifiers: https://github.com/sql-formatter-org/sql-formatter/wiki/Identifiers

test/trino.test.ts
```diff
@@ -51,6 +51,56 @@ describe('TrinoFormatter', () => {
...
+            orderdate
+          MEASURES
+            A.totalprice AS starting_price,
+            LAST (B.totalprice) AS bottom_price,
+            LAST (U.totalprice) AS top_price
+          ONE ROW PER MATCH
+          AFTER MATCH
+            SKIP PAST LAST ROW
+          PATTERN
+            ((A | B) {5} {- C + D + -} E +)
```
verhovsky, line 92: should be 

```
            ((A | B){5} {- C+ D+ -} E+)
```

test/trino.test.ts
```diff
@@ -51,6 +51,56 @@ describe('TrinoFormatter', () => {
...
+      MATCH_RECOGNIZE
+        (
+          PARTITION BY
+            custkey
+          ORDER BY
+            orderdate
+          MEASURES
+            A.totalprice AS starting_price,
+            LAST (B.totalprice) AS bottom_price,
+            LAST (U.totalprice) AS top_price
```
verhovsky, line 87: Should actually be

```
            LAST(B.totalprice) AS bottom_price,
```

Seems like these are documented at https://trino.io/docs/current/sql/match-recognize.html#row-pattern-recognition-expressions

Will adding these as both functions and keywords mess up formatting of the keywords? I don't people are really actually using this row pattern stuff, so we should prefer `LAST` as a keyword than a function if that's the case.

test/trino.test.ts
```diff
@@ -51,6 +51,56 @@ describe('TrinoFormatter', () => {
...
+          AFTER MATCH
+            SKIP PAST LAST ROW
+          PATTERN
+            ((A | B) {5} {- C + D + -} E +)
+          SUBSET
+            U = (C, D)
+          DEFINE
+            B AS totalprice < PREV (totalprice),
+            C AS totalprice > PREV (totalprice)
+            AND totalprice <= A.totalprice,
```
verhovsky, line 98: should actually be on one line

```
            C AS totalprice > PREV (totalprice) AND totalprice <= A.totalprice,

```

inferrinizzard: this is a config `logicalOperatorNewline`

verhovsky: Ideally this would be on the same line because this is an `AS` statement, but in a `WHERE` clause the operator should be on a new line.

nene: Yeah, we'll need to revise this `AND` formatting. Currently it gets applied aggressively everywhere, but it really only makes sense inside WHERE and HAVING.

test/trino.test.ts
nene: Well, the main question is, how else is the `LAST` keyword used in Trino? I found three usages:

- `SORT BY col NULLS LAST`
- `AFTER MATCH SKIP TO LAST ROW`
- `AFTER MATCH SKIP TO LAST identifier`

Importantly there doesn't seem to be any cases where we'd want `LAST` followed by `(` to be formatted as `LAST (`. Therefore it should be safe to list it under functions.

nene (approved):

nene (approved):

nene:
@verhovsky One general suggestion regarding pull requests: avoid modifying existing commits with `--force` pushes.

It's hard to tell what has changed as a result. I pretty much have to look through the code all over instead of just looking what has changed since the last time I viewed the pull request.

nene:
This is now released in 9.0.0-beta4

PR #339, opened 2022-07-22
Support number as first char of identifier

### LINKED ISSUE #334: [FORMATTING] Spaces are added when fields begin with numbers
**Input data**
SELECT
  total_pay,
  7d_pay,
  offline_days,
  15d_pay,
  30d_pay,
FROM
  role
**Expected Output**
SELECT
  total_pay,
  7 d_pay,
  offline_days,
  15 d_pay,
  30 d_pay,
FROM
  role
```sql
-- ADD SQL HERE
```

**Actual Output**

```sql
-- ADD SQL HERE
```

**Usage**

F8189081-09C4-44d9-93A0-D324F1FEC4BD

inferrinizzard:
Resolves #334 
Adds support for having a number as first char of identifiers, which is allowed for some languages (mariadb, etc)


nene (changes requested):
Let's not allow identifiers to begin with numbers in all dialects. There are only a few which support this. According to [our docs](https://github.com/sql-formatter-org/sql-formatter/wiki/identifiers) only MariaDB and MySQL support this. For these we already support numbers as first letters, as configured in code:

```ts
      variableTypes: [
        { regex: '@[A-Za-z0-9_.$]+' },
```

test/behavesLikeMariaDbFormatter.ts
```diff
@@ -47,6 +47,37 @@ export default function behavesLikeMariaDbFormatter(format: FormatFn) {
     `);
   });
 
+  // regression test for sql-formatter#334
+  it.only('supports identifiers that start with numbers', () => {
```
nene, line 51: Another `only` in here.

test/behavesLikeMariaDbFormatter.ts
```diff
@@ -47,6 +47,37 @@ export default function behavesLikeMariaDbFormatter(format: FormatFn) {
...
+    );
+    expect(format('SELECT one + 2three, 4 + 5six FROM tbl')).toBe(
+      dedent`
+        SELECT
+          one + 2three,
+          4 + 5six
+        FROM
+          tbl
+      `
+    );
```
nene, line 78: Do we really need three assertions to test this?

nene (changes requested):
The MySQL and MariaDB identifier regexes now allow for identifiers to consist of only numbers, which is not correct.

This situation was previously avoided by matching numbers before identifiers, but with the order reversed, we're now detecting `1`, `2`, `3` as IDENTIFIER tokens. This currently doesn't cause any issues in our formatting as IDENTIFIER and NUMBER are formatted exactly the same way, but it will likely bite us in the future.

I think the solution here is to include a word-boundary check at the end of the NUMBER regex.

src/lexer/Tokenizer.ts
```diff
@@ -73,7 +73,7 @@ export default class Tokenizer {
       [TokenType.QUOTED_IDENTIFIER]: { regex: regex.string(cfg.identTypes) },
       [TokenType.NUMBER]: {
         regex:
-          /(?:0x[0-9a-fA-F]+|0b[01]+|(?:-\s*)?[0-9]+(?:\.[0-9]*)?(?:[eE][-+]?[0-9]+(?:\.[0-9]+)?)?)/uy,
+          /(?:0x[0-9a-fA-F]+|0b[01]+|(?:-\s*)?[0-9]+(?:\.[0-9]*)?(?:[eE][-+]?[0-9]+(?:\.[0-9]+)?)?)(?!\w)/uy,
```
nene, line 76: Why the negative lookahead instead of just `\b`?

src/lexer/regexFactory.ts
```diff
@@ -165,7 +170,9 @@ export const identifierPattern = ({ first, rest, dashes }: IdentChars = {}): str
   const firstChars = escapeRegExp(first ?? '');
   const restChars = escapeRegExp(rest ?? '');
 
-  const pattern = `[${letter}${firstChars}][${letter}${number}${restChars}]*`;
+  const pattern = allowFirstCharNumber
+    ? `[${firstChars}]?[${letter}${number}${restChars}]*[${letter}${restChars}]+[${letter}${number}${restChars}]*`
```
nene, line 174: I don't understand this regex. What is it trying to achieve?

nene (changes requested):

src/lexer/Tokenizer.ts
inferrinizzard: \b would match `[1]two3` whereas the negative lookahead prevents it from being matched at all so the identifier pattern can match later in the priority

src/lexer/regexFactory.ts
inferrinizzard: matches any amount of letter|number|other char with at least 1 non-number char

src/lexer/Tokenizer.ts
nene: Sorry, I don't understand. What do you mean by `[1]two3`?

The `\b` (word boundary) does not include any extra text inside the match, just like negative lookahead.

src/lexer/regexFactory.ts
nene: So, according to this regex, when first character can be `_` we allow `_abc` as identifier, but `_123` is invalid.

I think it would be better to not tackle this problem in here at all. Let's just rely on the order of matching:

- we first try to match a number
- when it succeeds, we won't be attempting to match this same text as identifier.
- when it fails, we know we're not dealing with the invalid case of digits-only identifier.

That's IMHO how most lexers deal with similar issues.

src/lexer/Tokenizer.ts
inferrinizzard: using \b would mean "match if followed by a non-word character"
whereas the negative lookahead is "match unless followed by a word character"

inferrinizzard: the `1` in `1two3` would get tokenized as a number before reaching the identifier pattern, leaving it as `1 (number) two3 (identifier)`

nene: These two statements are logically equivalent.

Although `\b` does not refer to non-word character (that would be `\W`). It refers to word-boundary e.g. being preceded by word character and followed by non-word character, or the other way around.

inferrinizzard: the first statement is A if not B whereas the second is not(A if B)
simply: the \b _does_ match the leading digit before reaching the identifier regex (undesired behaviour) and the negative lookahead does not match so the identifier pattern can match (desired behaviour)

nene: Well, there is no word-boundary between `1` and `two3`. Both numbers and letters are considered to be word characters.

There is though one case where the simple `\b` in the end doesn't work: a number ending with a period like `100.` followed by space. In that case there wouldn't be a word-boundary at the end of the number. So, to cover this ends-with-period case, the negative lookahead is probably a better choice here.

src/lexer/regexFactory.ts
```diff
@@ -165,7 +170,9 @@ export const identifierPattern = ({ first, rest, dashes }: IdentChars = {}): str
   const firstChars = escapeRegExp(first ?? '');
   const restChars = escapeRegExp(rest ?? '');
 
-  const pattern = `[${letter}${firstChars}][${letter}${number}${restChars}]*`;
+  const pattern = allowFirstCharNumber
+    ? `[${firstChars}][${letter}${number}${restChars}]+|[${letter}${number}${restChars}]*[${letter}${restChars}]+[${letter}${number}${restChars}]*`
```
nene, line 174: So, I see you opted for even more complex solution. When I replace this line with:

```suggestion
    ? `[${letter}${number}${firstChars}][${letter}${number}${restChars}]*`
```

the tests still pass just fine.

inferrinizzard: I was avoiding the all number case but it seems that shouldn't happen with the new number regex

nene (approved):

PR #342, opened 2022-07-23
Add support for PLSQL Q strings

inferrinizzard:
added support for the 2 cases of PLSQL Q strings:
- special delimiters: <, {, (, [ (must match with the respective >, }, ), ])
- common delimiters: any char that is not the above

surprisingly, the syntax allows for >, }, ), ] to be used as common delimiters
so both `Q'[foobar]'` and `q']foobar]'` are valid

see: https://docs.oracle.com/cd/B19306_01/server.102/b14200/sql_elements003.htm#i42617
for docs

### DIFF
--- a/src/languages/plsql/plsql.formatter.ts
+++ b/src/languages/plsql/plsql.formatter.ts
@@ -94,8 +94,7 @@ export default class PlSqlFormatter extends Formatter {
       reservedLogicalOperators: ['AND', 'OR', 'XOR'],
       reservedKeywords: keywords,
       reservedFunctionNames: functions,
-      // TODO: support custom-delimited strings: Q'{..}' q'<..>' etc
-      stringTypes: [{ quote: "''", prefixes: ['N'] }],
+      stringTypes: [{ quote: "''", prefixes: ['N'] }, "q''"],
       identTypes: [`""`],
       identChars: { rest: '$#' },
       variableTypes: [{ regex: '&{1,2}[A-Za-z][A-Za-z0-9_$#]*' }],
--- a/src/languages/plsql/plsql.formatter.ts
+++ b/src/languages/plsql/plsql.formatter.ts
@@ -94,7 +94,10 @@ export default class PlSqlFormatter extends Formatter {
       reservedLogicalOperators: ['AND', 'OR', 'XOR'],
       reservedKeywords: keywords,
       reservedFunctionNames: functions,
-      stringTypes: [{ quote: "''", prefixes: ['N'] }, "q''"],
+      stringTypes: [
+        { quote: "''", prefixes: ['N'] },
+        { quote: "q''", prefixes: ['N'] },
+      ],
       identTypes: [`""`],
       identChars: { rest: '$#' },
       variableTypes: [{ regex: '&{1,2}[A-Za-z][A-Za-z0-9_$#]*' }],
--- a/src/lexer/regexFactory.ts
+++ b/src/lexer/regexFactory.ts
@@ -76,6 +76,33 @@ export const parameter = (paramTypes: string[], pattern: string): RegExp | undef
   return patternToRegex(`(?:${typesRegex})(?:${pattern})`);
 };
 
+const buildQStringPatterns = () => {
+  const specialDelimiterMap = {
+    '<': '>',
+    '[': ']',
+    '(': ')',
+    '{': '}',
+  };
+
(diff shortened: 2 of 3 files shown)


src/lexer/regexFactory.ts
```diff
@@ -76,6 +76,33 @@ export const parameter = (paramTypes: string[], pattern: string): RegExp | undef
...
+
+  // replace {left} and {right} with delimiters, collect as array
+  const patternList = Object.entries(specialDelimiterMap).map(([left, right]) =>
+    singlePattern.replace(/{left}/g, escapeRegExp(left)).replace(/{right}/g, escapeRegExp(right))
+  );
+
+  // standard pattern for common delimiters, ignores keys of specialDelimiterMap
+  const standardDelimiterPattern = String.raw`(?<tag>[^\s${escapeRegExp(
+    Object.keys(specialDelimiterMap).join('')
+  )}])(?:(?!\k<tag>).|\k<tag>(?!'))+\k<tag>`;
```
nene, line 98: I think it would be simpler to just hard-code the list of `<{({` inside regex:

```suggestion
  const standardDelimiterPattern = String.raw`(?<tag>[^\s<[({])(?:(?!\k<tag>).|\k<tag>(?!'))+\k<tag>`;
```

Alternatively I'd suggest moving the construction of this character list out of the regex string, which is complex enough without multiple function calls in the middle.

nene, line 98: I'm not sure whether this regex is correct. There are lots of negative lookaheads :)

Specifically It's this middle part that concerns me:

```
(?:(?!\k<tag>).|\k<tag>(?!'))+
```

Wouldn't we get the same result with:

```
(?:(?!\k<tag>').)+
```

Didn't run the code. Pretty tired today as well :)

test/plsql.test.ts
```diff
@@ -91,6 +91,15 @@ describe('PlSqlFormatter', () => {
     `);
   });
 
+  it.only('supports Q custom delimiter strings', () => {
+    expect(format("q'<test string < > 'foo' bar >'")).toBe("q'<test string < > 'foo' bar >'");
+    expect(format("Q'[test string [ ] 'foo' bar ]'")).toBe("Q'[test string [ ] 'foo' bar ]'");
+    expect(format("q'(test string ( ) 'foo' bar )'")).toBe("q'(test string ( ) 'foo' bar )'");
+    expect(format("Q'{test string { } 'foo' bar }'")).toBe("Q'{test string { } 'foo' bar }'");
+    expect(format("q'%test string % % 'foo' bar %'")).toBe("q'%test string % % 'foo' bar %'");
+    expect(format("Q'Xtest string X X 'foo' bar X'")).toBe("Q'Xtest string X X 'foo' bar X'");
```
nene, line 100: I suggest adding tests for code like:

```
q'X hello X'X''
```

To check that it gets tokenized as `q'X hello X'`, `X`, `''`.

Making sure that we don't greedily consume until last `X'`, but stop at the very first one.

src/lexer/regexFactory.ts
nene: I also notice that we are using a greedy quantifier `+` here. The regex might end up simpler if we instead used a non-greedy `+?`.

test/plsql.test.ts
```diff
@@ -91,6 +91,15 @@ describe('PlSqlFormatter', () => {
     `);
   });
 
+  it.only('supports Q custom delimiter strings', () => {
```
nene, line 94: Should not leave `only` in here :)

src/languages/plsql/plsql.formatter.ts
```diff
@@ -94,8 +94,10 @@ export default class PlSqlFormatter extends Formatter {
       reservedLogicalOperators: ['AND', 'OR', 'XOR'],
       reservedKeywords: keywords,
       reservedFunctionNames: functions,
-      // TODO: support custom-delimited strings: Q'{..}' q'<..>' etc
-      stringTypes: [{ quote: "''", prefixes: ['N'] }],
+      stringTypes: [
+        { quote: "''", prefixes: ['N'] },
+        { quote: "q''", prefixes: ['N'] },
```
nene, line 99: Should also test for the `N`-prefixed Q-strings.

nene (changes requested):
Looks good in general.

There's an `it.only(` left inside tests and I have some doubts about regexes.

nene (approved):

PR #343, opened 2022-07-25
Separate sublist for Trino row pattern functions

nene:
Reorganizing the code added in: https://github.com/sql-formatter-org/sql-formatter/pull/333/commits/03d6bffef1fedca3997585dc6ad4f4ee78c149f5

Instead of placing the row functions inside the main functions list, separated them out, so that when one goes to review the functions list in the future, one wouldn't expect to find these functions in the [Trino functions list](https://github.com/trinodb/trino/blob/432d2897bdef99388c1a47188743a061c4ac1f34/docs/src/main/sphinx/functions/list-by-topic.rst).

Also restored `PERMUTE` to keywords list, so it lists all the keywords, regardless of whether SQL Formatter also recognizes them as function names.


nene:
@verhovsky FYI.

nene:
Not much to review here. Merging straight in.

verhovsky:
I did submit a PR to add them to that list, I guess we'll see if they think that makes sense.

https://github.com/trinodb/trino/pull/13327

PR #344, opened 2022-07-25
Add support for PLSQL tbl@dblink syntax

### LINKED ISSUE #338: [FORMATTING] PLSQL does not format dblink syntax properly
**Input data**
Indentation style: Tabular, Right.
```
select * from andy.t_test@dblink
where id between 1 and 5000;
```

**Expected Output**

```sql
 select * 
   from andy.t_test@dblink
  where id between 1 and 5000;
```

**Actual Output**

```sql
   select *
     from andy.t_test @ dblink
    where id between 1       and 5000;
```

**Usage**

https://sql-formatter-org.github.io/sql-formatter/

PLSQL

Newest version, should be 8.x.

inferrinizzard:
- moved language to part of FormatOptions
- added edge case for `@` operator only when on PLSQL language
- Resolves #338


nene (approved):
Looks great!

PR #346, opened 2022-07-26
Review and revise joins and set operations

nene:
Extended the `expandPhrases()` function with `{FOO | BAR}` syntax.

Used it to generate `JOIN` commands and set operations (`UNION`, `EXCEPT`, ...). Revised these lists based on the docs in wiki, which resulted in removal of support for various types of joins from many dialects. Likely these were there as a result of copy-paste creation of new dialects based on existing ones.

Notably moved `CROSS APPLY` and `OUTER APPLY` out of set operations. They work really more like joins.

Updated tests for joins and added tests for set operations, which previously had none.

### DIFF
--- a/src/languages/bigquery/bigquery.formatter.ts
+++ b/src/languages/bigquery/bigquery.formatter.ts
@@ -106,9 +106,9 @@ const reservedCommands = expandPhrases([
 ]);
 
 const reservedBinaryCommands = expandPhrases([
-  'INTERSECT [ALL | DISTINCT]',
-  'UNION [ALL | DISTINCT]',
-  'EXCEPT [ALL | DISTINCT]',
+  'UNION {ALL | DISTINCT}',
+  'EXCEPT DISTINCT',
+  'INTERSECT DISTINCT',
 ]);
 
 const reservedJoins = expandPhrases([
--- a/src/languages/bigquery/bigquery.formatter.ts
+++ b/src/languages/bigquery/bigquery.formatter.ts
@@ -112,9 +112,9 @@ const reservedBinaryCommands = expandPhrases([
 ]);
 
 const reservedJoins = expandPhrases([
-  '[LEFT | RIGHT | FULL] [OUTER] JOIN',
-  'INNER JOIN',
-  'CROSS JOIN',
+  'JOIN',
+  '{LEFT | RIGHT | FULL} [OUTER] JOIN',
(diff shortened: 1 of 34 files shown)


test/options/indentStyle.ts
```diff
@@ -60,7 +60,7 @@ export default function supportsIndentStyle(format: FormatFn) {
           dedent`
             SELECT *
             FROM a
-            UNION DISTINCT
+            UNION ALL
```
nene, line 63: `UNION` and `UNION ALL` are the only ones supported by all dialects.

src/languages/bigquery/bigquery.formatter.ts
```diff
@@ -106,15 +106,15 @@ const reservedCommands = expandPhrases([
 ]);
 
 const reservedBinaryCommands = expandPhrases([
-  'INTERSECT [ALL | DISTINCT]',
-  'UNION [ALL | DISTINCT]',
-  'EXCEPT [ALL | DISTINCT]',
+  'UNION {ALL | DISTINCT}',
+  'EXCEPT DISTINCT',
+  'INTERSECT DISTINCT',
```
nene, line 111: Reordered these all to match the ordering in wiki.

src/languages/spark/spark.formatter.ts
```diff
@@ -80,54 +81,23 @@ const reservedCommands = [
...
-  'UNION DISTINCT',
-  'EXCEPT',
-  'EXCEPT ALL',
-  'EXCEPT DISTINCT',
-  'MINUS',
-  'MINUS ALL',
-  'MINUS DISTINCT',
-  // apply
-  'CROSS APPLY',
-  'OUTER APPLY',
```
nene, line 99: Spark doesn't actually support `APPLY`, just like it also doesn't support `MINUS`. Likely just a result of copy-pasting these here without checking when Spark dialect was created.

inferrinizzard:
This is a good support addition but definitely would have preferred to have seen this before it was merged

inferrinizzard:
also, I think the NATURAL JOIN and NATURAL INNER JOIN lines can be simplified to NATURAL [INNER] JOIN

nene:
Tackled the last comment in here: https://github.com/sql-formatter-org/sql-formatter/commit/93f298c2f0bbb9a00d535329f8013405bcc7181b

PR #348, opened 2022-07-26
Revise keyword lists

nene:
Went through keyword lists of all dialects and synced them with official docs.

Previously in many places I had removed keywords from `reservedKeywords` if they appear in other lists like `reservedCommands`. I think better to always have them present in the main keywords list. Makes it simpler to compare it with official documentation.

Dropped lots of keywords from Transact-SQL and Postgres:

- Transact-SQL list contained future reserved words, which don't concern us. 
- PostgreSQL list contained full list of keywords in Postgres documentation, which however also lists lots of keywords supported by SQL standards, but not by Postgres. Filtered out only those that Postgres actually supports.


src/languages/mariadb/mariadb.keywords.ts
```diff
@@ -621,19 +671,24 @@ export const keywords = flatKeywordList({
...
     'X509',
+    'XOR',
     'XA',
     'XML',
     'YEAR',
     'YEAR_MONTH',
     'ZEROFILL',
   ],
   constraints: ['ON DELETE', 'ON UPDATE'],
+  charset: ['CHARACTER SET'],
```
nene, line 693: Moved `CHARACTER SET` out of the main keywords list. Now the main lists only contain single-word keywords.

Planning to move these multi-word ones out of the `*.keywords.ts` files completely, to treat them similarly to lists like `reservedCommands`.

PR #350, opened 2022-07-27
Fix BETWEEN..AND formatting in tabular style

### LINKED ISSUE #341: [FORMATTING] when set Indentation style to "Tabular, Right", it doesn't format "between...and" well.
SQL:
```
select * from andy.t_test where id between 1 and 5000;
```

When set Indentation style to "Tabular, Right", output from demo website:
```sql
   select *
     from andy.t_test
    where id between 1       and 5000;
```
Many thanks for your great work!

nene:
Fixes #341

### DIFF
--- a/src/formatter/ExpressionFormatter.ts
+++ b/src/formatter/ExpressionFormatter.ts
@@ -125,7 +125,7 @@ export default class ExpressionFormatter {
       WS.SPACE,
       this.show(node.expr1),
       WS.SPACE,
-      this.show(node.andToken),
+      this.showNonTabular(node.andToken),
       WS.SPACE,
       this.show(node.expr2),
       WS.SPACE
@@ -369,7 +369,12 @@ export default class ExpressionFormatter {
     }
   }
 
-  // don't call this directly, always use show() instead.
+  // Like show(), but skips tabular formatting
+  private showNonTabular(token: Token): string {
+    return this.showToken(token);
+  }
+
+  // don't call this directly, always use show() or showNonTabular() instead.
   private showToken(token: Token): string {
     if (isReserved(token)) {
       switch (this.cfg.keywordCase) {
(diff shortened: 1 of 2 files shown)


inferrinizzard:
the solution here would have been to default to setting `logicalOperatorNewline` to true for all tabular modes since the purpose of the tabular mode is to align _all_ keywords along the central column

nene:
Emm... do you suggest formatting `WHERE x BETWEEN 10 AND 20` as:

```sql
SELECT *
  FROM tbl
 WHERE x BETWEEN 10
   AND 20
```

That doesn't really make sense to me. It's also not how the person reporting the bug expected the result to look.

nene:
> the purpose of the tabular mode is to align *all* keywords along the central column

I don't think that's the case. You probably meant something else. Aligning all keywords would give us a pretty silly result like:

```sql
 SELECT *
   FROM tbl
   LEFT
   JOIN bar
     ON id = t_id
  WHERE x
BETWEEN 10
    AND 20
```

inferrinizzard:
the example you gave is not unreasonable, the only change I would make there is 
```sql
    SELECT *
      FROM tbl
 LEFT JOIN bar
        ON id = t_id
     WHERE x
   BETWEEN 10
       AND 20
```

since this structure prefers verticality over horizontal growth

but I could also see the case for having WHERE x BETWEEN y AND z on the same line, which could be configurable via (`keywordNewline`)

nene:
Well, do you actually know of somebody formatting the `WHERE x BETWEEN y AND z` expression like that?

My understanding of the `tabularLeft` and `tabularRight` styles has been that this only really changes how the text after `RESERVED_COMMAND` keywords gets indented. (Though for now it also changes the indentation level `AND` and joins.) But all other keywords are not effected.

To take the above example further, I'd expect you would not want the following:

```sql
    SELECT tbl.n
        AS name, tbl.f
        AS foo 
      FROM tbl;

ALTER TABLE foo
     MODIFY col1 char(100)
   NOT NULL;
```

I mean, the nice alignment of keywords in a single column surely isn't a goal in itself?

PR #351, opened 2022-07-27
Rename BINARY_COMMAND to SET_OPERATION

nene:
Previously this contained both set operations and joins.

The new name better reflects that it only contains `UNION`, `EXCEPT`, `INTERSECT` and other set operations.

### DIFF
--- a/src/languages/bigquery/bigquery.formatter.ts
+++ b/src/languages/bigquery/bigquery.formatter.ts
@@ -105,7 +105,7 @@ const reservedCommands = expandPhrases([
   'EXPORT DATA',
 ]);
 
-const reservedBinaryCommands = expandPhrases([
+const reservedSetOperations = expandPhrases([
   'UNION {ALL | DISTINCT}',
   'EXCEPT DISTINCT',
   'INTERSECT DISTINCT',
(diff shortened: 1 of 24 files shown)


src/lexer/token.ts
```diff
@@ -68,18 +68,14 @@ export const isToken = {
   WITH: testToken({ value: 'WITH', type: TokenType.RESERVED_COMMAND }),
 };
 
-/** Checks if token is a Reserved Command or Reserved Binary Command */
-export const isCommand = (token: Token): boolean =>
-  token.type === TokenType.RESERVED_COMMAND || token.type === TokenType.RESERVED_BINARY_COMMAND;
```
nene, line 73: Inlined this to AliasAs - the single place it's used. Also switched it to only check for `RESERVED_COMMAND`, which seems to be the only token type for which we really need to check in there.

src/languages/bigquery/bigquery.formatter.ts
```diff
@@ -105,7 +105,7 @@ const reservedCommands = expandPhrases([
   'EXPORT DATA',
 ]);
 
-const reservedBinaryCommands = expandPhrases([
+const reservedSetOperations = expandPhrases([
   'UNION {ALL | DISTINCT}',
```
inferrinizzard, line 109: should this line be brackets?

inferrinizzard (approved):

src/languages/bigquery/bigquery.formatter.ts
nene: Yep. BigQuery does not allow plain `UNION`: https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#set_clause

PR #354, opened 2022-07-28
Add escapes to string quotes

### LINKED ISSUE #308: More than one string with trailing backslash
**Describe the bug**

I tried formatting

```sql
select
  *
from
  mytable
where
  mycol_a like 'prefix\_%' escape '\'
  and mycol_b like 'prefix\_%' escape '\'
```

**Expected behavior**

Output exactly the same as input.

**Actual behavior**

```
Error: Parse error: Unexpected "\_%'"
    at E.value (https://unpkg.com/sql-formatter@latest/dist/sql-formatter.min.js:1:34317)
    at E.value (https://unpkg.com/sql-formatter@latest/dist/sql-formatter.min.js:1:41347)
    at I.value (https://unpkg.com/sql-formatter@latest/dist/sql-formatter.min.js:1:30283)
    at Module.SS (https://unpkg.com/sql-formatter@latest/dist/sql-formatter.min.js:1:210009)
    at HTMLTextAreaElement.format (https://sql-formatter-org.github.io/sql-formatter/index.js:50:31)
```

If I add a space after the first `'\'` to `'\ '` it works. If I remove the last `and [etc.]` line then it also works.

**Usage**

- How are you calling / using the library? https://sql-formatter-org.github.io/sql-formatter/
- What SQL language(s) does this apply to? I tried every formatter. This is definitely [valid syntax in Trino](https://github.com/trinodb/trino/blob/91ce0ec5c98a5019f5fb66683bae0ee6c18201b0/core/trino-parser/src/main/antlr4/io/trino/sql/parser/SqlBase.g4#L1147) but also other SQL dialects I think
- Which SQL Formatter version are you using? `latest`. v4.0.0 (with some modifications which shouldn't affect this case) adds a bunch of spaces to the last line `and mycol_b like ' prefix \ _ % ' escape ' \ '`

inferrinizzard:
- adds support for specific escapes for string types
- adds tests for all quote types + prefixes + escapes
- updates some string regex
- adds string escapes for all languages
- moves string types to class static members
- removes old language specific string tests
- Resolves #308


src/lexer/regexFactory.ts
```diff
@@ -112,23 +112,33 @@ const buildQStringPatterns = () => {
...
-  '""': '(?:"[^"\\\\]*(?:\\\\.[^"\\\\]*)*(?:"|$))+',
-  "''": "(?:'[^'\\\\]*(?:\\\\.[^'\\\\]*)*(?:'|$))+",
-  '$$': '(?<tag>\\$\\w*\\$)[\\s\\S]*?(?:\\k<tag>|$)',
-  "'''..'''": "'''[^\\\\]*?(?:\\\\.[^\\\\]*?)*?(?:'''|$)",
-  '""".."""': '"""[^\\\\]*?(?:\\\\.[^\\\\]*?)*?(?:"""|$)',
-  '{}': '(?:\\{[^\\}]*(?:$|\\}))',
+  '``': '`(?:[^`]|``)*(?:`|$)',
+  '[]': String.raw`\[(?:[^\]]|\]\])*(?:\]|$)`,
+  '""': String.raw`(?:"{escape}*?(?:\\.[^"\\]*)*(?:"|$))+`,
+  "''": String.raw`(?:'{escape}*?(?:\\.[^'\\]*)*(?:'|$))+`,
```
nene, line 118: I'm pretty confident this doesn't work as intended.

Let me simplify this regex a bit with additional whitespace and removal of  `:?` optional capture groups:

```
( '{escape}*? (\\.[^'\\]*)* ('|$) )+
```

a) When escapes array contains a single quote `'`, the resulting regex will be:

```
( '[^']*? (\\.[^'\\]*)* ('|$) )+
```

This regex will happily match the following string:

```
'hello\'world'
```

b) When escapes array contains a backslash, the resulting regex will be:

```
( '[^\\]*? (\\.[^'\\]*)* ('|$) )+
```

This regex happily matches the following string:

```
'bla'bla''bla'
```

test/trino.test.ts
```diff
@@ -25,8 +25,8 @@ describe('TrinoFormatter', () => {
   supportsComments(format);
   supportsCreateTable(format);
   supportsDeleteFrom(format);
-  supportsStrings(format, ["''", "X''", "U&''"]);
-  supportsIdentifiers(format, ['""']);
+  supportsStrings(format, TrinoFormatter.stringTypes);
+  supportsIdentifiers(format, ['""', '``']);
```
nene, line 29: Trino doesn't support `` `..` `` quoted identifiers.

See the comment in wiki: https://github.com/sql-formatter-org/sql-formatter/wiki/identifiers

test/plsql.test.ts
```diff
@@ -116,11 +116,30 @@ describe('PlSqlFormatter', () => {
...
+  it('formats SELECT query with OUTER APPLY', () => {
+    const result = format('SELECT a, b FROM t OUTER APPLY fn(t.id)');
+    expect(result).toBe(dedent`
+      SELECT
+        a,
+        b
+      FROM
+        t
+      OUTER APPLY
+      fn (t.id)
```
nene, line 142: I think something has gone wrong with the merge in here.

I moved these two tests inside `supportsJoin()` helper.

nene (changes requested):
Thanks for the effort on this.

The escaping implementation looks buggy though. Also the tests are way too complex to understand.

inferrinizzard:
Placing these changes on pause until further notice

PR #355, opened 2022-07-28
Update regexTypes import

### LINKED ISSUE #353: Missing regexTypes.d.ts in distributed module
There seems to be a problem with the distribution of the [`regexTypes.d.ts`](https://github.com/sql-formatter-org/sql-formatter/blob/master/src/lexer/regexTypes.d.ts) file in the module available on npm.

When typescripts checks a project using the library, it complains:

```
node_modules/sql-formatter/lib/src/lexer/Tokenizer.d.ts:2:29 - error TS2307: Cannot find module './regexTypes' or its corresponding type declarations.

2 import * as regexTypes from "./regexTypes";
                              ~~~~~~~~~~~~~~


Found 1 error in node_modules/sql-formatter/lib/src/lexer/Tokenizer.d.ts:2
```

`lib/src/lexer` indeed does not contain the requisite file. It appears that the compilation step considers it as a plain typescript file and emits an (empty) module in `lib/lexer` instead:

```console
$ find node_modules/sql-formatter -name 'regexTypes.*'
node_modules/sql-formatter/lib/lexer/regexTypes.d.js
node_modules/sql-formatter/lib/lexer/regexTypes.d.js.map
```

The issue appears in (at least) 8.2 and 8.1.

inferrinizzard:
- Resolves #353

### DIFF
--- a/src/lexer/Tokenizer.ts
+++ b/src/lexer/Tokenizer.ts
@@ -1,6 +1,6 @@
 import { Token, TokenType } from 'src/lexer/token';
 import * as regex from 'src/lexer/regexFactory';
-import * as regexTypes from 'src/lexer/regexTypes';
+import type * as regexTypes from 'src/lexer/regexTypes';
 
 import { escapeRegExp } from './regexUtil';
 import TokenizerEngine, { type TokenRule } from './TokenizerEngine';


nene:
It seems to me that the main source of this whole problem is that we're using `.d.ts` files in our code. What are the benefits of using `.d.ts` files versus just always using plain `.ts` files?

I googled around to understand this issue here in particular, but didn't really manage to find any good information so far. Found one [similar-looking issue](https://github.com/TypeStrong/ts-loader/issues/1036#issuecomment-630179801), where the author remarks:

> Reading @appzuka's explanation it does make sense for .d.ts/declaration files to serve the purpose of declaring types to be used for type checking during build and then never emitted. This makes it possible for each package to provide its own types without them getting duplicated by every package that depends on it.

It seems to me that we're better off not using these declaration files. That way we don't run into issues like this in the future and there will be less file extensions in the code base.

IMHO it's also not a good practice in general to use different file extensions based on the content of a file. Like today `foo.d.ts` only contains types, but tomorrow we decide to add an actual function inside it, so we'd have to rename it to `foo.ts`. Then a few days later we drop the function and we'd need to rename the file again.

I once experienced a similar issue in React codebase where we used `.tsx` extensions for files that contained JSX syntax and normal `.ts` extension on files that did not. In that case at least there was a solid reason to use `.tsx` extension, as TypeScript only supports JSX syntax with these files. But it was a nuisance to change the extension when one decided to include some JSX syntax to a file. A better approach was to just switch all files to `.tsx` extension and avoid messing around with different extensions.

inferrinizzard:
The distinction is made not on content but on purpose of the file and declaration files are _only_ for types, there shouldn't be utility functions or non-type members included so no file renaming should ever take place - any non-type items should go into their own file

nene:
Well... a "purpose" of a file in codebase is rarely completely fixed.

For example `ast.ts` contains types for the AST nodes, plus a little helper function, which actually isn't used any more. So if we remove this unused helper function, should we rename this file to `.d.ts` or not? Is the purpose of this file now to contain only types or not? Though actually we can't rename it to `.d.ts` because enums aren't allowed in `.d.ts` files - so although our purpose might be to only store types, and we only use the enum as a type, the technical quirks of typescript don't allow us to use `.d.ts` extension here.

Anyway, regardless of this whole purpose thing.... what do we gain by using `.d.ts` files? Which problem does it solve?

Issue #356, opened 2022-07-28
[FORMATTING] formatter doesn't format multiword commands correctly with keywordCase = "upper"


m-ahmadkhan:
**Input data**
```sql
CREATE
OR
REPLACE
TABLE
FUNCTION
  `mydataset.mytablefunc` (p1 STRING, p2 STRING) AS (
    SELECT *
    FROM
      mtytable2
    WHERE
      col1 = p1 AND col2 = p2
  );
```

**Actual Output**

Formatting it with the following config:

```
{
  "language": "bigquery",
  "keywordCase": "upper"
}
```

produces the following result:
```sql
CREATE
OR
REPLACE
TABLE
FUNCTION
  `mydataset.mytablefunc` (p1 STRING, p2 STRING) AS (
    SELECT
      *
    FROM
      mtytable2
    WHERE
      col1 = p1
      AND col2 = p2
  );
```

Bug: It didn't format `CREATE OR REPLACE TABLE FUNCTION` to a single line.

And if I format the same query with keywordCase: "preserve", it formats the query as expected.

```
{
  "language": "bigquery",
  "keywordCase": "preserve"
}
```

Output:

```sql
CREATE OR REPLACE TABLE FUNCTION
  `mydataset.mytablefunc` (p1 STRING, p2 STRING) AS (
    SELECT
      *
    FROM
      mtytable2
    WHERE
      col1 = p1
      AND col2 = p2
  );
```

**Usage**

- How are you calling / using the library? `yarn sql-formatter -o query.sql -l bigquery -c config.json query.sql`
- What SQL language(s) does this apply to? `bigquery`
- Which SQL Formatter version are you using? `9.0.0-beta4`

nene:
Fix for this is now released in 9.0.0-beta.5

PR #357, opened 2022-07-28
Equalize whitespaces for keywordCase upper/lower

m-ahmadkhan:
Fixes https://github.com/sql-formatter-org/sql-formatter/issues/356


m-ahmadkhan:
@inferrinizzard didn't notice that you had self-assigned the bug. I've created this PR to fix it.

src/formatter/ExpressionFormatter.ts
```diff
@@ -381,9 +381,9 @@ export default class ExpressionFormatter {
         case 'preserve':
           return equalizeWhitespace(token.text);
         case 'upper':
-          return token.value;
+          return equalizeWhitespace(token.value);
         case 'lower':
-          return token.value.toLowerCase();
+          return equalizeWhitespace(token.value.toLowerCase());
```
nene, line 386: The real problem is that `token.value` should already be in upper case and with single spaces (i.e. the `value` field should be in a canonical form).

This allows us to perform simple comparisons like `if (token.value === 'INSERT INTO`)` instead of having to use a regex to compare. So this issue really needs to be handled inside Tokenizer.

test/behavesLikeSqlFormatter.ts
```diff
@@ -160,15 +160,20 @@ export default function behavesLikeSqlFormatter(format: FormatFn) {
...
       INSERT INTO
         Customers (ID, MoneyBalance, Address, City)
       VALUES
         (12, -123.4, 'Skagen 2111', 'Stv');
-    `);
+    `;
+    expect(format(input)).toBe(expected);
+
+    // regression test for #356
+    expect(format(input, { keywordCase: 'upper' })).toBe(expected);
```
nene, line 176: Better to not repurpose existing tests. Should instead create a new test inside `test/options/keywordCase.ts`.

nene (changes requested):
Thanks for the PR, but this issue needs to be fixed at the Tokenizer level, not inside formatting part.

test/options/keywordCase.ts
```diff
@@ -61,4 +61,40 @@ export default function supportsKeywordCase(format: FormatFn) {
...
+        *
+      FROM
+        mytable
+        INNER JOIN mytable2 ON mytable1.col1 = mytable2.col1
+      WHERE
+        mytable2.col1 = 5;
+    `;
+
+    expect(format(input, { keywordCase: 'upper' })).toBe(expected);
+  });
```
nene, line 99: I'd suggest combining these two test to a single one that covers multiple things. For example we could a `SELECT` with `INNER JOIN` and `ORDER BY` - covering both multi-word commands and joins.

Additionally it might be good to use lowercase keywords in input, like most other `keywordCase:upper` tests do.

Just a suggestion. The tests are good in their current form as well.

nene (changes requested):
Thanks. That's much better.

But let's apply this `equalizeWhitespace()` to all `RESERVED_*` token types. Some of these might only contain single words, but it's better to be consistent. e.g. `RESERVED_FUNCTION_NAME` already contains multi-word names.

nene:
Great stuff. Thanks.

nene (approved):

nene:
@m-ahmadkhan  FYI, what I really had in mind about simplifying this test was more like so: https://github.com/sql-formatter-org/sql-formatter/commit/3d22c907aee53d04de06ca9c091c9d2fc7a386d3

Testing for `ORDER BY` and `GROUP BY` is really about the same as testing for `INSERT INTO`. The formatter treats these as the same kind of thing (like in, these are both indented to the same level).

PR #359, opened 2022-07-28
Add # as Identifier first char for Redshift

### LINKED ISSUE #358: [FORMATTING] # prefixed temporary tables are not formatted correctly
**Input data**

Using both the SQL formatter and Redshift formatter, a space is added between a `#` prefix and a table name. Redshift allows table names beginning with `#` to denote a temporary table, without the use of the `TEMP` keyword: https://docs.aws.amazon.com/redshift/latest/dg/r_CREATE_TABLE_NEW.html

The result is a statement which Redshift cannot parse.

**Input**

```sql
CREATE TABLE #tablename AS ...;
```

**Expected Output**

```sql
CREATE TABLE #tablename AS ...;
```

**Actual Output**

```sql
CREATE TABLE # tablename AS ...;
```

**Usage**

- How are you calling / using the library? NodeJS, but I have noticed it in the VSCode extension too
- What SQL language(s) does this apply to? Redshift at minimum.
- Which SQL Formatter version are you using? Redshift

inferrinizzard:
Resolves #358 
\# added to valid identifier chars for Redshift, first char only

### DIFF
--- a/src/languages/redshift/redshift.formatter.ts
+++ b/src/languages/redshift/redshift.formatter.ts
@@ -130,6 +130,7 @@ export default class RedshiftFormatter extends Formatter {
       reservedFunctionNames: functions,
       stringTypes: ["''"],
       identTypes: [`""`],
+      identChars: { first: '#' },
       numberedParamTypes: ['$'],
       operators: RedshiftFormatter.operators,
     });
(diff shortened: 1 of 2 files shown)


nene (approved):
Should also document the change in identifier syntax in wiki.

PR #362, opened 2022-07-30
Query syntax update

nene:
Reviewed supported SELECT statement clauses (based on our wiki docs).

Main changes:

- Moved list of these clauses to top of reservedCommands list.
- Added support for syntax variations like `SELECT [DISTINCT]` and `WITH [RECURSIVE]`.
- Overhauled support for limiting: dropped `LIMIT` clause support from several dialects, added `FETCH FIRST` support to several others.
- Separate tests for `WITH` clause.
- Simplified some old tests - deleted some completely.

Didn't attempt covering all the minor syntax variations with tests. Concentrated on the more substantial differences: like testing `LIMIT`, `FETCH FIRST` and `WITH` support, but not testing `WITH RECURSIVE` and `ORDER SIBLINGS BY` support.


src/languages/bigquery/bigquery.formatter.ts
```diff
@@ -6,19 +6,19 @@ import { keywords } from './bigquery.keywords';
 import { functions } from './bigquery.functions';
 
 const reservedCommands = expandPhrases([
-  // DQL, https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax
-  'SELECT',
+  // Queries: https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax
```
inferrinizzard, line 9: replace with DQL to follow pattern of DDL, DML, DQL, DCL

inferrinizzard (approved):

src/languages/bigquery/bigquery.formatter.ts
nene: I specifically opted to not use DQL. These three-letter-acronymes are easy to mix up, and I always have to pause and think what they mean.

PR #363, opened 2022-08-01
Separate multi-word keywords from the rest

nene:
Added `RESERVED_PHRASE` token type. This serves as a place for multi-word sequences of keywords, to separate them from the single-word list of ordinary keywords.

Cleaned up all keyword lists. So now finally the `*.keyword.ts` files only list single-word keywords.


inferrinizzard:
Doesn't the current regex ordering already prioritise by length?
These phrases should already be matching before their single word counterparts
Seems a little weird to split them only based on length vs semantic meaning

nene:
You're correct, we already prioritize by length. In the end it'll work out exactly the same. You're also right that there's nothing semantic about these sequences of keywords to group them together, other than the fact that they're multi-word sequences.

I had two goals for doing this reorganization:

- First, our current keywords lists contain some "keywords" which aren't really single words. For me, this feels far from intuitive. Even the token name is `RESERVED_KEYWORD`, which suggest a single word, but currently it might be more than one word.
- Second, I'd like to evolve our code more towards a parser-based approach. That is, separate keywords should be detected by lexer, but sequences of these keywords should really be handled at the parser level. For example, currently these keyword-sequence-tokens will break down when one adds a comment between them (e.g. `INSERT /* comment */ INTO`). Handling such cases is really not feasible inside lexer, so it makes sense to me to start separating the data needed by lexer (keywords) from data that's really territory of a parser (sequences of keywords).

inferrinizzard (approved):

PR #364, opened 2022-08-01
Review of "verify" comments in reservedCommands lists

nene:
One notable change: dropped support for `CREATE TABLE` and `DROP TABLE` from N1QL.

As a result had to reorganize some tests, as we were assuming these statements are supported by all dialects.

### DIFF
--- a/test/options/aliasAs.ts
+++ b/test/options/aliasAs.ts
@@ -206,23 +206,6 @@ export default function supportsAliasAs(format: FormatFn) {
       );
     });
 
-    it('does not remove AS keyword when "AS WITH" or "AS ("', () => {
-      const result = format(
-        dedent`CREATE TABLE 'test.example_table' AS WITH cte AS (SELECT a AS alpha)`,
-        { aliasAs: 'never' }
-      );
-
-      expect(result).toBe(dedent`
-        CREATE TABLE
-          'test.example_table' AS
-        WITH
-          cte AS (
-            SELECT
-              a alpha
-          )
-      `);
-    });
-
     it('does not remove AS keywords from CAST(... AS ...)', () => {
       const result = format(
         dedent`SELECT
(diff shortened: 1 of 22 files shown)


test/options/aliasAs.ts
```diff
@@ -206,23 +206,6 @@ export default function supportsAliasAs(format: FormatFn) {
...
-      expect(result).toBe(dedent`
-        CREATE TABLE
-          'test.example_table' AS
-        WITH
-          cte AS (
-            SELECT
-              a alpha
-          )
-      `);
-    });
```
nene, line 224: I wanted to fix this test, but it turned out to be very broken:

- `CREATE TABLE 'test.example_table'` is incorrect syntax (don't know of any SQL dialect supporting table names as strings). I removed the quotes, in which case the test started to fail :(
- `AS WITH` seems to be incorrect syntax as well. Didn't find any example of this (expect for one stack-overflow post where the author complained over this syntax not working). I removed the `AS WITH` part to check that we at lest correctly handle `CREATE TABLE foo AS (SELECT ...)`, but again the test broke.

Really the current implementation of aliasAs is pretty broken. The largest problem being with `aliasAs: "never"` where we attempt to remove `AS` keywords from all other places besides the `SELECT` clause. But there are so many of these and in most cases we shouldn't remove the `AS`. But our implementation has it backwards: it first removes the `AS` and later attempts to restore it. It's really hard to wrap your head around that logic.

I'm thinking of a few options:

- completely throwing this aliasAs option away.
- making the aliasAs only work with SELECT clauses. This would at least bring some sanity. The current `aliasAs: "select"` is a pretty weird option - it enforces `AS` in select, but forbids it elsewhere. Like, there's no option to forbid or enforce it inside select without changing it elsewhere.

inferrinizzard (approved):

PR #370, opened 2022-08-03
Remove .d.ts file extensions

nene:
As a follow-up to #355, where a question was raised about the benefits of using TypeScript type declaration files.

As no benefits where brought up, and there are known downsides to using them (like the increased likelyhood of imported file not being included to released version), I'm switching all files back to plain .ts extension.


inferrinizzard (approved):
approved, can we also rename regexTypes.ts to regex.types.ts

nene:
> can we also rename regexTypes.ts to regex.types.ts

Actually, I think regexTypes isn't really a best name for this file, as these types there don't necessarily describe regexes - that's more of an implementation detail. They really just define different kinds of identifiers, quotes, variables. That is, these are really just sub-types used by the TokenizerOptions type. I think a better approach would be to extract TokenizerOptions to separate `TokenOptions.ts` file and place all these types in there.

The other file `types.ts` has even worse name - it really doesn't tell anything about the purpose of these types there. Again, would be better to rename it to `FormatOptions.ts`, which is the main config option type that is exported from `types.ts`.

PR #371, opened 2022-08-03
Drop aliasAs config option

nene:
I started to refactor/rewrite our existing implementation, to only support aliasAs inside SELECT clauses, but came to the realization that there are so many cases where this option is broken, and it's not really sensible to attempt fixing all these issues with the current token-list-manipulation approach.

Really the better way is to perform this transformation after parsing.  But for now our AST doesn't contain enough information for this.

I should also note that this aliasAs implementation has gradually degraded, as it doesn't really have enough tests. For example:

- we splitted our identifiers to `IDENTIFIER` and `QUOTED_IDENTIFIER`, but failed to properly update aliasAs, so now it doesn't work with quoted identifiers.
- with the addition of support for `SELECT DISTINCT`, `SELECT ALL` and other combination, the part where we're detecting if we're inside SELECT clause no more works in all cases.

This PR is also a follow-up to the comment in #364


inferrinizzard (approved):
Agreed in that this would be much easier with a parser approach, will restore this feature later eventually

Issue #365, opened 2022-08-01
[FORMATTING] Comments move their lines after formatting


primeapple:
**Input data**
```sql
select
  name, -- student name
  age -- student age
from
  students
-- ending comment
/* another ending comment */
```

**Expected Output**

```sql
select
  name, -- student name
  age -- student age
from
  students
-- ending comment
/* another ending comment */
```

**Actual Output**

```sql
select
  name,
  -- student name
  age -- student age
from
  students -- ending comment
  /* another ending comment */
```

**Usage**

I made these using the default settings on https://sql-formatter-org.github.io/sql-formatter/

I would be interested in getting this fixed, if you could guide me in the right direction. Seems to be similar to https://github.com/sql-formatter-org/sql-formatter/issues/50

nene:
Currently the formatter works so that it's blind regarding original whitespace. Basically it sees the above code like this:

```
<SELECT> <IDENT name> <COMMA> <COMMENT "student name"> <IDENT age> <COMMENT "student age">
```

Or to put it another way. For the formatter these two select statements look exactly the same:

```SQL
select
  name, -- student name
  age -- student age
from
  students; -- comment

select
  name,
  -- student name
  age
  -- student age
from
  students;
  -- comment
```

This is currently by design. The formatter preserves no original formatting, completely reindenting the code regardless of how the whitespace was distributed originally. In that regard, it's expected that comments might be positioned differently than they were in the input. Like in your example there's identifier `age` followed by comment and identifier `students` followed by comment, but you're expecting them to be formatted differently. That's not really compatible with the current design of the formatter.

Another question is: how should the comments be placed? There's really two main options here: a) keep them at the end of preceding line, b) place each comment on a separate line. At the moment the formatter leans towards the first approach, though not quite purely. It basically just places them at the position where the next ordinary (non-comment) token would go.

Note: the comments placement works much better when organized like so:

```sql
select
  -- student name
  name,
  -- student age
  age
from
  -- comment
  students;
```

In general, I think it might be a better approach to just place all comments on a separate line. In my experience, it's nearly always possible to organize your comments so that each one is on a separate line, but it's not really possible to organize all comments so that they all are at the end of a preceding line (e.g. when there are multiple single-line comments after one another).

primeapple:
Thank you for your answer!


How I see it there are three options for comments, that reference a single line:
1. making them above the line
2. making them behind the content of the line
3.  making them below the line

[...]

Issue #329, opened 2022-07-18
Feature Request: Add support for linesBetweenQueries="preserve"


m-ahmadkhan:
**Describe the Feature**
This feature will help for SQL scripting. For example, I don't link putting too many lines between the DECLARE statements, if the total number of variables defined are too many. And sometimes, I like to group a few variables based on the purpose that they're defined for (no lines in between, but 1 or 2 lines between groups). But I like separating the actual queries with 1 or 2 lines.

Take the following script as an example:
```sql
DECLARE target_word STRING DEFAULT 'methinks';
DECLARE corpus_count INT64;
DECLARE word_count INT64;

SET (corpus_count, word_count) = (
  SELECT AS STRUCT COUNT(DISTINCT corpus), SUM(word_count)
  FROM bigquery-public-data.samples.shakespeare
  WHERE LOWER(word) = target_word
);

SELECT
  FORMAT('Found %d occurrences of "%s" across %d Shakespeare works',
         word_count, target_word, corpus_count) AS result;
```

The formatter would format it as:
```sql
DECLARE
  target_word STRING DEFAULT 'methinks';

DECLARE
  corpus_count INT64;

DECLARE
  word_count INT64;

SET
  (corpus_count, word_count) = (
    SELECT
      AS STRUCT COUNT(DISTINCT corpus),
      SUM(word_count)
    FROM
      bigquery-public-data.samples.shakespeare
    WHERE
      LOWER(word) = target_word
  );

SELECT
  FORMAT(
    'Found %d occurrences of "%s" across %d Shakespeare works',
    word_count,
    target_word,
    corpus_count
  ) AS result;
```

But I'd like it better if it were:
```sql
DECLARE
  target_word STRING DEFAULT 'methinks';
DECLARE
  corpus_count INT64;
DECLARE
  word_count INT64;

SET
  (corpus_count, word_count) = (
    SELECT
      AS STRUCT COUNT(DISTINCT corpus),
      SUM(word_count)
    FROM
      bigquery-public-data.samples.shakespeare
    WHERE
      LOWER(word) = target_word
  );

SELECT
  FORMAT(
    'Found %d occurrences of "%s" across %d Shakespeare works',
    word_count,
    target_word,
    corpus_count
  ) AS result;
```

nene:
Makes sense.

[...]

PR #374, opened 2022-08-04
Add location attributes to token object

inferrinizzard:
adds the following location attributes to tokens:
- `begin`: raw index in overall query string
- `end`: where token ends in query string


.editorconfig
```diff
@@ -7,3 +7,4 @@ trim_trailing_whitespace = true
 insert_final_newline = true
 indent_style = space
 indent_size = 2
+use_tabs = false
```
nene, line 10: What's the purpose of `use_tabs`? I don't see such config documented in [editorconfig docs](https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties).

We already have `indent_style = space`.

src/lexer/regexUtil.ts
```diff
@@ -4,6 +4,7 @@ import type { PrefixedQuoteType } from './regexTypes';
 export const escapeRegExp = (string: string) => string.replace(/[.*+?^${}()|[\]\\]/gu, '\\$&');
 
 export const WHITESPACE_REGEX = /\s+/uy;
+export const LINEBREAK_REGEX = /\v|\n|\r\n/g;
```
nene, line 7: A vertical tab? Really? If somebody uses the vertical tab character, they're really asking for trouble.

Also, we should match `\r\n` and `\n` in different order. Currently `\n` always matches first, so `\r\n` never matches.

src/lexer/TokenizerEngine.ts
```diff
@@ -145,12 +159,18 @@ export default class TokenizerEngine {
     if (matches) {
       const matchedToken = matches[0];
 
+      const tokenStart = this.index;
+      const tokenColInLine = this.col;
       // Advance current position by matched token length
       this.index += matchedToken.length;
+      this.col += matchedToken.length;
```
nene, line 166: Tokens can be multi-line. Like strings, multi-line comments, and our KEYWORD ... SEQUENCE tokens.

src/lexer/token.ts
```diff
@@ -35,13 +35,16 @@ export interface Token {
   raw: string; // The raw original text that was matched
   text: string; // Cleaned up text e.g. keyword converted to uppercase and extra spaces removed
   key?: string;
+  index: number; // index of the token in the whole query string
+  line?: number; // line number that the token appears on
+  col?: number; // column within the line on which the token appears
```
nene, line 40: When dealing with line and column numbers, one always needs to specify whether these are zero- or one-based. i.e. is the first line number `0` or `1`. 

Usually one wants to use one-based indexing when displaying the info to user, while for internal usage, zero-based indexing is better for doing calculations with.

nene (commented):
We should record both start and end locations of a token. For example babel parser uses:

```ts
loc: {
  start: {line, column, index},
  end: {line, column, index},
},
```

While Esprima parser uses a simple array of start-end indexes:

```ts
range: [15, 20],
```

I'm also not sure whether we actually need line and column numbers. Do you have a concrete plan on how you're planning to use this info?

The line and column number can also be calculated afterwards when index is known: one just needs to count lines until the index in original source code.

nene (commented):
You haven't responded to my question of how are you planning to use this location information. Currently this is all effectively dead code, as we don't use this info anywhere. To judge whether this is a good approach, I'd need to know where exactly are you planning to head with this.

I know that the start/end indexes would be useful for associating comments with AST nodes. So I'm on board with that.

For the line and column numbers I can currently only think of one usage: reporting errors. For example if our Tokenizer currently fails, we report `Parse error: Unexpected "some text"`. Would be nice to say at which line and column the problem happened. But for that we don't really need to store the line and column positions inside each token - we could instead compute that on-demand, which would save us the overhead of computing line & col positions which we wouldn't use most of the time at all. But perhaps you had some different use case in mind?

I also think it would be better to group this data inside one field of a Token instead of adding 4 new fields. That should also simplify doing operations with this location info later. Like instead of having to extract 4 fields from a token to pass them to `doSomethingWithLocation()` function, one would only need to extract a single field.

src/lexer/TokenizerEngine.ts
```diff
@@ -52,13 +53,32 @@ export default class TokenizerEngine {
...
+    // if whitespace contains linebreaks
+    if (LINEBREAK_REGEX.test(token)) {
+      this.line++;
+      // increment line for each newline match
+      let lastIndex = 1;
+      while (LINEBREAK_REGEX.exec(token) !== null) {
+        this.line++;
+        lastIndex = LINEBREAK_REGEX.lastIndex;
+      }
+      this.col = token.length - lastIndex + 1;
```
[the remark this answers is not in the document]
inferrinizzard: the + 1 is actually for the 1-based col index, not for the line match length
setting the `lastIndex` to 1 is to account for the line char but you're right that it doesn't account for `\r\n`

inferrinizzard:
the line and col attributes could be used for more easily formatting in `preserve` modes, such as that one other PR regarding the placement of line comments that were previously inline - the raw `start` and `end` attributes wouldn't encode which line the comment was originally on

src/lexer/TokenizerEngine.ts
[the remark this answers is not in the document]
inferrinizzard: the initial line increment is because `REGEX.test` and `REGEX.exec` have an interaction where the test counts as the first exec - ie. if there is only 1 line break, the exec loop would never start

nene:
I suppose you're referring to this comment-placement issue: #365

Another whitespace-preserving related issue is #329

I think these both might be better solved by bringing back the `precedingWhitespace` field to tokens. Though, it might also be that storing line numbers is a better approach. However the current implementation won't work for that as we'd need to also store the line number of the token end (so we could compare it with next token start to determine if there are newlines in between).

Perhaps though it would be better to leave this line/column number part completely out for now. As we don't yet know if and how exactly we're going to use it - it's hard to design an API for a future unknown usage. I think it might be better to add this line/col-numbers support together with the feature that actually uses it. It's not really that complex of a functionality to bundle together with another feature, and perhaps we end up not using it at all.

src/lexer/token.ts
```diff
@@ -35,13 +35,21 @@ export interface Token {
...
  * For use as a "missing token"
  * e.g. in lookAhead and lookBehind to avoid dealing with null values
  */
-export const EOF_TOKEN = { type: TokenType.EOF, raw: '«EOF»', text: '«EOF»' };
+export const EOF_TOKEN: Token = {
+  type: TokenType.EOF,
+  raw: '«EOF»',
+  text: '«EOF»',
+  start: Number.POSITIVE_INFINITY,
+  end: Number.POSITIVE_INFINITY,
```
nene, line 51: FYI, you could have just used:

```suggestion
  start: Infinity,
  end: Infinity,
```

nene (approved):

PR #375, opened 2022-08-05
Overhaul of DDL syntax variants

nene:
Covering the following:

- CREATE TABLE
- CREATE VIEW
- DROP TABLE
- ALTER TABLE

There are of course more, but these are the ones so far documented in our wiki.

The grammar variations of ALTER TABLE are huge. I only covered the very basic and most widespread syntax:

- ADD COLUMN
- DROP COLUMN
- ALTER COLUMN
- RENAME COLUMN
- RENAME TO
- MODIFY

The tests again don't attempt to cover all the variations in syntax, only the most important bits.


test/postgresql.test.ts
```diff
@@ -117,4 +124,50 @@ describe('PostgreSqlFormatter', () => {
...
+        foo
+      DROP DEFAULT
+      ;
+
+      ALTER TABLE
+        t
+      ALTER COLUMN
+        foo
+      SET NOT NULL
+      ;
```
nene, line 164: Looks like we have a bug in formatting of semicolons. Won't fix in this PR. Will tackle it separately.

inferrinizzard:
Why DDL this time and not Data Definition like the other 2 PRs ?

nene:
Valid point :) will change.

test/bigquery.test.ts
```diff
@@ -441,56 +394,23 @@ describe('BigQueryFormatter', () => {
...
+          field_2 INT64
+        )
+        OPTIONS(
+          format = 'CSV',
+          uris = ['gs://bucket/path1.csv']
+        )`;
+      const expected = dedent`
+        CREATE EXTERNAL TABLE
+          dataset.CsvTable
+        WITH PARTITION COLUMNS
```
inferrinizzard, line 411: this output seems worse than the input somehow

inferrinizzard (approved):

test/bigquery.test.ts
nene: I agree.

There's a larger problem with our current formatting style, which wants to format everything as:

```
SOME CLAUSE
  content
ANOTHER CLAUSE
  clause content
```

While it often makes more sense to not create an indented block, but instead place things one the same line. For example `SORT BY` and `LIMIT` are more commonly formatted on the same line:

```sql
SORT BY col1 ASC, col2 DESC
LIMIT 10, 8
```

Similarly `CREATE TABLE` is usually formatted as:

```sql
CREATE TABLE customer (
  id INT NOT NULL,
  name VARCHAR NOT NULL,
  address VARCHAR
);
```

But the formatter currently produces:

```sql
CREATE TABLE
  customer (
    id INT NOT NULL,
    name VARCHAR NOT NULL,
    address VARCHAR
  );
```

...which isn't really how anybody in would format it manually.

We'll need to do some rethinking and reorganization of the current formatting approach.

PR #376, opened 2022-08-06
Fix semicolon-on-separate-line bug

nene:
Add `WS.NO_NEWLINE` and `WS.MANDATORY_NEWLINE` to Layout engine. Also cover Layout class with tests.

Use these new features to:

1. Eliminate trailing newline before inserting semicolon.
2. Add non-removable newline at the end of a line-comment.


inferrinizzard:
Looks good, can you also add tests for having the semicolon on a newline ?

nene:
You mean a test for line-comment followed by semicolon?

There already exists one for that: https://github.com/sql-formatter-org/sql-formatter/blob/master/test/features/comments.ts#L59

inferrinizzard:
No, I mean a standard statement followed by newline
ie.
```sql
SELECT *
FROM tbl
;
```

nene:
So, you mean like: https://github.com/sql-formatter-org/sql-formatter/blob/master/test/options/newlineBeforeSemicolon.ts#L24

nene:
I merged this in, as you approved this in general, and the only question is about an additional test.

Please do say if the test you were thinking of is still missing.

inferrinizzard:
I meant with the new layout test file but I suppose the existing test is fine

nene:
> I meant with the new layout test file but I suppose the existing test is fine

Well, for that there is a test for `MANDATORY_NEWLINE` followed by `NO_NEWLINE`. That's really the part that was needed to make it work. Doesn't really make much sense to test an actual semicolon in there - for Layout all plain strings (non-whitespace related) are the same.

PR #379, opened 2022-08-06
S2db support

mgiannakopoulos:
Adding support for `singlestoredb` sql dialect.


nene:
Hi, thanks for contributing a new dialect support.

As I understand, this SingleStoreDB uses mostly the same syntax as MySQL. Though it's a bit tricky to confirm as SingleStoreDB docs don't document the full SQL syntax that's supported. I found a page that lists [unsupported MySQL features](https://docs.singlestore.com/managed-service/en/developer-resources/unsupported-mysql-features.html). Is there perhaps a place that documents SELECT syntax?

My first question is, how exactly did you put together this configuration for SingleStoreDB. It seems that a lot of it is copy-pasted from the MySQL configuration. How much of the MySQL syntax that you copied over did you actually verify?

src/languages/singlestoredb/singlestoredb.functions.ts
```diff
@@ -0,0 +1,302 @@
...
+    'VAR_POP',
+    'VAR_SAMP',
+    'VECTOR_SUB',
+    'VERSION',
+    'WEEK',
+    'WEEKDAY',
+    'WEEKOFYEAR',
+    'YEAR',
+    // Data types with parameters
+    // https://dev.mysql.com/doc/refman/8.0/en/data-types.html
```
nene, line 272: So, is the following list a list of MySQL data types or a list of SingleStoreDB data types?

src/languages/singlestoredb/singlestoredb.formatter.ts
```diff
@@ -0,0 +1,296 @@
...
+  'FROM',
+  'GROUP BY',
+  'HAVING',
+  'INSERT INTO',
+  'LIMIT',
+  'OFFSET',
+  'ORDER BY',
+  'WHERE',
+  'WINDOW',
+  'PARTITION BY',
```
nene, line 239: The `reservedCommands` list contains several duplicates like `SELECT`, `FROM`, `LIMIT`, `GROUP BY`, `INSERT` ...

src/languages/singlestoredb/singlestoredb.formatter.ts
```diff
@@ -0,0 +1,296 @@
...
+  'LOAD XML',
+  'LOCK INSTANCE FOR BACKUP',
+  'LOCK TABLES',
+  'MASTER_POS_WAIT',
+  'OPTIMIZE TABLE',
+  'PREPARE',
+  'PURGE BINARY LOGS',
+  'REBALANCE PARTITIONS',
+  'RELEASE SAVEPOINT',
+  'RENAME TABLE',
```
nene, line 130: This looks suspicious. Above we already support one syntax for renaming tables:

```sql
ALTER TABLE foo RENAME TO bar;
```

How does this other `RENAME TABLE` syntax work?

src/languages/singlestoredb/singlestoredb.formatter.ts
```diff
@@ -0,0 +1,296 @@
...
+  'DROP [COLUMN]',
+  'MODIFY [COLUMN]',
+  'RENAME [TO | AS]',
+
+  // https://docs.singlestore.com/managed-service/en/reference/sql-reference.html
+  'ADD AGGREGATOR',
+  'ADD LEAF',
+  'AGGREGATOR SET AS MASTER',
+  'ALTER DATABASE',
+  'ALTER EVENT',
```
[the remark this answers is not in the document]
mgiannakopoulos: For the first three commands here is the reference: https://docs.singlestore.com/managed-service/en/reference/sql-reference/operations-that-take-either-a-database-or-a-workspace-lock/operations-that-take-workspace-locks.html

These are specific commands that control the ability to manually scale-up and scale-down the cluster of machines that are running the SingleStoreDB.

Here is the documentation for "ALTER DATABASE": https://docs.singlestore.com/managed-service/en/reference/sql-reference/data-definition-language-ddl/alter-database.html

Yes, for "ALTER EVENT" I will need to remove it. "ALTER PIPELINE" is a valid syntax: https://docs.singlestore.com/managed-service/en/reference/sql-reference/pipelines-commands/alter-pipeline.html

src/languages/singlestoredb/singlestoredb.formatter.ts
```diff
@@ -0,0 +1,296 @@
...
+  '{LEFT | RIGHT | FULL} [OUTER] JOIN',
+  '{INNER | CROSS} JOIN',
+  'NATURAL {LEFT | RIGHT} [OUTER] JOIN',
+  // non-standard joins
+  'STRAIGHT_JOIN',
+]);
+
+// https://dev.mysql.com/doc/refman/8.0/en/
+export default class SingleStoreDBFormatter extends Formatter {
+  static operators = ['=', '<=>', '>', '>=', '<', '<=', '!=', '<>'];
```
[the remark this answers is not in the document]
mgiannakopoulos: Need to verify `&&` and `||` operators. My expectaion is that both of them are not supported by will double-check.

src/languages/singlestoredb/singlestoredb.functions.ts
mgiannakopoulos: Yes, according to the reserved keywords: https://docs.singlestore.com/managed-service/en/reference/sql-reference/restricted-keywords/list-of-restricted-keywords.html

src/languages/singlestoredb/singlestoredb.formatter.ts
mgiannakopoulos: Okay, so if I understand it correctly, you are saying that if I have a definition like `INSERT [INTO]`, then it is okay to delete the `INSERT INTO` definition later in the `reservedCommands` section.

src/languages/singlestoredb/singlestoredb.formatter.ts
mgiannakopoulos: Yeah, this is wrong, will need to fix it.

mgiannakopoulos:
@nene Yes, this is correct, SingleStoreDB started as a fork of MySQL back in `2011` but since then a couple of things have changes and some support that `MySQL` offers was narrowed down for SingleStoreDB. I started by the copy-pasting approach from `MySQL` but then I deleted all `keywords` and `functions`. I went though the supported keywords documented here: https://docs.singlestore.com/managed-service/en/reference/sql-reference/restricted-keywords/list-of-restricted-keywords.html

After that for the `reserved commands` I searched directly the reference about DML and DDL constructs:
* SELECT construct: https://docs.singlestore.com/managed-service/en/reference/sql-reference/data-manipulation-language-dml/select.html
* CREATE construct: https://docs.singlestore.com/managed-service/en/reference/sql-reference/data-definition-language-ddl/create-table.html
* INSERT construct: https://docs.singlestore.com/managed-service/en/reference/sql-reference/data-manipulation-language-dml/insert.html
* ALTER construct: https://docs.singlestore.com/managed-service/en/reference/sql-reference/data-definition-language-ddl/alter-table.html, https://docs.singlestore.com/managed-service/en/reference/sql-reference/pipelines-commands/alter-pipeline.html, https://docs.singlestore.com/managed-service/en/reference/sql-reference/data-definition-language-ddl/alter-view.html, https://docs.singlestore.com/managed-service/en/reference/sql-reference/data-definition-language-ddl/alter-database.html, https://docs.singlestore.com/managed-service/en/reference/sql-reference/security-management-commands/alter-user.html
* UPDATE construct: https://docs.singlestore.com/managed-service/en/reference/sql-reference/data-manipulation-language-dml/update.html
* DELETE construct: https://docs.singlestore.com/managed-service/en/reference/sql-reference/data-manipulation-language-dml/delete.html

In general I searched a lot via the SQL reference documentation from the official SingleStoreDB documentation. After I land this PR, will add additional tests for the customized features that SingleStoreDB supports.

nene:
Thanks for the updates. I have one request for the future: please avoid using force-push. It makes it very hard to tell what has changed since I last reviewed your code.

src/languages/singlestoredb/singlestoredb.functions.ts
nene: So, why do we link to the MySQL docs here? Instead of e.g. https://docs.singlestore.com/managed-service/en/reference/sql-reference/data-types.html

src/languages/singlestoredb/singlestoredb.formatter.ts
nene: Yep, you're correct.

test/singlestoredb.test.ts
```diff
@@ -0,0 +1,43 @@
...
+    modify: true,
+    renameTo: true,
+    renameColumn: false,
+  });
+  supportsDeleteFrom(format);
+  supportsInsertInto(format, { withoutInto: true });
+  supportsUpdate(format);
+  supportsTruncateTable(format, { withoutTable: true });
+  supportsBetween(format);
+  supportsParams(format, { positional: false });
```
nene, line 42: Really there's no point of calling `supportsParams()` in this test at all.

src/languages/singlestoredb/singlestoredb.formatter.ts
```diff
@@ -0,0 +1,267 @@
...
+      ],
+      identTypes: ['``'],
+      identChars: { first: '$', rest: '$', allowFirstCharNumber: true },
+      variableTypes: [
+        { regex: '@[A-Za-z0-9_.$]+' },
+        { quote: '""', prefixes: ['@'], requirePrefix: true },
+        { quote: "''", prefixes: ['@'], requirePrefix: true },
+        { quote: '``', prefixes: ['@'], requirePrefix: true },
+      ],
+      positionalParams: false,
```
nene, line 250: Better to just delete this line, for consistency with other *Formatter classes.

src/languages/singlestoredb/singlestoredb.formatter.ts
```diff
@@ -0,0 +1,267 @@
...
+  'JOIN',
+  '{LEFT | RIGHT | FULL} [OUTER] JOIN',
+  '{INNER | CROSS} JOIN',
+  'NATURAL {LEFT | RIGHT} [OUTER] JOIN',
+  // non-standard joins
+  'STRAIGHT_JOIN',
+]);
+
+// https://dev.mysql.com/doc/refman/8.0/en/
+export default class SingleStoreDBFormatter extends Formatter {
```
nene, line 225: Should rename to:
```suggestion
export default class SingleStoreDbFormatter extends Formatter {
```

For consistency with other formatter classes like `Db2Formatter`, `MariaDbFormatter`.

src/languages/singlestoredb/singlestoredb.formatter.ts
```diff
@@ -0,0 +1,267 @@
...
+  'START PIPELINE',
+  'START REPLICA',
+  'START SLAVE',
+  'START TRANSACTION',
+  'STOP GROUP_REPLICATION',
+  'STOP PIPELINE',
+  'STOP REPLICA',
+  'STOP REPLICATING',
+  'STOP SLAVE',
+  'TABLE',
```
nene, line 191: I suspect there's no `TABLE`-statement or `TABLE`-clause within another statement.

I know MySQL also lists `TABLE` here, which is also wrong. There's unfortunately quite a bit of such mistakes in the existing `reservedCommands` lists for all dialects. I'm currently in the process of reviewing these lists and throwing out incorrect stuff. Only the top of the list (until an empty line) is the reviewed part, that I'm more certain of. The rest should be taken with grain of salt, and better not to be copied over without checking.

src/languages/singlestoredb/singlestoredb.keywords.ts
```diff
@@ -0,0 +1,852 @@
...
+    'XACT_ID',
+    'XOR',
+    'YEAR',
+    'YEAR_MONTH',
+    'YES',
+    'ZEROFILL',
+    'ZONE',
+  ],
+  constraints: ['ON DELETE', 'ON UPDATE'],
+  charset: ['CHARACTER SET'],
```
nene, line 851: These two lines should be moved inside the `*.formatter.ts` file.

This part was changed recently. See the `mysql.formatter.ts` file for reference or the PR #363

src/languages/singlestoredb/singlestoredb.keywords.ts
```diff
@@ -0,0 +1,852 @@
+import { flatKeywordList } from '../../utils';
+
+export const keywords = flatKeywordList({
+  // https://docs.singlestore.com/managed-service/en/reference/sql-reference.html
+  // https://docs.singlestore.com/managed-service/en/reference/sql-reference/restricted-keywords/list-of-restricted-keywords.html
```
nene, line 5: I wonder over the meaning of this "restricted keywords" list in SingleStoreDB documentation.

Are all these keywords actually supported by SingleStoreDB in the sense that they serve some function, or are they just reserved, in the sense that one can't use them as names of identifiers? The docs of SingleStoreDB aren't really clear on this matter:

> Keywords that are restricted (or reserved) require special treatment if they need to be used as an identifier such as a table or column name. This is because their specific meaning **could** cause issues when used in the context of such a name.

src/languages/singlestoredb/singlestoredb.formatter.ts
```diff
@@ -0,0 +1,267 @@
...
+// https://dev.mysql.com/doc/refman/8.0/en/
+export default class SingleStoreDBFormatter extends Formatter {
+  static operators = ['=', '<=>', '>', '>=', '<', '<=', '!=', '<>'];
+
+  tokenizer() {
+    return new Tokenizer({
+      reservedCommands,
+      reservedSetOperations,
+      reservedJoins,
+      reservedDependentClauses: ['WHEN', 'ELSE', 'ELSEIF'],
```
nene, line 233: I found no example of `ELSEIF` syntax from SingeStoreDB docs.

src/languages/singlestoredb/singlestoredb.formatter.ts
```diff
@@ -0,0 +1,267 @@
...
+      reservedLogicalOperators: ['AND', 'OR', 'XOR'],
```
nene, line 234: Failed to find an example of `XOR` keyword support from SingleStoreDB docs.

src/languages/singlestoredb/singlestoredb.formatter.ts
```diff
@@ -0,0 +1,267 @@
...
+      reservedKeywords: keywords,
+      reservedFunctionNames: functions,
+      // TODO: support _ char set prefixes such as _utf8, _latin1, _binary, _utf8mb4, etc.
+      stringTypes: [
+        { quote: "''", prefixes: ['B', 'N', 'X'] },
+        { quote: '""', prefixes: ['B', 'N', 'X'] },
+      ],
```
nene, line 241: Does this "TODO" also apply to SingleStoreDB?

Is the string syntax of SingleStoreDB documented somewhere? I failed to find such a place.

src/languages/singlestoredb/singlestoredb.formatter.ts
```diff
@@ -0,0 +1,267 @@
...
+      identTypes: ['``'],
+      identChars: { first: '$', rest: '$', allowFirstCharNumber: true },
```
nene, line 243: I found one mention of the backtick-quoting support from the docs. But I found no docs regarding the syntax of identifiers. Like, can they contain `$` character and begin with a number?

src/languages/singlestoredb/singlestoredb.formatter.ts
```diff
@@ -0,0 +1,267 @@
...
+      variableTypes: [
+        { regex: '@[A-Za-z0-9_.$]+' },
+        { quote: '""', prefixes: ['@'], requirePrefix: true },
+        { quote: "''", prefixes: ['@'], requirePrefix: true },
+        { quote: '``', prefixes: ['@'], requirePrefix: true },
+      ],
+      positionalParams: false,
+      lineCommentTypes: ['--', '#'],
```
nene, line 251: Found no mention of comment syntax from SingleStoreDB docs :(

src/languages/singlestoredb/singlestoredb.formatter.ts
```diff
@@ -0,0 +1,267 @@
...
+        { quote: '""', prefixes: ['B', 'N', 'X'] },
+      ],
+      identTypes: ['``'],
+      identChars: { first: '$', rest: '$', allowFirstCharNumber: true },
+      variableTypes: [
+        { regex: '@[A-Za-z0-9_.$]+' },
+        { quote: '""', prefixes: ['@'], requirePrefix: true },
+        { quote: "''", prefixes: ['@'], requirePrefix: true },
+        { quote: '``', prefixes: ['@'], requirePrefix: true },
+      ],
```
nene, line 249: Found no clear description of variable syntax from SingleStoreDB docs. Saw some examples of variables, so it's clear that they do begin with `@`, but whether you can also use `@"some var"` syntax... found no docs :(

src/languages/singlestoredb/singlestoredb.functions.ts
```diff
@@ -0,0 +1,302 @@
+import { flatKeywordList } from '../../utils';
+
+export const functions = flatKeywordList({
+  // https://docs.singlestore.com/managed-service/en/reference/sql-reference.html
+  // https://docs.singlestore.com/managed-service/en/reference/sql-reference/restricted-keywords/list-of-restricted-keywords.html
```
nene, line 5: There's no list of function names at this URL. Well, I presume the list of keywords also includes all function names, but it's not really of much help.

src/languages/singlestoredb/singlestoredb.formatter.ts
```diff
@@ -0,0 +1,267 @@
...
+  'SET',
+  // - delete:
+  'DELETE [FROM]',
+  // - truncate:
+  'TRUNCATE [TABLE]',
+  // Data definition
+  'CREATE [ROWSTORE] [REFERENCE | TEMPORARY | GLOBAL TEMPORARY] TABLE [IF NOT EXISTS]',
+  'CREATE [OR REPLACE] [TEMPORARY] PROCEDURE [IF NOT EXISTS]',
+  'CREATE [OR REPLACE] [EXTERNAL] FUNCTION',
+  'DROP [TEMPORARY] TABLE [IF EXISTS]',
```
nene, line 34: Should also move `CREATE VIEW` into this section like in other `formatter.ts` files.

nene (changes requested):
Thanks for the links to SELECT syntax. For some reason failed to find it on my own.

I'm still having trouble locating information about more details of the supported syntax. Especially things like, what kind of quotes are supported and how are they escaped.

I've documented the query syntax of SingleStoreDB in our wiki: https://github.com/sql-formatter-org/sql-formatter/wiki

If you'd like to put in the effort, it would be nice if you could update other parts of the wiki to reflect SingleStoreDB syntax. The important thing there is to have links to some sources, as without citations the info there would be hard to verify. In the past we have accepted several contributions of new SQL dialect support, discovering later that what the implementation claims to support is not actually what the official docs say. Now we're trying to be more organized.

test/singlestoredb.test.ts
inferrinizzard: Should add more documentation on _why_ this is the case for outsiders building new tests since it may not be as intuitive if they haven't been working with the code as much as we have

src/languages/singlestoredb/singlestoredb.formatter.ts
mgiannakopoulos: Will need to try it out...

src/languages/singlestoredb/singlestoredb.functions.ts
mgiannakopoulos: Here is the list of all the functions that are documented:
* https://docs.singlestore.com/managed-service/en/reference/sql-reference/vector-functions/vector-functions.html
* https://docs.singlestore.com/managed-service/en/reference/sql-reference/window-functions/window-functions.html
* https://docs.singlestore.com/managed-service/en/reference/sql-reference/string-functions/string-functions.html
* https://docs.singlestore.com/managed-service/en/reference/sql-reference/conditional-functions/conditional-functions.html
* https://docs.singlestore.com/managed-service/en/reference/sql-reference/numeric-functions/numeric-functions.html
* https://docs.singlestore.com/managed-service/en/reference/sql-reference/geospatial-functions/geospatial-functions.html
* https://docs.singlestore.com/managed-service/en/reference/sql-reference/json-functions/json-functions.html
* https://docs.singlestore.com/managed-service/en/reference/sql-reference/information-functions/information-functions.html
* https://docs.singlestore.com/managed-service/en/reference/sql-reference/aggregate-functions/aggregate-functions.html
* https://docs.singlestore.com/managed-service/en/reference/sql-reference/time-series-functions/time-series-functions.html
* https://docs.singlestore.com/managed-service/en/reference/sql-reference/identifier-generation-functions.html
* https://docs.singlestore.com/managed-service/en/reference/sql-reference/date-and-time-functions/date-and-time-functions.html
* https://docs.singlestore.com/managed-service/en/reference/sql-reference/distinct-count-estimation-functions.html
* https://docs.singlestore.com/managed-service/en/reference/sql-reference/full-text-search-functions/full-text-search-functions.html
* https://docs.singlestore.com/managed-service/en/reference/sql-reference/regular-expression-functions.html

src/languages/singlestoredb/singlestoredb.formatter.ts
mgiannakopoulos: Not sure, need to investigate. Here is the supported character sets and the collations:
* https://docs.singlestore.com/managed-service/en/reference/sql-reference/character-encoding/character-sets-supported.html
* https://docs.singlestore.com/managed-service/en/reference/sql-reference/character-encoding/collations-supported.html

src/languages/singlestoredb/singlestoredb.formatter.ts
```diff
@@ -0,0 +1,274 @@
...
+  'NATURAL {LEFT | RIGHT} [OUTER] JOIN',
+  // non-standard joins
+  'STRAIGHT_JOIN',
+]);
+
+const reservedPhrases = ['ON DELETE', 'ON UPDATE', 'CHARACTER SET'];
+
+// https://docs.singlestore.com/managed-service/en/reference/sql-reference/comparison-operators-and-functions.html
+export default class SingleStoreDbFormatter extends Formatter {
+  static operators = ['=', '<=>', '>', '<'];
```
nene, line 233: We should definitely include `<<` and `>>`

- https://docs.singlestore.com/managed-service/en/reference/sql-reference/numeric-functions/bitwise-left-shift-----.html
- https://docs.singlestore.com/managed-service/en/reference/sql-reference/numeric-functions/bitwise-right-shift-----.html

No need to include `=`, `<`, `>`. The single-letter operators are all detected automatically.

nene:
I did some testing of SingleStoreDB syntax and documented my findings:

- [identifiers](https://github.com/sql-formatter-org/sql-formatter/wiki/identifiers)
- [strings](https://github.com/sql-formatter-org/sql-formatter/wiki/Strings)
- [variables](https://github.com/sql-formatter-org/sql-formatter/wiki/variables)
- [operators](https://github.com/sql-formatter-org/sql-formatter/wiki/Operators)

src/languages/singlestoredb/singlestoredb.formatter.ts
mgiannakopoulos: I have verified that both `--` and `#` are used as placeholders for line comments. Also `/*` and `*/` are used for multi-line comments.

src/languages/singlestoredb/singlestoredb.formatter.ts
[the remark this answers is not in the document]
mgiannakopoulos: I have verified that `&&` and `||`. I have fixed the issue.

src/languages/singlestoredb/singlestoredb.formatter.ts
mgiannakopoulos: Syntax like this is supported (I have just tested it):
```
CREATE TABLE people (`$name` text, `1age` bigint)
SELECT `$name` FROM people
```

src/languages/singlestoredb/singlestoredb.formatter.ts
mgiannakopoulos: Yes, `XOR` is not supported. I have removed it.

src/languages/singlestoredb/singlestoredb.formatter.ts
mgiannakopoulos: I have verified that `ELSEIF` is supported. Here is the example I have run:

```
DELIMITER // ;
CREATE OR REPLACE PROCEDURE courses_sp (course_code TEXT, section_number INT, number_students INT) AS   DECLARE
     code TEXT = UCASE(course_code); 
     num_students INT = number_students + 1;
     BEGIN
     IF '' = '' THEN INSERT INTO courses VALUES (code, section_number, num_students);
     ELSEIF 1 = 1 THEN INSERT INTO courses VALUES(code, section_number, num_students);
     END IF;
     END //
DELIMITER ; //
```
Here is the execution:
```
memsql> show procedures;
+-----------------------+------------------+---------+
| Procedures_in_example | Routine Lifetime | Definer |
+-----------------------+------------------+---------+
| courses_sp            | Permanent        | root@%  |
+-----------------------+------------------+---------+
1 row in set (0.001 sec)

memsql> CALL courses_sp("cs-101",1,13);
Query OK, 0 rows affected (0.504 sec)

memsql> select * from courses;
+-------------+----------------+-----------------+
| course_code | section_number | number_students |
+-------------+----------------+-----------------+
| CS-101      |              1 |              14 |
+-------------+----------------+-----------------+
1 row in set (0.192 sec)
```

src/languages/singlestoredb/singlestoredb.formatter.ts
mgiannakopoulos: I have tried the following successful example with back-ticks 
```
SELECT 'hello' INTO @`S_Course`;
SELECT @`S_course`;
```
I am getting errors if I try to use `@'S_Course'` or `@"S_Course"` syntax.

test/singlestoredb.test.ts
```diff
@@ -0,0 +1,43 @@
...
+import supportsParams from './options/param';
+import supportsStrings from './features/strings';
+import supportsTruncateTable from './features/truncateTable';
+import supportsUpdate from './features/update';
+
+describe('SingleStoreDbFormatter', () => {
+  const language = 'singlestoredb';
+  const format: FormatFn = (query, cfg = {}) => originalFormat(query, { ...cfg, language });
+
+  // behavesLikeMariaDbFormatter(format);
```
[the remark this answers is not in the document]
mgiannakopoulos: Added the aforementioned tests. After testing, `NATURAL JOINS` are not supported so I excluded them, however I enabled `FULL` joins. As far as operators are concerned, I have removed the unsupported `XOR`.

src/languages/singlestoredb/singlestoredb.keywords.ts
mgiannakopoulos: I would say that all of them they are acting as reserved keywords. A subset of those is used as functions. (I have listed a couple of links in another discussion about which functions are documented).

src/languages/singlestoredb/singlestoredb.formatter.ts
```diff
@@ -230,7 +233,7 @@ const reservedPhrases = ['ON DELETE', 'ON UPDATE', 'CHARACTER SET'];
 
 // https://docs.singlestore.com/managed-service/en/reference/sql-reference/comparison-operators-and-functions.html
 export default class SingleStoreDbFormatter extends Formatter {
-  static operators = ['~', '<=>', '<<', '>>', '&&', '||'];
+  static operators = ['~', ':=', '<=>', '<<', '>>', '&&', '||'];
```
nene, line 236: Could you reference a documentation for the `:=` operator, or provide an example of how it's used in SingleStoreDB.

src/languages/singlestoredb/singlestoredb.formatter.ts
```diff
@@ -221,7 +224,7 @@ const reservedJoins = expandPhrases([
   'JOIN',
   '{LEFT | RIGHT | FULL} [OUTER] JOIN',
   '{INNER | CROSS} JOIN',
-  'NATURAL {LEFT | RIGHT} [OUTER] JOIN',
+  '{LEFT | RIGHT} [OUTER] JOIN',
```
nene, line 227: Why did you opt to remove NATURAL JOIN support?

- [the official docs](https://docs.singlestore.com/managed-service/en/reference/sql-reference/data-manipulation-language-dml/select.html#join-and-subqueries) say it's supported.
- I even tested it out and it worked just fine.

src/languages/singlestoredb/singlestoredb.formatter.ts
nene: Yep.

src/languages/singlestoredb/singlestoredb.formatter.ts
nene: Perhaps you made the mistake of testing `NATURAL JOIN`, which isn't supported, while e.g. `NATURAL LEFT JOIN` is.

src/languages/singlestoredb/singlestoredb.formatter.ts
nene: Found the answer. It's used in `SET @foo := 123;` syntax.

src/languages/singlestoredb/singlestoredb.formatter.ts
nene: Tested this and found that only `_binary"some string"` syntax works. No other char set prefix seems to be supported.

nene:
Hey!

To speed things up (as this ping-pong could drag on for a while with us living in pretty different timezones) I fixed up the remaining problems by myself and merged this whole feature in.

The additional commits can be seen here: #387

Thanks a lot for your efforts on this!

If something seems to have gone wrong with my fixups, please feel free to comment.

mgiannakopoulos:
Thanks @nene for all your help!

nene:
The SingleStoreDB support is now released in 9.1.0 🍾

PR #386, opened 2022-08-11
Add paramTypes config option

### LINKED ISSUE #367: Formatting issue when using named parameter(Java sql syntax)
**Input data**
```sql
select supplier_name,city from
(select * from suppliers join addresses on suppliers.address_id=addresses.id)
as suppliers
where supplier_id>:supplier_id
order by supplier_name asc,city desc;
```

**Expected Output**

```sql
select
  supplier_name,
  city
from
  (
    select
      *
    from
      suppliers
      join addresses on suppliers.address_id = addresses.id
  ) as suppliers
where
  supplier_id > :supplier_id
order by
  supplier_name asc,
  city desc;
```

**Actual Output**

```sql
select
  supplier_name,
  city
from
  (
    select
      *
    from
      suppliers
      join addresses on suppliers.address_id = addresses.id
  ) as suppliers
where
  supplier_id >: supplier_id
order by
  supplier_name asc,
  city desc;
```

**Usage**

Default query on the example page. 
For using sql in java with parameterized statements, ":supplier_id" syntax is used to pass named parameter's value
After formatting this is changed with a space in between ": supplier_id".

nene:
Added new config option `paramTypes` which allows overriding the prepared statement parameter types supported by default.

Did some slight restructuring of Tokenizer to allow config options to effect which kind of placeholder syntax is supported. This means that the parameter token regexes are no more cached, which however shouldn't be much of a performance concern as the really expensive part is building those large regexes form matching keywords. But will run some benchmarks to verify.

Fixes #367


nene:
Ran some performance tests and as expected the differences were pretty much non-existant:

- in Firefox the new version was a few percentages slower
- in Chrome the new version was a few percentages faster

So, no measurable difference really.

src/lexer/Tokenizer.ts
```diff
@@ -105,21 +95,57 @@ export default class Tokenizer {
...
-    this.engine = new TokenizerEngine(rules);
+  // These rules can't be blindly cached as the paramTypesOverrides object
+  // can differ on each invocation of the format() function.
+  private buildParamRules(
+    cfg: TokenizerOptions,
+    paramTypesOverrides: ParamTypes
+  ): Partial<Record<TokenType, TokenRule>> {
+    // Each dialect has its own default parameter types (if any),
+    // but these can be overriden by the user of the library.
+    const paramTypes = {
```
inferrinizzard, line 108: would it be better to merge the custom additions instead of just replacing the default behaviour ?

nene: I intentionally opted to not merge them to offer more flexibility:

- With overriding one can also turn off the built-in param types, in case they conflict with something. Not sure if that's needed, but with merging it wouldn't be possible.
- On the other hand the merging itself might not really be useful, as one would likely anyway use just one syntax for parameters, not mix several.
- The code for merging would also be more complex.

inferrinizzard (approved):

PR #390, opened 2022-08-11
Add SingleStoreDB to demo page

nene:


nene:
These are trivial changes, so I just go ahead and merge these right in.

PR #392, opened 2022-08-12
User defined formatter

### LINKED ISSUE #306: Feature Request: user defined formatter
**Describe the Feature**
I want to be able to define my own language which is a slight modification of an existing language. Something like this

```js
import { format, Formatter, TrinoFormatter, register } from 'sql-formatter'

class MyFormatter extends Formatter {
  /* copy pasted code from TrinoFormatter */
}
// or
class MyFormatter extends TrinoFormatter {
  cfg.quotes.push({quote: "{{}}"})
}

const result = format('select {{blah}}\n\n\n from mytable', { formatter: MyFormatter })
// or
register('my-trino-dialect', MyFormatter)
const result = format('select {{blah}}\n\n\n from mytable', { language: 'my-trino-dialect' })

```

We have a preprocessor/templating engine that puts stuff in our Trino queries, so for example the user writes

```sql
select {{col_name}} from mytable
```

and then we can have a web ui with a drop down for that value. We need to format this Trino+"{{}}}" dialect because it's what users write and format but it makes no sense to add this as its own language to sql-formatter. Currently, I have to fork this repo to do this.

nene:
Implemented as proposed in #306

Fixes #306

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -6,3 +6,5 @@ export type {
   LogicalOperatorNewline,
   FormatOptions,
 } from './FormatOptions';
+export { default as Formatter } from './formatter/Formatter';
+export { default as Tokenizer } from './lexer/Tokenizer';
--- a/src/index.ts
+++ b/src/index.ts
@@ -8,3 +8,4 @@ export type {
 } from './FormatOptions';
 export { default as Formatter } from './formatter/Formatter';
 export { default as Tokenizer } from './lexer/Tokenizer';
+export { expandPhrases } from './expandPhrases';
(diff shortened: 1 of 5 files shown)


src/index.ts
```diff
@@ -6,3 +6,5 @@ export type {
   LogicalOperatorNewline,
   FormatOptions,
 } from './FormatOptions';
+export { default as Formatter } from './formatter/Formatter';
+export { default as Tokenizer } from './lexer/Tokenizer';
```
tararoshan, line 10: Thank you so much for this quick PR! Could `{ expandPhrases }` also be exported, please? I find it to be very useful.

Thanks again!

inferrinizzard (approved):

src/index.ts
nene: Good point. Added it.

PR #395, opened 2022-08-16
Add more token types

nene:
Added the following new token types:

- `LIMIT` (previously one of the `RESERVED_COMMAND` tokens)
- `BETWEEN` (previously one of the `RESERVED_KEYWORD` tokens)
- `AND`, `OR`, `XOR` (previously `RESERVED_LOGICAL_OPERATOR`)
- `CASE`, `END` (previously named `RESERVED_CASE_START`, `RESERVED_CASE_END`)

Opted to eliminate the `RESERVED_` prefix from token types which represent just a single word. Makes the code simpler to read, especially when writing a Nearley grammar.

The purpose of all this is mostly the Nearley integration, where I need to distinguish e.g. `LIMIT` from other `RESERVED_COMMAND` tokens.

Functionality-wise everything works just like before.

### DIFF
--- a/src/lexer/TokenizerEngine.ts
+++ b/src/lexer/TokenizerEngine.ts
@@ -115,7 +115,9 @@ export default class TokenizerEngine {
       this.matchToken(TokenType.RESERVED_DEPENDENT_CLAUSE) ||
       this.matchToken(TokenType.RESERVED_JOIN) ||
       this.matchToken(TokenType.RESERVED_PHRASE) ||
-      this.matchToken(TokenType.RESERVED_LOGICAL_OPERATOR) ||
+      this.matchToken(TokenType.AND) ||
+      this.matchToken(TokenType.OR) ||
+      this.matchToken(TokenType.XOR) ||
       this.matchToken(TokenType.RESERVED_FUNCTION_NAME) ||
       this.matchToken(TokenType.RESERVED_KEYWORD)
     );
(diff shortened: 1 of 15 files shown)


src/lexer/TokenizerEngine.ts
```diff
@@ -106,14 +106,18 @@ export default class TokenizerEngine {
...
       this.matchToken(TokenType.RESERVED_SET_OPERATION) ||
       this.matchToken(TokenType.RESERVED_DEPENDENT_CLAUSE) ||
       this.matchToken(TokenType.RESERVED_JOIN) ||
       this.matchToken(TokenType.RESERVED_PHRASE) ||
-      this.matchToken(TokenType.RESERVED_LOGICAL_OPERATOR) ||
+      this.matchToken(TokenType.AND) ||
+      this.matchToken(TokenType.OR) ||
+      this.matchToken(TokenType.XOR) ||
       this.matchToken(TokenType.RESERVED_FUNCTION_NAME) ||
       this.matchToken(TokenType.RESERVED_KEYWORD)
```
inferrinizzard, line 122: +1 for converting this to
```ts
[TokenType.CASE, ..., TokenType.RESERVED_KEYWORD].any(tokenType => this.matchToken(tokenType))
```

nene:
> Why do these ones in particular deserve their own token type?

Currently no other reason than the fact that our existing Parser deals with these tokens and I need to distinguish these from other tokens in Nearley grammar to be able to port our existing parser functionality to Nearley.

There most likely will be need for several other token types in the future as well.

nene:
I merge this in, as I understand you approved this in general. And I'd like to try out these changes with Nearley.

PR #397, opened 2022-08-17
Change BetweenPredicate node field types

nene:
To be consistent with other AST nodes.

Discovered while working with Nearley that BetweenPredicate type expects single tokens as its sub-expressions:

```ts
type BetweenPredicate = {
  type: NodeType.between_predicate;
  betweenToken: Token;
  expr1: Token;
  andToken: Token;
  expr2: Token;
};
```

But we really want this:

```ts
type BetweenPredicate = {
  type: NodeType.between_predicate;
  betweenToken: Token;
  expr1: AstNode[];
  andToken: Token;
  expr2: AstNode[];
};
```

Already implemented it in Nearley branch the latter way, as that was a natural way to express it.

### DIFF
--- a/src/formatter/ExpressionFormatter.ts
+++ b/src/formatter/ExpressionFormatter.ts
@@ -120,16 +120,11 @@ export default class ExpressionFormatter {
   }
 
   private formatBetweenPredicate(node: BetweenPredicate) {
-    this.layout.add(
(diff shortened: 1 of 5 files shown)


nene:
I merge this in to move forward with Nearley integration. Feel free to still review it and express your thoughts.

PR #398, opened 2022-08-17
Add RESERVED_SELECT and ASTERISK tokens

nene:
These are needed to finish the final bit of functionality (distinguishing `*` in `SELECT *` from other uses of `*` operator) that Nearley parser currently lacks compared to our current parser.

It also eliminates the hack for detecting `SELECT` token using a regex.

Using the name `ASTERISK` as that's what [SQL standard](https://jakewheat.github.io/sql-overview/sql-2008-foundation-grammar.html#asterisk) uses.


nene:
I merge this in to move forward with Nearley integration. Feel free to still review it and express your thoughts.

PR #400, opened 2022-08-17
Remove braces and brackets from operators

nene:
Another small thing I discovered when working with Nearley integration.

This can cause ambiguity in parser, and really they shouldn't have been there in the first place.

This is somewhat of a breaking change, in a sense that code that got previously formatted without crashing will now cause an error to be thrown. But that'll only effect you when you used a `langauge` option that didn't support curly-bracws/square-brackets to parse code which contained them.


inferrinizzard (approved):
yes, I remember leaving them in as a failsafe incase the preceding token type didn't match but removing for parser's sake also works

PR #404, opened 2022-08-23
Remove unnecessary type imports

nene:
I suspect these type-only imports were added to deal with cyclic dependencies for which ESLint `no-cycle` rule will throw an error. I think the `no-cycle` rule is worth having as cyclic dependencies can indeed cause problems at runtime. They won't be problem with types, as these will get compiled away, so the solution to use `import type` is IMHO a valid one.

This leaves only one such import, for which I also added a comment to explain why we're using it.

Another reason to `import type` would be if we wanted to only import e.g. a type of a class, but for some reason not the class implementation itself. I don't think we have any such uses.

This PR also eliminates all uses of `import { type Foo }` syntax, which makes the library compatible with older versions of TypeScript, which I think is nice, as several people have bumped into this.

### DIFF
--- a/vscode/src/extension.ts
+++ b/vscode/src/extension.ts
@@ -1,6 +1,6 @@
 import * as vscode from 'vscode';
-import { format } from 'sql-formatter';
 import {
+  format,
   SqlLanguage,
   KeywordCase,
   IndentStyle,
--- a/vscode/src/extension.ts
+++ b/vscode/src/extension.ts
@@ -1,6 +1,6 @@
 import * as vscode from 'vscode';
 import { format } from 'sql-formatter';
-import type {
+import {
   SqlLanguage,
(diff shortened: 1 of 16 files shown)


vscode/src/extension.ts
```diff
@@ -1,6 +1,6 @@
 import * as vscode from 'vscode';
-import { format } from 'sql-formatter';
-import type {
+import {
+  format,
```
nene, line 3: Initially made this change in vscode extension unintentionally with global search/replace.

It looked fine though, so decided to keep it. But do check this one more carefully.

inferrinizzard (approved):

inferrinizzard:
I personally prefer having the type and value imports separated but I think it's fine without the type imports as well

PR #405, opened 2022-08-24
Replace open- and closeParens configs with extraParens

nene:
This eliminates the possibility of making several bugs:

- not including '(' and ')', which are always mandatory.
- only specifying one type of the pair, e.g. '[' without ']'.

Plus it also makes the config shorter.

Also eliminated Trino `{-` and `-}` paren types support for now. The previous implementation didn't work anyway as the `{` was always matched before `{-`, so it ended up being tokenized as `{` and `-`, which ended up being formatted correctly, as the contents between `{` and `}` is formatted without spaces surrounding it. The contents of these `{- ... -}` expressions really need to be parsed as a different thing altogether. I'm thinking it might be better to treat them as strings with `{-` and `-}`  delimiters - that would eliminate lots of tricky corner cases with formatting of these and also we would avoid spending too much effort on a Trino-specific syntax that's not supported anywhere else.


src/lexer/regexFactory.ts
```diff
@@ -17,11 +11,13 @@ export const lineComment = (lineCommentTypes: string[]) =>
   new RegExp(`(?:${lineCommentTypes.map(escapeRegExp).join('|')}).*?(?=\r\n|\r|\n|$)`, 'uy');
 
 /**
- * Builds a RegExp for matching parenthesis patterns, escaping them with `escapeParen`
- * @param {string[]} parens - list of strings that denote parenthesis patterns
+ * Builds a RegExp for matching either open- or close-parenthesis patterns
  */
-export const parenthesis = (parens: string[]): RegExp =>
-  patternToRegex(parens.map(escapeParen).join('|'));
+export const parenthesis = (kind: 'open' | 'close', extraParens: ('[]' | '{}')[] = []): RegExp => {
```
inferrinizzard, line 16: can we move this `'[]' | '{}'` to a ParenType `type` in regexTypes ?

inferrinizzard (approved):

src/lexer/regexFactory.ts
nene: Opted to skip this for now. It's just two places where this type is used and it's a very simple one.

PR #407, opened 2022-08-25
Add missing string tests

nene:
Just cover the already existing string types with tests.


nene:
Also discovered that some strings had invalid prefixes. Didn't add tests for these, instead did another PR: https://github.com/sql-formatter-org/sql-formatter/pull/408

inferrinizzard (approved):
I think this is fine for now but it would be nice to eventually have a procedural test suite for strings and string types / prefixes

PR #409, opened 2022-08-26
Eliminate skipped Redshift tests

nene:
In an effort to eliminate skipped tests altogether...

- The formatting of the first one is IMHO mostly OK as it is now.
- The second one could surely be better, but I think it's better to keep the active test v/s keeping it in skipped state.

Neither of these tests is for something crucial for which we have concrete plans of improvement.

### DIFF
--- a/test/redshift.test.ts
+++ b/test/redshift.test.ts
@@ -86,7 +86,7 @@ describe('RedshiftFormatter', () => {
     );
   });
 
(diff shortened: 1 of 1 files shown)


inferrinizzard (approved):

inferrinizzard:
Hmm, I think this would eventually need special parser support but we'll cross that bridge once we get there

PR #414, opened 2022-08-30
Fix tabular style formatting of LIMIT clause

### LINKED ISSUE #412: [FORMATTING] LIMIT format error
**Input data**
select * from account limit 10;
Indentation style: Tabular, Left
**Expected Output**

```sql
select    *
from      account
limit       10;
```

**Actual Output**

```sql
select    *
from      account
limit
          10;
```

**Usage**

- How are you calling / using the library?  I using https://sql-formatter-org.github.io/sql-formatter/
- What SQL language(s) does this apply to? SQL
- Which SQL Formatter version are you using? 10.0.0

nene:
Fixes #412


nene:
I merge this in to get the fix released. It's a pretty trivial fix.

PR #416, opened 2022-08-30
Reduce coupling between Tokenizer and TokenizerEngine

nene:
`TokenizerEngine` now no more has knowledge of individual token types, eliminating the need to update both `Tokenizer` and `TokenizerEngine` each time we need a new token type.

`TokenizerEngine` now takes and array of `TokenRule` objects as input configuration, the order of which also defines the order of tokenization. (Instead of taking a map of `TokenType-TokenRule` pairs and looking up rules by type.)

To make this change possible, I also eliminated all token-type specific logic from the engine:

- The `key` field transform can now be applied to any token type, not only to `*_PARAMETER` tokens (though we still use it only for the latter).
- The no-keyword-can-come-after-dot logic has been moved into `disambiguateTokens()` function, which already deals with several similar sorts of transformations of the tokens list.


nene:
Merging this in to build other things on top of it.

PR #417, opened 2022-08-30
Fix crash caused by ROWS BETWEEN

### LINKED ISSUE #415: Regression with the token PRECEDING with version 10 for bigquery formatting
**Describe the bug**
There seems to be a regression in what is supported by sql-formatter for the bigquery SQL language.

**Expected behavior**
I expected the version 10 to still support the token PRECEDING.

**Actual behavior**

I got the following error : 
187225626-99e10276-3d9e-49a1-914e-4015bfc19f6d

**Usage**

```
format(sql, {
    language: 'bigquery',
});
```
 bigquery
10

nene:
A quick fix for BigQuery. There's most likely the same problem with many other dialects. Will investigate further and apply the fix to other dialects as well. 

Fixes #415

### DIFF
--- a/src/languages/bigquery/bigquery.formatter.ts
+++ b/src/languages/bigquery/bigquery.formatter.ts
@@ -127,14 +127,16 @@ const reservedJoins = expandPhrases([
   '{INNER | CROSS} JOIN',
 ]);
 
-const reservedPhrases = [
+const reservedPhrases = expandPhrases([
   // https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#tablesample_operator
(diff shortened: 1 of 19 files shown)


nene:
Updated [wiki](https://github.com/sql-formatter-org/sql-formatter/wiki/WINDOW-clause) in regards to window functions.

Confirmed that window functions are supported in all dialects. Applied the fix to all dialects and made the formatting of window specifications also work everywhere.

nene:
Merging this in as it's a bug that effects window-functions formatting in all dialects and I'd like to get a bugfix release out.

PR #422, opened 2022-09-01
Add support for nested block comments

### LINKED ISSUE #411: [FORMATTING] Inline block comment has error
I using Postgresql support block comment.
**Input data**
/* 
/* inline block comment */
*/

**Expected Output**

```sql
/* 
/* inline block comment */
*/
```

**Actual Output**

```sql
/* 
/* inline block comment */
* /
```

It has one space character in * and /

**Usage**

- How are you calling / using the library? YES
- What SQL language(s) does this apply to? SQL
- Which SQL Formatter version are you using? 10.0.0

nene:
PostgreSQL and Transact-SQL now allow for nested block comments.

To implement parsing of such comments (which can't be done with a regular expression) I created NestedComment class that implements a RegExp like interface, but internally contains a small tokenizer/parser logic for counting the proper nesting of the inner comments.

**Note:** Also dropped the old behavior of treating and unclosed `/* ...` as an actual comment. We've been gradually moving away from supporting this sort of incomplete syntax. We used to support unbalanced parenthesis, which we don't do any more (after the switch to Nearley parser). We're still supporting unclosed string literals, which I'm planning to drop as well. This all is a left-over from times when the formatter had very rudimentary understanding of SQL syntax and the goal was to avoid breaking when encountering syntax we did not expect. The goal was never to actually support formatting of broken SQL. So I think it's time for this to go.

Fixes #411

### DIFF
--- a/src/languages/postgresql/postgresql.formatter.ts
+++ b/src/languages/postgresql/postgresql.formatter.ts
@@ -332,6 +332,7 @@ export default class PostgreSqlFormatter extends Formatter {
       reservedPhrases,
       reservedKeywords: keywords,
       reservedFunctionNames: functions,
+      nestedBlockComments: true,
       extraParens: ['[]'],
       stringTypes: ['$$', { quote: "''", prefixes: ['B', 'E', 'X', 'U&'] }],
       identTypes: [{ quote: '""', prefixes: ['U&'] }],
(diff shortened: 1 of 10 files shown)


inferrinizzard:
As an aside, this reminded me of this blog post by one of the leading authorities of Regex in JS about recursive matching: https://blog.stevenlevithan.com/archives/javascript-match-nested

inferrinizzard (approved):

PR #424, opened 2022-09-02
Add PropertyAccessNode

### LINKED ISSUE #423: Using aliases in Postgres BETWEEN
Hi there, sql-formatter 10.0.0 introduced a bug when using aliased columns with BETWEEN

This fails:
```sql
SELECT
  1 BETWEEN t.low AND t.high AS test
FROM
  test AS t
```

This passes:
```sql
SELECT
  1 BETWEEN low AND t.high AS test
FROM
  test AS t
```

Here is the error;
```
Parse error at token: .
Unexpected OPERATOR token: {"type":"OPERATOR","raw":".","text":".","start":29,"end":30}. Instead, I was expecting to see one of the following:

A AND token based on:
    between_predicate → %BETWEEN commaless_expression ● %AND commaless_expression
    simple_expression$subexpression$1 →  ● between_predicate
    simple_expression →  ● simple_expression$subexpression$1
    expression$subexpression$1 →  ● simple_expression
    expression →  ● expression$subexpression$1
    select_clause$subexpression$1$ebnf$2 → select_clause$subexpression$1$ebnf$2 ● expression
    select_clause$subexpression$1 → asteriskless_expression ● select_clause$subexpression$1$ebnf$2
    select_clause → %RESERVED_SELECT ● select_clause$subexpression$1
    clause$subexpression$1 →  ● select_clause
    clause →  ● clause$subexpression$1
    expressions_or_clauses$ebnf$2 → expressions_or_clauses$ebnf$2 ● clause
    expressions_or_clauses → expressions_or_clauses$ebnf$1 ● expressions_or_clauses$ebnf$2
    statement →  ● expressions_or_clauses statement$subexpression$1
    main$ebnf$1 → main$ebnf$1 ● statement
    main →  ● main$ebnf$1
```

Calling using jest
POSTGRESQL
10.0.0

nene:
Replaced plain `"."` operator with `PropertyAccessNode`.

This solves the immediate problem of parsing `BETWEEN t.foo AND t.bar`, as `t.bar` is now considered a single expression (v/s previously being considered 3 separate expressions, which the BETWEEN-predicate didn't allow for).

It is a step in the right direction, but here the limits of our current approach are really starting to show. Like when one would write `BETWEEN 5+5 AND 20` this would still fail to parse, although `BETWEEN (5+5) AND 20` parses just fine.

Fixes #423

### DIFF
--- a/src/lexer/Tokenizer.ts
+++ b/src/lexer/Tokenizer.ts
@@ -152,7 +152,7 @@ export default class Tokenizer {
       },
       {
         type: TokenType.OPERATOR,
-        regex: regex.operator('+-/%&|^><=.:$@#?~!', [
+        regex: regex.operator('-+/%&|^><=:$@#?~!', [
           '<>',
           '<=',
           '>=',
@@ -161,6 +161,7 @@ export default class Tokenizer {
         ]),
       },
       { type: TokenType.ASTERISK, regex: /[*]/uy },
+      { type: TokenType.DOT, regex: /[.]/uy },
     ]);
   }
 
--- a/src/parser/grammar.ne
+++ b/src/parser/grammar.ne
@@ -106,11 +106,11 @@ set_operation -> %RESERVED_SET_OPERATION expression:* {%
   })
 %}
 
-expression -> ( simple_expression | asterisk | comma ) {% unwrap %}
+expression -> ( simple_expression | between_predicate | asterisk | comma ) {% unwrap %}
 
-asteriskless_expression -> ( simple_expression | comma ) {% unwrap %}
+asteriskless_expression -> ( simple_expression | between_predicate | comma ) {% unwrap %}
 
-commaless_expression -> ( simple_expression | asterisk ) {% unwrap %}
+commaless_expression -> ( simple_expression | between_predicate | asterisk ) {% unwrap %}
 
 simple_expression ->
   ( array_subscript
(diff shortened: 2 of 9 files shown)


src/lexer/Tokenizer.ts
```diff
@@ -152,7 +152,7 @@ export default class Tokenizer {
       },
       {
         type: TokenType.OPERATOR,
-        regex: regex.operator('+-/%&|^><=.:$@#?~!', [
+        regex: regex.operator('-+/%&|^><=:$@#?~!', [
```
nene, line 155: Discovered a grave error in this code.

Previously this string started with `+-/`, which ended up being placed inside regex character class, like `[+-/]`. Which instead of matching `+`, `-` and `/`, ended up matching all characters between `+` and `/` (which by chance happens to include `+`, `,`, `-`, `.`, `/`).

src/parser/grammar.ne
```diff
@@ -106,19 +106,19 @@ set_operation -> %RESERVED_SET_OPERATION expression:* {%
   })
 %}
 
-expression -> ( simple_expression | asterisk | comma ) {% unwrap %}
+expression -> ( simple_expression | between_predicate | asterisk | comma ) {% unwrap %}
```
inferrinizzard, line 109: this seems a little awkward for now but I suppose it'll look better as we expand the grammar more

inferrinizzard (approved):
good catch on the operator regex (which will be updated soon with the other PR)

src/parser/grammar.ne
nene: Yeah. This part is not the prettiest for now :P

PR #425, opened 2022-09-03
Overhaul of string/identifier/variable escaping

### LINKED ISSUE #308: More than one string with trailing backslash

nene:
instead of quote types `""` and `''`, there are now four versions of each to pick the escaping style:

- `""-qq` - with repeated-quote escape
- `""-bs` - with backslash escape
- `""-qq-bs` - with both repeated-quote and backslash escape
- `""-raw` - with out any escaping

The old plain `""` and `''` have been dropped, to force one to explicitly make a choice of escaping style.

Added bunch of additional tests to check for escaping (and also to check for particular styles of escaping not working).

Also dropped support for incomplete strings (like `"...`), just like I previously dropped support for incomplete block-comments.

Fixes #308


src/lexer/regexFactory.ts
```diff
@@ -96,25 +96,35 @@ const buildQStringPatterns = () => {
...
+  // PostgreSQL dollar-quoted
+  '$$': String.raw`(?<tag>\$\w*\$)[\s\S]*?\k<tag>`,
+  // BigQuery '''triple-quoted''' (using \' to escape)
+  "'''..'''": String.raw`'''[^\\]*?(?:\\.[^\\]*?)*?'''`,
+  // BigQuery """triple-quoted""" (using \" to escape)
+  '""".."""': String.raw`"""[^\\]*?(?:\\.[^\\]*?)*?"""`,
+  // Hive and Spark variables: ${name}
+  '{}': String.raw`(?:\{[^\}]*\})`,
+  // Oracle q'' strings: q'<text>' q'|text|' ...
   "q''": buildQStringPatterns(),
```
nene, line 128: The main changes here involve `""` and `''`.

The only change done with all other quote styles is that support for missing closing quote has been dropped.

test/features/identifiers.ts
```diff
@@ -71,15 +80,22 @@ export default function supportsIdentifiers(format: FormatFn, identifierTypes: I
       `);
     });
 
+    it('supports escaping in U&"" strings by repeated quote', () => {
+      expect(format('U&"foo "" JOIN bar"')).toBe('U&"foo "" JOIN bar"');
+    });
+
     it('detects consecuitive U&"" identifiers as separate ones', () => {
```
inferrinizzard, line 87:
```suggestion
    it('detects consecutive U&"" identifiers as separate ones', () => {
```

inferrinizzard:
would this -qq and -bs suffix also work for some of the prefixed quote types ?
I know a few of the dialect have similar prefixed- types but may have different escape methods between them

inferrinizzard (approved):

nene:
> would this -qq and -bs suffix also work for some of the prefixed quote types ?

Yep. I'm already using it for various prefixed strings. For example the `-raw` suffix with `R'...'` strings.

PR #426, opened 2022-09-04
Overhaul handling of single-char operators

nene:
Instead of supporting all possible single-char operators in all dialects we're now only supporting them in the dialects that support them.

Also separated operators lists in tests from operators lists in .formatter.ts files. This adds some duplication, but makes the tests ultimately more useful.


src/languages/postgresql/postgresql.formatter.ts
```diff
@@ -311,11 +292,40 @@ const binaryOperators = [
...
+  // Similarity (TODO: not in wiki)
+  '<%',
+  '<<%',
+  '%>',
+  '%>>',
+  // Byte comparison (TODO: not in wiki)
+  '~>~',
+  '~<~',
+  '~>=~',
+  '~<=~',
```
nene, line 321: These were added in #296, but I'm unable to locate documentation for these.

@inferrinizzard do you perhaps know where I could locate these in PostgreSQL docs?

inferrinizzard: the `~_~` operators are not in the docs, found from this SO link: https://stackoverflow.com/a/35808554/10028067
if you run `\do ~>~` or similar in pg cli, you get these results:
image

inferrinizzard: the `%` ish operators are in table F.26 here: https://www.postgresql.org/docs/current/pgtrgm.html#id-1.11.7.42.6

inferrinizzard: it seems there are more operators I missed:
```
<->
<<->
<->>
<<<->
<->>>
```

src/languages/postgresql/postgresql.formatter.ts
```diff
@@ -337,7 +262,88 @@ export default class PostgreSqlFormatter extends Formatter {
...
+        '~<~',
+        '~>=~',
+        '~<=~',
+        // String concatenation
+        '||',
+        // Text search
+        '@@@',
+        '!!',
+        // Type cast
+        '::',
```
inferrinizzard, line 345:
```suggestion
        '::',
        // String distance operators
        '<->>',
        '<<->',
        '<->>>',
        '<<<->'
```

inferrinizzard (approved):

PR #428, opened 2022-09-05
Upgrade SQL Formatter to 10.4.0 in VSCode extension

nene:


inferrinizzard:
I'll update this with all the new configs after #429 is merged

PR #429, opened 2022-09-05
Rewrite inline block formatting logic

nene:
Instead of having an InlineBlock class, which deals with inline
block length calculation, the inline formatting is now handled
by subclass of Layout.

The old InlineBlock class was very error-prone, needing an update
each time a new AST node type was added, resulting in lots of bugs
over time.  It also duplicated the logic in ExpressionFormatter,
with the length calculation having to be constantly synced with
between the two.  On top of it, it also had bugs, like not taking
account denseMode and parameter value expansions.

Now the InlineLayout simply monitors all the chunks that get
passed through it:

- when encountering a newline an error is thrown,
- when things add up to too long of a string, an error is thrown,

...and we switch to formatting in non-inline mode.


nene:
I'll merge this in to release it and have it tested in real world. I don't really expect any problems though (besides some old problems going away). If you do have some comments, feel free to still add them.

PR #431, opened 2022-09-05
Update VSCode Extension for v10.x

inferrinizzard:
- bump sql-formatter to v10.x
- updated TSQL with TransactSQL
- added SingleStoreDB
- updated CONTRIBUTING with vscode instructions
- upgrade vsce


vscode/package.json
```diff
@@ -123,10 +125,11 @@
             "plsql",
             "postgresql",
             "redshift",
+            "singlestoredb",
             "spark",
             "sqlite",
             "trino",
-            "tsql"
+            "transactsql"
```
nene, line 132: I'd suggest not removing tsql, but keeping both.

When you remove it, it's likely the config will break for anyone who uses this value. Though I don't really know how VSCode will behave in this case. Will switch to the default value? Will it keep the old value regardless it being invalid according to this config? Something else?

nene (changes requested):
I released sql-formatter 10.5.0. I suggest you upgrade to that version before releasing as otherwise the new `language: "transactsql"` value won't work.

vscode/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # CHANGELOG
 
+## [1.3.0] - 2022-09-05
+
+- updated `sql-formatter` to v10.x
+- updated to Nearley parser
```
nene, line 6: IMHO there's no point in listing internal sql-formatter changes which don't effect VSCode extension in any way. Like the switch to Nearly parser required no changes in the VSCode extension.

It would be more useful though to mention the exact version name of sql-formatter.

inferrinizzard: Added this as a note for potential performance impact

nene (approved):

PR #436, opened 2022-09-06
Add location data to AST

nene:
- Grouped `start`/`end` fields into single `loc: {start, end}` object. This makes it simple to just copy the whole location data from a token and place it into AST Node.
- Added `loc` fields to all AstNode objects, computing locations based on tokens and/or child nodes.
- Changed the appended EOF token to use actual source code length as location to avoid `Infinity` locations.

The locations of StatementNode objects are still problematic. Currently, a statement starts at first visible token and ends at semicolon, which leaves gaps in the location data:

```
USE db1;         SELECT * FROM tbl;
^------^         ^----------------^
```

Would be better to measure a statement from the end of previous one:

```
USE db1;         SELECT * FROM tbl;
^------^^-------------------------^
```


nene:
This whole thing is to explore associating comments with AST nodes. Not really sure of this approach. Will try tackling the problem also from a different angle.

nene:
This approach turned out to be a dead end. Or at least I didn't manage to figure out how to make it work. I'm probably gonna close down this branch.

Though I'm thinking of preserving some changes, like the grouping of location data to single field: aad8dd9b74a8b946f5da60cce662f162749a41a3

PR #439, opened 2022-09-08
Attach comments to AST nodes

nene:
Added two optional fields to each node:

- `leadingComments`
- `trailingComments`

Chose these names after babel-eslint9 parser (see in [astexplorer page](https://astexplorer.net/)), which also adds comment data to its AST in similar fashion.

Instead of throwing removing comment tokens from token stream, and then later attempting to re-associate them with AST nodes based on the locations (which was my original idea that I tried to explore in #436) the comments are now intentionally picked up during parsing. Commonly with a pattern like:

```
between_predicate -> %BETWEEN _ simple_expression _ %AND _ simple_expression
```

Where `_` stands for optional comments which, when present, get added to `leading/trailingComments` field of nodes.

This approach seems quite promising. At least I got it working for several node types. Unlike with the locations approach where I realized it's either going to be really hard or maybe not even possible at all. Plus, I've already managed to add support for comments in places where previously a comment would have caused a crash.

For now there are still "free floating" comments which aren't inside `leading/trailingComments` fields. These should go away eventually as the AST evolves into a more correct form. For now they're still there and I'm not planning to eliminate them within this PR.

The formatting of many of these in-between comments is also not quite good. Again, I'm not gonna tackle that problem within this PR. These badly-formatted comments are still way better than comments that crash the parser. And these comments are mostly pretty odd cases like `table /*comment*/ . /*comment*/ column` which should occur pretty rarely.

### DIFF
--- a/src/lexer/disambiguateTokens.ts
+++ b/src/lexer/disambiguateTokens.ts
@@ -33,7 +33,7 @@ const dotKeywordToIdent = (token: Token, i: number, tokens: Token[]): Token => {
 
 const funcNameToKeyword = (token: Token, i: number, tokens: Token[]): Token => {
   if (token.type === TokenType.RESERVED_FUNCTION_NAME) {
-    const nextToken = tokens[i + 1];
+    const nextToken = nextNonCommentToken(tokens, i);
     if (!nextToken || !isOpenParen(nextToken)) {
       return { ...token, type: TokenType.RESERVED_KEYWORD };
     }
--- a/src/lexer/disambiguateTokens.ts
+++ b/src/lexer/disambiguateTokens.ts
@@ -23,7 +23,7 @@ export function disambiguateTokens(tokens: Token[]): Token[] {
 
 const dotKeywordToIdent = (token: Token, i: number, tokens: Token[]): Token => {
   if (isReserved(token.type)) {
-    const prevToken = tokens[i - 1];
+    const prevToken = prevNonCommentToken(tokens, i);
     if (prevToken && prevToken.text === '.') {
       return { ...token, type: TokenType.IDENTIFIER, text: token.raw };
     }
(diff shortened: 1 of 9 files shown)


src/lexer/disambiguateTokens.ts
```diff
@@ -53,14 +53,32 @@ const identToArrayIdent = (token: Token, i: number, tokens: Token[]): Token => {
...
 
+const prevNonCommentToken = (tokens: Token[], index: number): Token | undefined =>
+  nextNonCommentToken(tokens, index, -1);
+
+const nextNonCommentToken = (
+  tokens: Token[],
+  index: number,
+  dir: -1 | 1 = 1
+): Token | undefined => {
+  let i = 1;
```
inferrinizzard, line 72: can be `tokens.slice(index).find(token => !isComment(token))`

src/parser/ast.ts
```diff
@@ -21,46 +21,51 @@ export enum NodeType {
   block_comment = 'block_comment',
 }
 
-export type StatementNode = {
+type BaseNode = {
+  leadingComments?: CommentNode[];
+  trailingComments?: CommentNode[];
+};
+
+export type StatementNode = BaseNode & {
```
inferrinizzard, line 29: these should be `export interface XNode extends BaseNode`

inferrinizzard (changes requested):

src/lexer/disambiguateTokens.ts
nene: `slice()` has linear complexity. The function itself is called in a loop over all tokes, which would lead to square complexity in the worst case. I'd rather avoid that.

Though it might be that JS engine optimizes away the overhead of creating a slice. Would need to be measured to be sure.

src/parser/ast.ts
nene: I've actually been pondering over switching away from using `interface` completely. Everything done with `interface` can also be done with `type`, with one exception:

- declaring two interfaces with same name will result in them being merged together into a single interface. A behavior one almost never wants.

On the other side, there are lots of things one can do with `type` which can't be done with `interface`.

So there's two options:

- use both type and interface (with rules to decide when to use one or the other)
- use only type (with no decisions required over which syntax to use in which case).

inferrinizzard: I've always followed the convention of:
`interface` for all object-like declarations, `type` for everything else (ie. unions, aliases, etc.)

src/lexer/disambiguateTokens.ts
nene: Ran the performance comparison which showed there being no real performance difference.

However I'll stick to the original implementation as I need the function to work for searching both forward and backward through token list and the proposed simplification only works for forward-search.

src/parser/ast.ts
nene: I'll switch to interface for now.

PR #440, opened 2022-09-08
Group token location data to loc field

nene:
Cherry-picked changes from #436:

- Added basic unit test for Tokenizer. 
- Grouped `start`/`end` fields into single `loc: {start, end}` object. This simplifies processing the location info separately from the whole token. Not used right now, but planning to use in the future.
- Changed the appended EOF token to use actual source code length as location to avoid `Infinity` locations.

### DIFF
--- a/src/languages/bigquery/bigquery.formatter.ts
+++ b/src/languages/bigquery/bigquery.formatter.ts
@@ -204,8 +204,10 @@ function combineParameterizedTypes(tokens: Token[]) {
         type: TokenType.IDENTIFIER,
         raw: typeDefTokens.map(formatTypeDefToken('raw')).join(''),
         text: typeDefTokens.map(formatTypeDefToken('text')).join(''),
-        start: token.start,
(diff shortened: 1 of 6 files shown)


inferrinizzard:
should use `location` or `pos` instead as key, otherwise, lgtm

nene:
I chose the name `loc` based on the babel-eslint9 parser after which I also modeled the `leadingComments`/`trailingComments` fields.

I've been thinking though whether the end location will be needed at all. Currently only the start location is ever used. So we could eliminate the whole group and just have a single numeric field.

inferrinizzard:
I think removing end location should be fine and if ever we need it, we can just restore it from this branch
still not fully on-board with `loc` as a shorthand, definitely not obvious that it stands for location

nene:
I'll merge this in as-is for now, as another PR already depends on changes in here. Then I'll add another PR to change the loc field.

PR #442, opened 2022-09-09
Include line & column number to Tokenizer and Parser errors

nene:
Created `lineColFromIndex()` helper, which computes line & column number from character index in source code.

### DIFF
--- a/src/lexer/lineColFromIndex.ts
+++ b/src/lexer/lineColFromIndex.ts
@@ -1,0 +1,9 @@
+/**
+ * Determines line and column number of character index in source code.
+ */
+export function lineColFromIndex(source: string, index: number): LineCol {
+  const lines = source.slice(0, index).split(/\n/);
+  return { line: lines.length, col: lines[lines.length - 1].length + 1 };
+}
+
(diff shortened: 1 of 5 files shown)


inferrinizzard (approved):

src/lexer/lineColFromIndex.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Determines line and column number of character index in source code.
+ */
+export function lineColFromIndex(source: string, index: number): LineCol {
+  const lines = source.slice(0, index).split(/\n/);
+  return { line: lines.length, col: lines[lines.length - 1].length + 1 };
+}
+
+export type LineCol = { line: number; col: number };
```
inferrinizzard, line 9: should be interface

src/lexer/lineColFromIndex.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Determines line and column number of character index in source code.
+ */
+export function lineColFromIndex(source: string, index: number): LineCol {
```
inferrinizzard, line 4: can be moved to utils.ts

PR #449, opened 2022-09-20
Store script for performance testing in repo

nene:
I've been using this script for performance testing. Opted to store it in repository, so I don't have to go and search for it from old PR comments whenever I have a need for it again.


inferrinizzard (approved):

inferrinizzard:
you should also be able to store this as a .sql file and import into a jest test, if you like

PR #451, opened 2022-09-20
Add support for Redshift type-cast operator

### LINKED ISSUE #450: [FORMATTING] Redshift dialect fails when using :: to cast after v10.2.0
**Input data**

Casting to any type using the `::` format. The simplest example is:

```sql
SELECT 1::float;
```

**Expected Output**

```sql
SELECT 1:: float; -- or similar
```

**Actual Output**

On every version of the parser over 10.2.0 the the parser throws a JS error.

```
Parse error: Unexpected "::float"
```

**Usage**

- How are you calling / using the library? Node 16
- What SQL language(s) does this apply to? Redshift (others untested)
- Which SQL Formatter version are you using? 10.3.0+

nene:
Fixes #450

### DIFF
--- a/src/languages/redshift/redshift.formatter.ts
+++ b/src/languages/redshift/redshift.formatter.ts
@@ -167,6 +167,7 @@ export default class RedshiftFormatter extends Formatter {
         '<<',
         '>>',
         '||',
+        '::',
       ],
     });
   }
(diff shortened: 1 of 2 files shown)


nene:
A trivial fix. Will merge this straight in to do a bugfix release.

Issue #366, opened 2022-08-01
Feature Request: Add Snowflake as a SQL Dialect


gvanstee:
**Describe the Feature**
The `sql-formatter` currently supports many SQL dialects, including BigQuery and SQL Server. Would it be possible to add Snowflake?

**Why do you want this feature?**
I use Snowflake. I am particularly interested in whether the auto-formatter provided from the open source code by Snowflake, [GitHub - Snowflake-Labs/snowsql-formatter](https://github.com/Snowflake-Labs/snowsql-formatter), is materially different (in whether it will produce compilation errors) from what is provided here by `sql-formatter`.

inferrinizzard:
Duplicate of #173 
I'll make note of the consistent interest of adding Snowflake as a supported dialect.

As for the differences between this upstream library and `snowsql-formatter`, this package has had several major revisions in across all parts of the codebase in the past 6-12months so I would definitely recommend this package if it works for use cases.

We would gladly accept a PR for Snowflake as a new language if you feel up to it, please feel free to reach out to me or @nene for assistance - the main effort there would be to research/document the syntax (ie. special keywords, supported operators, string types, etc.)

primeapple:
I'm also interested in this and will have a look!

inferrinizzard:
Great, #297 is a good example of a PR with a recently added language (Trino) so you can reference that to see what changes are needed

primeapple:
Question:
For some Languages you have split the functions in categories: https://github.com/sql-formatter-org/sql-formatter/blob/master/src/languages/plsql/plsql.functions.ts#L5 (`numeric`, `character`, .etc.)

For others there is only a single category: https://github.com/sql-formatter-org/sql-formatter/blob/1d83dc245c3257ec61f4b15a63b73b1b32e63482/src/languages/trino/trino.functions.ts#L10 (`all`)

What is preferred here?  
Snowflake offers categories for functions as well, however some functions belong to multiple categories. (see https://docs.snowflake.com/en/sql-reference/functions-all.html , have a look at the `Category` column). What should be done in these cases if wemove with seperate categories?

nene:
The main reason for grouping or not grouping is to make it easier to compare with language documentation. Like when official docs of an SQL dialect have functions listed in groups, it would also be better to list them in groups in the functions list. Duplicates don't matter, the `flatKeywordList()` function eliminates duplicates.

nene:
Snowflake support is now present since version 11.0.0

PR #454, opened 2022-09-22
Add support for Snowflake Language

primeapple:
This is first draft to solve #366 . It does not work yet, mainly because I had some troubles with the `QuoteTypes`.

## Review
Please review, have a look at the following things:

1. `TODO` comments show lines where I'm totally lost
2. I'm very unsure about the `reservedComments`, the `functions` and the `keywords`.
3.  Did I put the correct things in there, or do they belong somewhere else?
4. Please check the places where I used the `[thing] {a|b|c} ...` syntax, if I used them correctly.
5. Did I miss any tests?
6. Any other places that have to be changed?

## Questions
I want to support accessing elements in unstructured data.
This can be done via square brackets `[,]` or via colons `:` , like [here](https://docs.snowflake.com/en/user-guide/querying-semistructured.html#bracket-notation):
```sql
select src:salesperson.name from car_sales

select src['salesperson']['name'] from car_sales
```
How should I do this? I added `:` as an operator, is this right? How about the brackets?

Thanks for this Repo!


nene:
Thanks for the effort.

#### Regarding your main question about colon and square-bracket syntax.

- For brackets, adding `extraParens: ['[]']` should do the trick. See for example in `n1ql.formatter.ts`
- With colon, things are a bit trickier. You should indeed add it as an operator. However, currently the default is to format colon with a space after (as in `name: value`). To fix this, a language-specific condition should be added to `ExpressionFormatter.formatOperator()` - there already is a similar thing for `@` operator in PL/SQL.

#### In general

It would be really helpful if you could also update our [wiki](https://github.com/sql-formatter-org/sql-formatter/wiki) with information about Snowflake SQL syntax. Otherwise I'll have to do all this research on my own before I can merge this PR.

primeapple:
I will do all the stuff you mentioned.
Please go other the `TODO` comments in the changes. I can't run tests because I don't understand the Quote types yet...

src/languages/snowflake/snowflake.formatter.ts
```diff
@@ -0,0 +1,282 @@
...
+//
+// 1. run in console on this page: $x('//tbody/tr/*[1]//a/span/text()').map(x => x.nodeValue)
+// 2. delete all lines that contain a sting like '(.*)', they are already covered in the list
+// 3. delete all lines that contain a sting like '<.*>', they are already covered in the list
+// 4. delete all lines that contain '…', they are part of a regex statement that can't be covered here
+// 5. Manually add 'COPY INTO'
+//
+// Steps 1-4 can be combined by the following script in the developer console:
+// $x('//tbody/tr/*[1]//a/span/text()').map(x => x.nodeValue) // Step 1
+//   filter(x => !x.match(/\(.*\)/) && !x.match(/…/) && !x.match(/<.*>/)) // Step 2-4
```
nene, line 19: This generated list is not really particularly useful. For example `INSERT` shouldn't be listed as a separate command, instead there should be `INSERT [OVERWRITE] INTO`.

I suggest you instead look at the `reservedCommands` in `sql.formatter.ts` and use it as a base and modify it to match Snowflake syntax. Also, it would be nice to keep things in the same order (e.g. starting with `WITH`, `FROM`, `WHERE`), so it's easier to compare with other other dialect definitions in *.formatter.ts files.

src/languages/snowflake/snowflake.formatter.ts
```diff
@@ -0,0 +1,282 @@
...
+      // https://docs.snowflake.com/en/sql-reference/functions/case.html
+      reservedDependentClauses: ['WHEN', 'ELSE'],
+      reservedPhrases,
+      reservedKeywords: keywords,
+      reservedFunctionNames: functions,
+      // TODO $$ is allowed, but also ' and ` how to achieve this?
+      // https://docs.snowflake.com/en/sql-reference/data-types-text.html#string-constants
+      // Backticks only allowed for inline Javascript:
+      // https://docs.snowflake.com/en/sql-reference/stored-procedures-javascript.html#line-continuation
+      stringTypes: ['$$', `'`, '`'],
```
nene, line 256: As I understand, backticks (`` ` ``) aren't actually part of Snowflake SQL syntax (the docs suggest their usage inside JavaScript when dealing with SQL strings).

nene:
Regarding quote types, see: https://github.com/sql-formatter-org/sql-formatter/blob/master/src/lexer/regexFactory.ts#L99

The `-qq` and `-bs` suffixes at the end of quote types define what sort of escaping (if any) is used inside the quotes.

nene:
Some comments about wiki updates:

- [Parameters page](https://github.com/sql-formatter-org/sql-formatter/wiki/parameters) says that Snowflake supports `$1` parameters and links to [select docs](https://docs.snowflake.com/en/sql-reference/sql/select.html#parameters). I'm pretty sure these are something else than parameters one would use in prepared statements. I think these things are more similar to variables - they also share a very similar syntax. It's probably better to reference this syntax from [Variables page](https://github.com/sql-formatter-org/sql-formatter/wiki/variables). Probably the simplest path for supporting this syntax is also to treat them the same as variables - at lest that would be my first gut-instinct-guess.

- [Comments page](https://github.com/sql-formatter-org/sql-formatter/wiki/comments) mentions support for `# comments` but says these aren't actually supported. Syntax highlighting really shouldn't be used as a guide for what is and isn't supported. So probably best to be removed. Too bad you haven't found any official documentation for the comment syntax. Found one [Knowledge Base page](https://community.snowflake.com/s/article/how-to-comment-or-uncomment-multiple-lines-in-the-sql-worksheet) that at least mentions the `//` syntax - we could link to that.

primeapple:
I updated all the wiki pages for snowflake (also updated your suggestions).
One problem is the [Window-Clause](https://github.com/sql-formatter-org/sql-formatter/wiki/WINDOW-clause). Snowflake seems to not have support for a `WINDOW` clause in it's [Query Syntax)(https://docs.snowflake.com/en/sql-reference/constructs.html). There is [QUALIFY](https://docs.snowflake.com/en/sql-reference/constructs/qualify.html) but I'm unsure if this is the same?  
However, there is support for [Window Functions](https://docs.snowflake.com/en/sql-reference/functions-analytic.html) but also here I'm unsure if it is the same as in the other languages.  
Could you have a look into this?

I will rewriting the code now.

nene:
Thanks. Updated the docs for Snowflake window functions syntax.

nene:
BTW. I would appreciate if you didn't do further force-pushes to this branch. They make it hard to tell what has changed.

src/languages/snowflake/snowflake.formatter.ts
primeapple: Ok, I did this like in the `postgresql.formatter.ts`. I added the basic commands with the `{..} [...]` syntax, as well as the generated List because it contains many many useful phrases.

primeapple:
First refactor done.
Please the `snowflake.test` shows an infinity loop. Could you have a look into it? It happens even with a single test case.

src/languages/snowflake/snowflake.functions.ts
```diff
@@ -0,0 +1,614 @@
...
+    'BOOLNOT',
+    'BOOLOR',
+    'BOOLOR_AGG',
+    'BOOLXOR',
+    'BOOLXOR_AGG',
+    'BUILD_SCOPED_FILE_URL',
+    'BUILD_STAGE_FILE_URL',
+    'CASE',
+    'CAST',
+    '::', // this is an infix operator
```
nene, line 108: operators shouldn't be in functions list

src/languages/snowflake/snowflake.functions.ts
```diff
@@ -0,0 +1,614 @@
...
+    'CHECK_XML',
+    'CHR',
+    'CHAR',
+    'COALESCE',
+    'COLLATE',
+    'COLLATION',
+    'COMPLETE_TASK_GRAPHS',
+    'COMPRESS',
+    'CONCAT',
+    '||', // this is an operator like "a" || "b"
```
nene, line 122: operators shouldn't be in functions list

src/languages/snowflake/snowflake.functions.ts
```diff
@@ -0,0 +1,614 @@
...
+    'FLOOR',
+    'GENERATE_COLUMN_DESCRIPTION',
+    'GENERATOR',
+    'GET',
+    'GET_ABSOLUTE_PATH',
+    'GET_DDL',
+    'GET_IGNORE_CASE',
+    'GET_OBJECT_REFERENCES',
+    'GET_PATH',
+    ':', // this is an infix operator AS WELL as a function
```
nene, line 201: operators shouldn't be in functions list

src/languages/snowflake/snowflake.functions.ts
```diff
@@ -0,0 +1,614 @@
...
+    'ILIKE',
+    'ILIKE ANY',
+    '[ NOT ] IN',
+    'INFER_SCHEMA',
+    'INITCAP',
+    'INSERT',
+    'INVOKER_ROLE',
+    'INVOKER_SHARE',
+    'IS [ NOT ] DISTINCT FROM',
+    'IS [ NOT ] NULL',
```
nene, line 235: This `[ ... ]` syntax won't work in here because this list of strings is not wrapped inside `expandPhrases()` call.

Besides, these are again operators, not function names. That it, you won't be calling them as `IS NOT NULL(value)`

src/languages/snowflake/snowflake.functions.ts
```diff
@@ -0,0 +1,614 @@
...
+    'SYSTEM$SHOW_OAUTH_CLIENT_SECRETS',
+    'SYSTEM$STREAM_GET_TABLE_TIMESTAMP',
+    'SYSTEM$STREAM_HAS_DATA',
+    'SYSTEM$TASK_DEPENDENTS_ENABLE',
+    'SYSTEM$TYPEOF',
+    'SYSTEM$USER_TASK_CANCEL_ONGOING_EXECUTIONS',
+    'SYSTEM$VERIFY_EXTERNAL_OAUTH_TOKEN',
+    'SYSTEM$WAIT',
+    'SYSTEM$WHITELIST',
+    'SYSTEM$WHITELIST_PRIVATELINK',
```
nene, line 518: Currently these won't work because the functions list gets concatenated to a regex and `$` will attempt to match end-of-line.

I should fix this by adding escaping of function names before constructing a regex.

nene (commented):
The reason for the infinite loop is `||` operator in functions list, which gets concatenated into a long regex containing all function names and ends up matching empty string.

nene (commented):
Some additional problems that popped up to me

nene:
Regarding the formatting of `:` operator. I made a change to make the customization of formatting simpler: #466

One should now be able to simply add this to SnowflakeFormatter class:

```ts
formatOptions() {
  return { alwaysDenseOperators: [':'] };
}
```

nene:
Also added escaping of keywords, so the `$` character in function names should no more pose a problem: #473

src/languages/snowflake/snowflake.formatter.ts
```diff
@@ -0,0 +1,349 @@
...
+      reservedKeywords: keywords,
+      reservedFunctionNames: functions,
+      // TODO: add escaping for single ''
+      stringTypes: ['$$', `''-qq`],
+      identTypes: ['""-qq'],
+      variableTypes: [
+        // for accessing columns at certain positons in the table
+        { regex: '$[1-9]d*' },
+        // identifier style syntax
+        { regex: '$([_a-zA-Z][_a-zA-Z0-9$]*)' },
```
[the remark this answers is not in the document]
primeapple: Changed `d*` to `\\d*` to support any number sequence

primeapple:
I'm sorry, I screwed up the commits by including your commits into this PR. This is why I usually force push, it allows me to cleanly integrate commits to master with `rebase` to the feature branch and then force push the feature branch.

primeapple:
Ok, I reverted your commits in this PR and instead merged master into this branch. Everything fine now. Puhhh...

The only new important commit for you is [8dc6f51](https://github.com/sql-formatter-org/sql-formatter/pull/454/commits/8dc6f518a2e0bf9d76458c6a24274795b14f5a3b)

test/snowflake.test.ts
```diff
@@ -0,0 +1,81 @@
...
+      SELECT
+        foo:bar
+      FROM
+        test
+    `);
+  });
+
+  it(`does not support ':' path-operator with spaces`, () => {
+    expect(format(`SELECT foo : bar FROM test`)).toThrowError('Parse error');
+  });
```
primeapple, line 73: These tests fail, unsure why.

src/languages/snowflake/snowflake.formatter.ts
```diff
@@ -0,0 +1,352 @@
...
+  'GROUP BY',
+  'HAVING',
+  'PARTITION BY',
+  'ORDER BY',
+  'LIMIT',
+  'OFFSET',
+  'FETCH [FIRST | NEXT]',
+  // Data manipulation
+  // - insert:
+  'INSERT [OVERWRITE] [ALL INTO | INTO | ALL | FIRST]',
```
primeapple, line 35: Are multiple following `[...]` blocks supported? I'm unsure because of the spaces in between, I dont want it to become something like `INSERT  INTO` (2 spaces in between).

primeapple:
1. Is there an option to define how identifiers are allowed to look like?
In Snowflake they can include any symbol when inside `""`, but without quoting they have a certain rule. There is the `identChars` Config option, but I'm unsure if this refers to quoted or unquoted identifiers.

2. Is there an option to test variables?

3. Did I miss any obvious `support...` tests?

src/languages/snowflake/snowflake.formatter.ts
```diff
@@ -0,0 +1,352 @@
...
+      reservedKeywords: keywords,
+      reservedFunctionNames: functions,
+      stringTypes: ['$$', `''-qq-bs`],
+      identTypes: ['""-qq'],
+      variableTypes: [
+        // for accessing columns at certain positons in the table
+        { regex: '[$][1-9]\\d*' },
+        // identifier style syntax
+        { regex: '[$][_a-zA-Z][_a-zA-Z0-9$]*' },
+      ],
```
primeapple, line 330: Are these correctly written?

1. Should be anything like `$number`, where `number > 0`
2. Shoud be anything like `$identifier`, where identifier starts with char `a-Z`, can contain any combination of multiple `$`, `a-Z` and `0-9`

test/snowflake.test.ts
nene: Ah yeah, I forgot that there's a special formatting rule about the `:` operator and so the `alwaysDenseOperators` won't work with `:`. I'll have to fix that.

However, the second test case surprising for me. Is it really the case that Snowflake does not allow spaces around `:` operator? In that case we'll have to treat it as part of identifier instead, otherwise extra whitespace always gets allowed.

src/languages/snowflake/snowflake.formatter.ts
nene: Yep that's fine. The spaces will be taken care of by `expandPhrases()`. In the end though any number of whitespace characters between these words will be allowed by the parser.

src/languages/snowflake/snowflake.formatter.ts
nene: Looks fine to me. I'd suggest simplifying the first one to just:

```
[$][0-9]+
```

It's OK if we accept SQL that's not fully correct - the goal is formatting, not validation. It would only pose a problem with `$0` would be some sort of special syntax that we'd like to handle elsewhere.

nene:
> Is there an option to define how identifiers are allowed to look like?

Yep, there are two options for that:

- `identTypes` describes quoted identifiers (with all characters allowed inside quotes).
- `identChars` describes unquoted identifiers.

Both of these seem to be configured correctly.

> Is there an option to test variables?

No, the variable syntax differs quite a bit between SQL dialects. Have a look at the variable tests in `behavesLikeMariaDbFormatter.ts` and create some custom tests for Snowflake.

> Did I miss any obvious support... tests?

Doesn't seem so at first glance.

test/snowflake.test.ts
primeapple: I'll investigate this.

test/features/truncateTable.ts
```diff
@@ -18,6 +19,16 @@ export default function supportsTruncateTable(
     `);
   });
 
+  if (ifExists) {
+    it('formats TRUNCATE TABLE IF EXISTS statments', () => {
+      const result = format('TRUNCATE TABLE IF EXISTS Customers;');
+      expect(result).toBe(dedent`
+        TRUNCATE TABLE IF EXISTS
+          Customers;
+`);
```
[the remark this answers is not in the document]
nene: Though this test too is not that useful. This syntax is only supported by Snowflake, and like elsewhere, testing all these syntax variations doesn't really add much value.

test/snowflake.test.ts
```diff
@@ -0,0 +1,81 @@
...
+  supportsJoin(format, { without: ['NATURAL INNER JOIN'] });
+  supportsSetOperations(format, ['UNION', 'UNION ALL', 'MINUS', 'EXCEPT', 'INTERSECT']);
+  supportsLimiting(format, { limit: true, offset: true, fetchFirst: true, fetchNext: true });
+
+  it('allows $ character as part of identifiers if inclosed by "', () => {
+    expect(format('SELECT "foo$"')).toBe(dedent`
+      SELECT
+        "foo$"
+    `);
+  });
```
nene, line 60: We should instead test that $ character can be used inside unquoted identifiers. The quoted ones are already covered by `supportsIdentifiers()` test.

src/languages/snowflake/snowflake.formatter.ts
```diff
@@ -0,0 +1,352 @@
...
+  'HAVING',
+  'PARTITION BY',
+  'ORDER BY',
+  'LIMIT',
+  'OFFSET',
+  'FETCH [FIRST | NEXT]',
+  // Data manipulation
+  // - insert:
+  'INSERT [OVERWRITE] [ALL INTO | INTO | ALL | FIRST]',
+  '[THEN | WHEN] INTO',
```
nene, line 36: This is not correct. As I understand, the syntax is:

```
WHEN condition THEN INTO tablename
ELSE INTO tablename
```

We could break this up into the following:

```
'WHEN',
'THEN INTO',
'ELSE',
```

test/snowflake.test.ts
primeapple: Wow, I did not know this. It does accept `:` with spaces around!

primeapple: I fixed these tests and added some for the bracket and dot notation as well. Now the only thing failing is the `:` operator since it gets succeeding whitespace.

src/languages/snowflake/snowflake.formatter.ts
```diff
@@ -0,0 +1,352 @@
...
+
+const reservedSetOperations = expandPhrases(['UNION [ALL]', 'MINUS', 'EXCEPT', 'INTERSECT']);
+
+const reservedJoins = expandPhrases([
+  '[INNER] JOIN',
+  '[NATURAL] {LEFT | RIGHT | FULL} [OUTER] JOIN',
+  '{CROSS | NATURAL} JOIN',
+]);
+
+const reservedPhrases = expandPhrases(['{ROWS | RANGE} BETWEEN']);
```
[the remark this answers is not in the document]
primeapple: Snowflake does not support these.

test/features/schema.ts
```diff
@@ -2,11 +2,11 @@ import dedent from 'dedent-js';
 
 import { FormatFn } from 'src/sqlFormatter';
 
-export default function supportsSchema(format: FormatFn) {
-  it('formats simple SET SCHEMA statements', () => {
-    const result = format('SET SCHEMA schema1;');
+export default function supportsSchema(format: FormatFn, setKeyword = 'SET') {
```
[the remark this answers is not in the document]
primeapple: You're right, restored this.

test/features/createView.ts
```diff
@@ -56,4 +57,13 @@ export default function supportsCreateView(
       `);
     });
   }
+
+  if (ifNotExists) {
+    it('formats short CREATE TABLE IF NOT EXISTS', () => {
```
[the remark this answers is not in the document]
primeapple: Do you want me to open an issue for this?

src/languages/snowflake/snowflake.formatter.ts
primeapple: Ahh, you are right. Good find.
We can just remove `WHEN` and write `[THEN | ELSE] INTO`

src/languages/snowflake/snowflake.formatter.ts
primeapple: I see your point, but why not just leave it? It's not that complicated, I can rewrite it to 
`[$][1-9][0-9]*`
if that's better readable?

primeapple:
I understand your concerns regarding the tests.
However in my opinion they have the purpose of explaining the accepted syntax without having to look into the actual source files.

For example with `supportsTruncateTable(format, { withoutTable: true, ifExists: true })` we can very easily see all the dialects that support `TRUNCATE IF EXISTS`, without having to manually look into the source files?

We've 3 options:
1. Leave the tests in this PR, add the other languages that support them later (I can do this if you want?)
2. Add the tests completly in a later PR
3. Remove the tests

Option 3. would decrease test coverage and verbosity for the sake of a few less lines and some milisecons of test time.

You are the maintainer, it's up to you.

test/features/createView.ts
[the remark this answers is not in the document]
nene: No worries, I mainly just wrote this comment as a not for myself.

src/languages/snowflake/snowflake.formatter.ts
nene: Might as well just leave it as is.

nene:
OK, let the tests be for now as they are. My main concern was with the `SET SCHEMA` test. Others are not that big of a deal.

nene:
OK. I fixed the `:` handling. If you now merge in the `master`, it should work.

Though now you'll also need to add the `::` operator to the `alwaysDenseOperators` config.

primeapple:
Everything done.
Is this mergable now? :smiley:

src/languages/snowflake/snowflake.formatter.ts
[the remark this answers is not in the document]
nene: Does it really not? I see them in [snowflake documentation.](https://docs.snowflake.com/en/sql-reference/sql/create-table-constraint.html)

Plus, to test these, we should include `supportsConstraints()` helper.

src/languages/snowflake/snowflake.formatter.ts
```diff
@@ -0,0 +1,352 @@
...
+  'HAVING',
+  'PARTITION BY',
+  'ORDER BY',
+  'LIMIT',
+  'OFFSET',
+  'FETCH [FIRST | NEXT]',
+  // Data manipulation
+  // - insert:
+  'INSERT [OVERWRITE] [ALL INTO | INTO | ALL | FIRST]',
+  '[THEN | ELSE] INTO',
```
nene, line 36: Should instead use:
```suggestion
  '{THEN | ELSE} INTO',
```

test/snowflake.test.ts
```diff
@@ -0,0 +1,87 @@
...
+        SELECT
+          foo:bar.baz
+      `);
+    });
+    it(`formats brackets notation without spaces`, () => {
+      expect(format(`SELECT foo [ 'bar' ]`)).toBe(dedent`
+        SELECT
+          foo['bar']
+      `);
+    });
```
nene, line 78: No need for this test, it's already covered by `supportsArrayAndMapAccessors()`

test/features/truncateTable.ts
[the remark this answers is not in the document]
nene: Let's also fix this inconsistent indentation in here.

nene (commented):
Found a few more issues.

primeapple:
I somehow can't answer your comments, so I'll do it here:

I added to reservedPhrases:
```js
  'MATCH {FULL | SIMPLE | PARTIAL}',
  'ON {UPDATE | DELETE} {CASCADE | SET NULL | SET DEFAULT | RESTRICT | NO ACTION}',
  'INITIALLY {DEFERRED | IMMEDIATE}',
```
This breaks the `supportConstraints` tests, because we are doing `ON UPDATE CURRENT_TIMESTAMP` there.
This test is problematic anyway because we are using it for many languages that don`t even support this line (Postgresql for example).

I fixed the test, we can add the correct specifications for other languages later.

nene (approved):

nene:
Thanks for putting up with all my complaints :)

I'll merge this in and create a beta release out of it.

primeapple:
Thank you for all the fast feedback. It was my first open source contribution and you made the process very easy.

nene:
It's now released as 11.0.0-beta.1

Issue #452, opened 2022-09-20
Explicitly include a single dialect


rogerfar:
We are using this library in production in the browser, so bundle sizes matter. 

When I analyze my bundle app, the sql-formatter library uses about 443kb, but a large part of this is all the dialetcs:

![image](https://user-images.githubusercontent.com/1142915/191344127-46b8a948-1463-41a0-b017-d0f00e55efc3.png)

We only need 1 dialect, so there is no need for the others to be in the bundle.

Would it be possible to have an option to explicitly import a language so that the rest can be tree-shaken from the bundle?

I was thinking, maybe something like this:
`import { format } from 'sql-formatter/sqlite';`

nene:
Yeah, that's definitely a feature that would be nice to have.

I did a quick comparison of the minified bundle sizes:

- with all dialects included: 256KiB
- with only the default "sql" dialect: 77KiB

Possibly a better approach syntax-wise would be to separately export the core engine and language definitions. Something like:

```ts
import { format } from 'sql-formatter/core';
import { sqlite } from 'sq;-formatter/sqlite';

format("SELECT *", { language: sqlite });
```

The `language` option already accepts passing in a custom dialect implementation, so that would work out more naturally.

However, currently this feature isn't really a priority. And for the time being, it's the first time such code-splitting has been requested.

To put things in perspective, the node-sql-parser library has [the same issue](https://github.com/taozhi8833998/node-sql-parser/issues/1100) and in its case each language definition is way larger than in sql-formatter.

verhovsky:
[highlight.js](https://highlightjs.org/usage/) provides this API for this:

```js
import hljs from 'highlight.js/lib/core'
import python from 'highlight.js/lib/languages/python'
hljs.registerLanguage('python', python)
hljs.highlightElement('<some Python in html>', {language: 'python'})
```

I wouldn't like want to pass Formatter objects everywhere for the same reason you didn't make this the API initially. It's tedious to import and pass around an object instead of just a string. If I have a bunch of SQL snippets and I'm storing the SQL dialect with the snippet, if I'm doing anything with those snippets besides passing them to sql-formatter, I'm storing the information "what dialect is this SQL snippet in?" as a string, not as an sql-formatter Formatter object. So that means in my application I'm going to have to have a mapping of `languageName -> FormatterObject` somewhere, but it makes much more sense for that mapping to live in the library, not everyone's applications.

[...]

PR #455, opened 2022-09-23
Update package to ESM

inferrinizzard:
- updated webpack and babel to ESM
- added additional entry points for each language
- TODO: debug tree shaking

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,18 +1,6 @@
 {
-  "presets": [
-    [
-      "@babel/preset-env",
-      {
-        "loose": true,
-        "modules": false
-      }
-    ],
-    "@babel/preset-typescript"
-  ],
+  "presets": [["@babel/preset-env", { "modules": false }], "@babel/preset-typescript"],
   "plugins": [
-    ["@babel/plugin-proposal-class-properties", { "loose": false }],
-    ["@babel/plugin-proposal-private-methods", { "loose": false }],
-    ["@babel/plugin-proposal-private-property-in-object", { "loose": false }],
     "add-module-exports",
     [
       "module-resolver",
--- a/.babelrc
+++ b/.babelrc
@@ -1,7 +1,18 @@
 {
-  "presets": ["@babel/preset-env", "@babel/preset-typescript"],
+  "presets": [
+    [
+      "@babel/preset-env",
+      {
+        "loose": true,
+        "modules": false
+      }
+    ],
+    "@babel/preset-typescript"
+  ],
   "plugins": [
-    "@babel/plugin-proposal-class-properties",
+    ["@babel/plugin-proposal-class-properties", { "loose": false }],
+    ["@babel/plugin-proposal-private-methods", { "loose": false }],
+    ["@babel/plugin-proposal-private-property-in-object", { "loose": false }],
     "add-module-exports",
     [
       "module-resolver",
--- a/package.json
+++ b/package.json
@@ -3,7 +3,8 @@
   "version": "10.7.1",
   "description": "Format whitespace in a SQL query to make it more readable",
   "license": "MIT",
-  "main": "lib/index.js",
+  "main": "dist/sql-formatter.min.js",
+  "module": "lib/index.js",
   "types": "lib/src/index.d.ts",
   "bin": {
     "sql-formatter": "./bin/sql-formatter-cli.js"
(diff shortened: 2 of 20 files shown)


nene:
I'm pretty confident this setup won't work as indented.

In each of these files in `export/` dir you're importing `format` from `src/sqlFormatter.ts` which depends on all the languages being loaded.

.babelrc
```diff
@@ -1,7 +1,18 @@
...
+        "modules": false
+      }
+    ],
+    "@babel/preset-typescript"
+  ],
   "plugins": [
-    "@babel/plugin-proposal-class-properties",
+    ["@babel/plugin-proposal-class-properties", { "loose": false }],
+    ["@babel/plugin-proposal-private-methods", { "loose": false }],
+    ["@babel/plugin-proposal-private-property-in-object", { "loose": false }],
```
nene, line 15: I don't think we need these. Especially as we use typescript which has its own syntax for private fields.

nene:
For the tree-shaking to work we need to decouple language definitions from the `format()` function. For example as I roughly outlined in #452:

- We should export a base-formatter-function which takes `language: Formatter` option.
- And also export each of the `*Formatter` classes we have.

We probably don't need separate paths for the exports, something like this should work as well:

```ts
import { baseFormat, BigQueryFormatter } from 'sql-formatter`;

baseFormat("SELECT *", {language: BigQueryFormatter});
```

I didn't manage to come up with a good name for it though (`baseFormat` doesn't sound right). As it's really an alternative API for using the formatter, it might indeed be better to expose it under a separate path, like:

```ts
import { format, BigQueryFormatter } from 'sql-formatter/core`;
```

It might also be better to name these exported language definitions with simpler names, like:

```ts
import { format, bigquery } from 'sql-formatter/core`;
```

This way the name will be less implementation-dependant.

nene:
There's currently one additional catch with making such a change. The `ExpressionFormatter.formatOperator()` contains currently a check `this.cfg.language === 'plsql'`. This won't work when language parameter is no more a string. We really should eliminate this language-specific switch from there and instead have some other config option like `denseAtOperator: boolean`. Currently there's no place for defining such language-specific formatting-configs. This could be achieved by adding a `formatterOptions()` method to `Formatter` class, which could be overridden by subclasses (in addition to the current `tokenizer()` method).

nene:
In general I think it's better to split this whole task into two independent parts:

- providing an ESM version of sql-formatter.
- decoupling individual language definitions from the core formatter.

.babelrc
```diff
@@ -1,7 +1,18 @@
...
+      {
+        "loose": true,
+        "modules": false
+      }
+    ],
+    "@babel/preset-typescript"
+  ],
   "plugins": [
-    "@babel/plugin-proposal-class-properties",
+    ["@babel/plugin-proposal-class-properties", { "loose": false }],
```
nene, line 13: `loose` option defaults to `false`, so it's unnecessary to specify it here.

.babelrc
```diff
@@ -1,7 +1,18 @@
 {
-  "presets": ["@babel/preset-env", "@babel/preset-typescript"],
+  "presets": [
+    [
+      "@babel/preset-env",
+      {
+        "loose": true,
```
nene, line 6: Why turn on the loose mode?

package.json
```diff
@@ -3,11 +3,17 @@
   "version": "10.7.1",
   "description": "Format whitespace in a SQL query to make it more readable",
   "license": "MIT",
-  "main": "lib/index.js",
+  "main": "dist/sql-formatter.min.js",
+  "module": "lib/index.js",
   "types": "lib/src/index.d.ts",
+  "exports": {
+    ".": "./lib/index.js"
+  },
```
nene, line 11: I wonder if it's really worth it to specify both `module` and `exports` fields. The `module` field was a proposed way of specifying exported module, but was never made official. Later the `exports` field was adopted in its place. Support for `exports` was added 2019 in NodeJS 12, which is no more actively supported.

I think `main` should serve as a good-enough fallback for those using some old NodeJS or some old bundler.

inferrinizzard: you're correct in that module is not an official spec for npm packaging purposes but it is used unofficially by webpack for their new experimental ESM export config

.babelrc
inferrinizzard: it errors otherwise with the loose transform but I'll actually turn that off

nene (approved):
LGTM

inferrinizzard:
Merging this as an initial conversion to ESM, tree-shaking to be adding in a later PR

PR #457, opened 2022-09-24
Support empty SELECT; statement

### LINKED ISSUE #447: [FORMATTING] Postgres: Valid query is triggering a syntax error
**Input data**

```sql
SELECT;
```

Though counter-intuitive, this is actually a valid query in Postgres. It returns a table with 1 row and 0 columns.

**Expected Output**

```sql
SELECT;
```

**Actual Output**

```
Error: Parse error at token: ;
Unexpected DELIMITER token: {"type":"DELIMITER","raw":";","text":";","start":6,"end":7}. Instead, I was expecting to see one of the following:
# <error message truncated>
```

**Usage**

```js
format("SELECT;, {
    language: "postgresql"
})
```

nene:
Fixes #447

### DIFF
--- a/src/parser/grammar.ne
+++ b/src/parser/grammar.ne
@@ -92,6 +92,13 @@ select_clause -> %RESERVED_SELECT (all_columns_asterisk expression:* | asteriskl
     children: [exp, ...expressions],
   })
 %}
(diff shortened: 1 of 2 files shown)


nene:
I'll merge this in and do a release as it's really a trivial change.

PR #459, opened 2022-09-26
Parse CASE expressions

nene:
Implemented proper parsing of case expressions.

Also changed the formatting so that instead of:

```sql
CASE
  foo
  WHEN 1 THEN 'A'
  ELSE 'B'
END
```

we now format the expression following `CASE` on the same line:

```sql
CASE foo
  WHEN 1 THEN 'A'
  ELSE 'B'
END
```

### DIFF
--- a/src/formatter/ExpressionFormatter.ts
+++ b/src/formatter/ExpressionFormatter.ts
@@ -185,20 +185,16 @@ export default class ExpressionFormatter {
   }
 
   private formatCaseWhen(node: CaseWhenNode) {
-    this.withComments(node.when, () => {
-      this.layout.add(WS.NEWLINE, WS.INDENT, this.showKw(node.when), WS.SPACE);
-    });
+    this.layout.add(WS.NEWLINE, WS.INDENT);
+    this.formatNode(node.when);
     this.layout = this.formatSubExpression(node.condition);
-    this.withComments(node.then, () => {
-      this.layout.add(this.showKw(node.then), WS.SPACE);
-    });
+    this.formatNode(node.then);
     this.layout = this.formatSubExpression(node.result);
   }
 
   private formatCaseElse(node: CaseElseNode) {
-    this.withComments(node.else, () => {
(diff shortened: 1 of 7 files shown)


src/parser/grammar.ne
```diff
@@ -235,17 +268,20 @@ keyword ->
...
   ([[token]]) => toKeywordNode(token)
 %}
 
+other_keyword ->
+  ( %WHEN
+  | %THEN
+  | %ELSE
+  | %END ) {%
+  ([[token]]) => toKeywordNode(token)
+%}
```
nene, line 283: This is to allow for these keywords to be used outside of CASE expressions, while only allowing them in specific locations inside a CASE expression. Otherwise we would get an ambiguous grammar.

src/formatter/ExpressionFormatter.ts
```diff
@@ -354,20 +389,8 @@ export default class ExpressionFormatter {
...
-      case TokenType.CASE:
-        return this.formatCaseStart(node);
-      case TokenType.END:
-        return this.formatCaseEnd(node);
-      case TokenType.WHEN:
-      case TokenType.ELSE:
-        return this.formatCaseWhenOrElse(node);
       default:
-        throw new Error(`Unexpected token type: ${node.tokenType}`);
+        return this.formatKeyword(node);
```
nene, line 393: Removed the error to turn this method into a more generic one that's used for formatting all kinds of keywords.

src/parser/grammar.ne
```diff
@@ -210,6 +216,33 @@ between_predicate -> %BETWEEN _ simple_expression _ %AND _ simple_expression {%
   })
 %}
 
+case_expression -> %CASE _ simple_expression:* case_clause:* _ %END {%
```
inferrinizzard, line 219: Can you have a CASE with no WHEN , only END ?

inferrinizzard (approved):

src/parser/grammar.ne
nene: Probably not, but it's not really important in our case. Just like we allow CASE with only an ELSE or even multiple ELSEs.

PR #461, opened 2022-09-26
Better API for adding comments to nodes

nene:
The new code is not shorter, but I think it's simpler to read this way. Especially the cases where we're adding both leading and trailing comments.

### DIFF
--- a/src/parser/grammar.ne
+++ b/src/parser/grammar.ne
@@ -23,11 +23,20 @@ const toKeywordNode = (token: Token): KeywordNode => ({
   raw: token.raw,
 });
 
-const addLeadingComments = (node: AstNode, comments: CommentNode[]): AstNode =>
-  comments.length > 0 ? { ...node, leadingComments: comments } : node;
+interface CommentAttachments {
+  leading?: CommentNode[];
+  trailing?: CommentNode[];
+}
 
-const addTrailingComments = (node: AstNode, comments: CommentNode[]): AstNode =>
-  comments.length > 0 ? { ...node, trailingComments: comments } : node;
+const addComments = (node: AstNode, { leading, trailing = [] }: CommentAttachments): AstNode => {
+  if (leading && leading.length > 0) {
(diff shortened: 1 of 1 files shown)


src/parser/grammar.ne
```diff
@@ -23,11 +23,20 @@ const toKeywordNode = (token: Token): KeywordNode => ({
...
-  comments.length > 0 ? { ...node, leadingComments: comments } : node;
-
-const addTrailingComments = (node: AstNode, comments: CommentNode[]): AstNode =>
-  comments.length > 0 ? { ...node, trailingComments: comments } : node;
+interface CommentAttachments {
+  leading?: CommentNode[];
+  trailing?: CommentNode[];
+}
+
+const addComments = (node: AstNode, { leading, trailing = [] }: CommentAttachments): AstNode => {
```
inferrinizzard, line 31: Should both or neither of `leading` and `trailing` have empty array initialisers?

src/parser/grammar.ne
```diff
@@ -23,11 +23,20 @@ const toKeywordNode = (token: Token): KeywordNode => ({
...
+  if (leading && leading.length > 0) {
```
inferrinizzard, line 32: can be
```
return {
  ...node,
  leading: leading?.length ? leading : undefined,
  trailing: trailing?.length ? trailing : undefined,
}
```

inferrinizzard (approved):

src/parser/grammar.ne
nene: Oops... this default value was indeed forgotten in.

src/parser/grammar.ne
nene: Thought of doing this originally, but wanted to completely avoid empty fields on the nodes.

But using `leading?.length` syntax would definitely improve things.

nene: Added PR with fixes #465

PR #466, opened 2022-09-27
Add dialect-specific format options

nene:
Instead of checking if `language === 'plsql'` the PL/SQL language definition now provides an options object with `{alwaysDenseOperators: ['@']}` which is used instead. This decouples the formatting logic from any specific language name.

Implementing what I suggested in here: #455


nene:
Merging this in, as it will be useful in implementing Snowflake support.

PR #467, opened 2022-09-27
Fix ESM & Commonjs exports definitions

### LINKED ISSUE #105: build es6 version of the sql-formatter

nene:
Fixes #105


package.json
```diff
@@ -3,11 +3,15 @@
...
-  "main": "dist/sql-formatter.min.js",
+  "main": "dist/sql-formatter.min.cjs",
   "module": "lib/index.js",
   "types": "lib/src/index.d.ts",
   "exports": {
-    ".": "./lib/index.js"
+    ".": {
+      "import": "./lib/index.js",
+      "require": "./dist/sql-formatter.min.cjs",
+      "types": "./lib/src/index.d.ts"
```
JounQin, line 13: > the "types" condition should always come first in "exports"

See https://devblogs.microsoft.com/typescript/announcing-typescript-4-7/#package-json-exports-imports-and-self-referencing

JounQin (changes requested):

package.json
nene: Thanks a lot. Wouldn't have guessed this gotcha on my own.

package.json
```diff
@@ -3,11 +3,15 @@
...
+    }
   },
   "bin": {
     "sql-formatter": "./bin/sql-formatter-cli.js"
```
JounQin, line 17: The bin file has not been migrated to ESM, so `.cjs` suffix is still required, and

```diff
- const { format, supportedDialects } = require('../lib/index');
+ const { format, supportedDialects } = require('./dist/sql-formatter.min.cjs');
```

JounQin (changes requested):

package.json
nene: Thanks again.

I'm really not good at this ESM conversion :(

JounQin (approved):

webpack.common.js
```diff
@@ -4,7 +4,7 @@ export default {
   entry: './src/index.ts',
   output: {
     path: path.resolve('./dist'),
-    filename: 'sql-formatter.js',
+    filename: 'sql-formatter.cjs',
     library: { type: 'module' },
```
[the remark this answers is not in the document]
nene: Indeed, that's clearly incorrect. Not sure why the webpack build was changed to module.

PR #483, opened 2022-10-01
Split clauses into two groups: multiline & single line

### LINKED ISSUE #456: [FORMATTING] [MySQL] Newline after USE keyword causes syntax error
**Input data**

```sql
CREATE DATABASE IF NOT EXISTS sample;

USE sample;
```

```sh
sql-formatter -l mysql
```

**Expected Output**

```sql
CREATE DATABASE
  IF NOT EXISTS sample;

USE sample;
```

**Actual Output**

```sql
CREATE DATABASE
  IF NOT EXISTS sample;

USE
  sample;
```

**Why?**

The current formatting result causes syntax error on MySQL 8.0.30. 
It seems should not place `\n` between `USE` and `<DB NAME>`.

```bash
$ mysql --version
mysql  Ver 8.0.30 for Linux on x86_64 (MySQL Community Server - GPL)

$ cat test.sql
CREATE DATABASE
  IF NOT EXISTS sample;

USE
  sample;
$ mysql < test.sql
ERROR at line 4: USE must be followed by a database name
ERROR 1064 (42000) at line 5: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'sample' at line 1
```

**Usage**

    - `cat test.sql | sql-formatter -l mysql`
    - `mysql`
    - `10.6.0`

nene:
So far pretty much all SQL clauses have been formatted like so:

```sql
CLAUSE_NAME
  more sql here
```

However, for many clauses this is not really how one would normally format SQL. For example, one rarely formats SQL like so:

```sql
TRUNCATE TABLE
  customers;
```

Rather one would format this on a single line:

```sql
TRUNCATE TABLE customers;
```

This change makes it so, that by default all clauses get formatted in the latter way. i.e. all clauses for which we don't have specific tests as to how we'd want them to be formatted. Most dialects have this sort of kitchen-sink list of clauses that's just been copied there from documentation. So these are all now being formatted in this one-line style. This will also aligns more with how the default `sql`-dialect formatter handles unknown SQL syntax.

Also hand-picked a few clauses for which I wanted to intentionally apply this style:

- `TRUNCATE TABLE`
- `SET SCHEMA`
- `ALTER TABLE`
- `DROP TABLE`
- `DELETE FROM`
- `UPDATE`

As a bonus, this also:
- Fixes #456


src/formatter/ExpressionFormatter.ts
```diff
@@ -45,12 +45,21 @@ interface ExpressionFormatterParams {
 export interface DialectFormatOptions {
   // List of operators that should always be formatted without surrounding spaces
   alwaysDenseOperators?: string[];
+  // List of clauses that should be formatted on a single line
+  onelineClauses: string[];
+}
+
+// Contains the same data as DialectFormatOptions,
+// but optimized for faster and more conventient lookup.
+export interface ProcessedDialectFormatOptions {
```
inferrinizzard, line 54: can we not just use this as the default interface for DialectFormatOptions and add a wrapper function to transform the string[] into Record ?

src/formatter/ExpressionFormatter.ts
```diff
@@ -45,12 +45,21 @@ interface ExpressionFormatterParams {
...
+  alwaysDenseOperators: string[];
+  onelineClauses: Record<string, boolean>;
```
inferrinizzard, line 56: can be
```suggestion
  onelineClauses: Record<string, true>;
```

inferrinizzard (approved):

src/formatter/ExpressionFormatter.ts
nene: Tried that, but in this case TypeScript thinks that `onlineClauses[anything]` always evaluates to true.

src/formatter/ExpressionFormatter.ts
nene: It's not only the `onlineClauses` field that differs. It's also the `alwaysDenseOperators` field which is originally optional, but then processed to be empty array if missing.

Issue #486, opened 2022-10-02
11.0.0 does not work well with Typescript module=NodeNext/Node16


akphi:
**Describe the bug**
After upgrading to `sql-formatter@11.0.0`, my project using `module: NodeNext` for Typescript no-longer compiles.

**Expected behavior**
It should compile just fine like in version `10.7.2`

**Actual behavior**
Checkout the code from https://github.com/akphi/issue-repo/pull/10
Run `tsc`:

```sh
yarn install
yarn dev:ts
```

I got the following error:

```
// skipLibCheck=true

src/index.ts:1:10 - error TS2305: Module '"sql-formatter"' has no exported member 'format'.

1 import { format } from "sql-formatter";
           ~~~~~~


// skipLibCheck=false

node_modules/sql-formatter/lib/src/index.d.ts:1:15 - error TS2834: Relative import paths need explicit file extensions in EcmaScript imports when '--moduleResolution' is 'node16' or 'nodenext'. Consider adding an extension to the import path.

1 export * from './sqlFormatter';
                ~~~~~~~~~~~~~~~~

node_modules/sql-formatter/lib/src/index.d.ts:2:102 - error TS2834: Relative import paths need explicit file extensions in EcmaScript imports when '--moduleResolution' is 'node16' or 'nodenext'. Consider adding an extension to the import path.

2 export type { IndentStyle, KeywordCase, CommaPosition, LogicalOperatorNewline, FormatOptions, } from './FormatOptions';
                                                                                                       ~~~~~~~~~~~~~~~~~

node_modules/sql-formatter/lib/src/index.d.ts:3:38 - error TS2834: Relative import paths need explicit file extensions in EcmaScript imports when '--moduleResolution' is 'node16' or 'nodenext'. Consider adding an extension to the import path.

3 export { default as Formatter } from './formatter/Formatter';
                                       ~~~~~~~~~~~~~~~~~~~~~~~

node_modules/sql-formatter/lib/src/index.d.ts:4:38 - error TS2834: Relative import paths need explicit file extensions in EcmaScript imports when '--moduleResolution' is 'node16' or 'nodenext'. Consider adding an extension to the import path.

4 export { default as Tokenizer } from './lexer/Tokenizer';
                                       ~~~~~~~~~~~~~~~~~~~

[...]

PR #487, opened 2022-10-02
Use `NodeNext` for `Typescript` module resolution

akphi:
Fixes https://github.com/sql-formatter-org/sql-formatter/issues/486

@nene  This is an extensive fix to upgrade the codebase to fully embrace ESM (i.e. specifying extensions in exports). This will make `sql-formatter` naturally work with Typescript projects which use `module = Node16/NodeNext`. Otherwise, we will have to find some other way to tweak `package.json` to make things work. 

For example, `uuid`

https://github.com/uuidjs/uuid/blob/main/package.json

doesn't use the field `type: module` and fully specify `exports` but I personally think it's just makeshift, we might as well move to adopt `ESM` fully.

---

I have run `yarn build` and `yarn test` locally to make sure things work.
One thing you might not like is I don't think we could use alias like `src` anymore so I resolved all imports to use relative paths, sorry about that 🙏


nene:
Thanks for the bug report and pull request.

It turns out that the [ts-transform-paths](https://www.npmjs.com/package/@zerollup/ts-transform-paths) plugin [doesn't work with latest TypeScript](https://github.com/zerkalica/zerollup/issues/37). It should be compiling away these `src/` paths, but it's not working. I tried applying some of the suggested patches, but they also did nothing. Given that this issue has remained unsolved for about a year, it's probably better to move away from using of that plugin.

nene (approved):

nene:
These fixes, plus a few more, are now released as 11.0.1

PR #488, opened 2022-10-03
Cleanup ESM/TypeScript build

nene:
Additional fixes/cleanup on top of  #487

### DIFF
--- a/src/parser/createParser.ts
+++ b/src/parser/createParser.ts
@@ -1,4 +1,4 @@
-import { Parser as NearleyParser, Grammar } from 'nearley';
+import nearley from 'nearley';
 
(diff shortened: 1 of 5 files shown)


src/parser/createParser.ts
```diff
@@ -8,6 +8,8 @@ import grammar from './grammar.js';
 import LexerAdapter from './LexerAdapter.js';
 import { createEofToken } from '../lexer/token.js';
 
+const { Parser: NearleyParser, Grammar } = nearley;
```
nene, line 11: This is the most important change.

Now that our code is a full ESM module, NodeJS complains when attempting to import named things from CommonJS module. The workaround is to get the default export and destructure it manually.

PR #511, opened 2022-11-02
Single dialect import

### LINKED ISSUE #452: Explicitly include a single dialect

nene:
New API for supporting import of a single dialect. As discussed in #484, I've added a separate formatting function that works a bit differently than the old one.

The old `format()` function works like before, with the exception that `language` parameter can only be string:

```ts
import { format } from "sql-formatter";

format("SELECT *", { language: "mysql" });
```

The new `formatDialect()` only works with explicitly imported dialect:

```ts
import { formatDialect, mysql } from "sql-formatter";

formatDialect("SELECT *", { dialect: mysql });
```

`formatDialect()` is also used under the hood by the old `format()` function.

I chose to name it `formatDialect` to make it clear that it accepts a `dialect` config option (instead of `language`). This also aligns with the naming of types, where the actual value is a `DialectOptions` object. (Originally I thought of using `language` option for both, but this seemed too confusing - an option expecting one type in one context and another type in another context - so I opted for clearly distinct option names).

Fixes #452

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,4 +1,12 @@
-export * from './sqlFormatter.js';
+export {
+  SqlLanguage,
+  supportedDialects,
+  FormatOptionsWithLanguage,
+  FormatOptionsWithDialect,
+  format,
+  formatDialect,
+  ConfigError,
+} from './sqlFormatter.js';
 export type {
   IndentStyle,
   KeywordCase,
--- a/src/index.ts
+++ b/src/index.ts
@@ -6,14 +6,14 @@ export {
   format,
   formatDialect,
 } from './sqlFormatter.js';
-export type {
+export {
   IndentStyle,
   KeywordCase,
   CommaPosition,
   LogicalOperatorNewline,
   FormatOptions,
 } from './FormatOptions.js';
-export type { DialectOptions } from './dialect.js';
+export { DialectOptions } from './dialect.js';
 export { ConfigError } from './validateConfig.js';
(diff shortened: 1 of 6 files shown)


src/index.ts
```diff
@@ -1,4 +1,17 @@
-export * from './sqlFormatter.js';
+export { supportedDialects, format, formatDialect } from './sqlFormatter.js';
+export { expandPhrases } from './expandPhrases.js';
+
+// Intentionally use "export *" syntax here to make sure when adding a new SQL dialect
+// we wouldn't forget to expose it in our public API.
+export * from './allDialects.js';
+
+// NB! To re-export types the "export type" syntax is required by webpack.
+// Otherwise webpack build will fail.
```
inferrinizzard, line 9: there's an eslint-typescript rule here we can leverage, `use-consistent-type-exports`

inferrinizzard: I can make a PR to update eslint rules soon

nene: Good find. I added the rule on my own. Seems to work just as needed.

PR #516, opened 2022-11-04
Add support for custom quote types

### LINKED ISSUE #503: Allow defining custom quote types
**Describe the Feature**
Allow defining a custom `quote` type for a dialect.

As I explained in #306, I would like to define a custom grammar of Trino SQL queries + `{{Handlebars templates}}` so that users can write (and format with sql-formatter) queries like this

```sql
select
  {{ country_code }}
from
  my_table
```

and we could render the query for other users with a dropdown in place of `{{ country_code }}`. The easiest way to do this is to make make sql-formatter treat `{{}}` like a type of string. In order to do this, we need to have a regex for this type of quote in `quotePatterns`. It makes no sense to add this to sql-formatter itself, so we have to fork it to add this custom quote type. It would be nice if I could pass the regex somehow through `DialectOptions`' `tokenizerOptions.stringTypes` or some other way.

nene:
Similarly to how variable types can be specified with a regex string, a string (and identifier) can now also be specified the same way.

Fixes #503

### DIFF
--- a/src/lexer/TokenizerOptions.ts
+++ b/src/lexer/TokenizerOptions.ts
@@ -22,13 +22,13 @@ export interface PrefixedQuoteType {
   requirePrefix?: boolean; // True when prefix is required
 }
 
-export type QuoteType = PlainQuoteType | PrefixedQuoteType;
-
-export interface VariableRegex {
+export interface RegexPattern {
   regex: string;
 }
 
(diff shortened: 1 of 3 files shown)


nene:
It's really a pretty trivial change, so I'll go ahead and merge it in.

PR #517, opened 2022-11-08
Avoid referencing allDialects outside of function calls

### LINKED ISSUE #452: Explicitly include a single dialect

nene:
That will break tree-shaking as allDialects object gets used regardless of whether any function gets executed, so Webpack ends up bundling all the languages.

Fixes #452

### DIFF
--- a/src/sqlFormatter.ts
+++ b/src/sqlFormatter.ts
@@ -5,12 +5,28 @@ import { createDialect, DialectOptions } from './dialect.js';
 import Formatter from './formatter/Formatter.js';
 import { ConfigError, validateConfig } from './validateConfig.js';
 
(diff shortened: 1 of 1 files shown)


nene:
Merging this in to do another beta release that users could try out.

Issue #505, opened 2022-10-24
[FORMATTING] 11.0.2 commaPositon after doesn't work well when come up with comment end with comma


MR-BH:
**Input data**

Which SQL and options did you provide as input?

```sql
SELECT
  name, -- user name,
  COUNT(distinct address) -- address 
FROM
  user;
```

**Expected Output**

```sql
SELECT
  name -- user name,
, COUNT(distinct address) -- address 
FROM
  user;
```

**Actual Output**

```sql
SELECT
  name, -- user name
, COUNT(distinct address) -- address 
FROM
  user;
```

**Usage**

- How are you calling / using the library?
   demo website & ESM
- What SQL language(s) does this apply to?
  {
      language: 'spark',
      commaPosition: 'before',
  }
- Which SQL Formatter version are you using? 
   11.0.2

nene:
Yeah, this is essentially the same bug as #236

The implementation of these features is just very problematic. I've kept them around, though maybe it would be altogether better to just drop these problematic features. For now there's no simple way to properly fix them.

Probably should add a warning to docs to label these as buggy/problematic.

MR-BH:
I am a little busy right now, I will submit a Merge Request to fix this problem when I have some free time, maybe in December.

PR #549, opened 2023-01-09
BUG-505 fix

d3x42:
Add support of comments to SQL formatter

Fix for https://github.com/sql-formatter-org/sql-formatter/issues/505

### DIFF
--- a/src/formatter/formatCommaPositions.ts
+++ b/src/formatter/formatCommaPositions.ts
@@ -32,7 +32,7 @@ export default function formatCommaPositions(
  *     [
  *       'SELECT',
  *       '  foo,',
- *       '  bar,',
+ *       '  bar, --comment',
  *       '  baz',
  *       'FROM'
  *     ]
@@ -41,7 +41,7 @@ export default function formatCommaPositions(
  *
  *     [
  *       ['SELECT'],
- *       ['  foo,', '  bar,', '  baz'],
+ *       ['  foo,', '  bar, --comment', '  baz'],
  *       ['FROM']
  *     ]
  */
@@ -52,7 +52,7 @@ function groupCommaDelimitedLines(lines: string[]): string[][] {
     // when line ends with comma,
     // gather together all following lines that also end with comma,
     // plus one (which doesn't end with comma)
-    while (lines[i].match(/.*,$/)) {
+    while (lines[i].match(/.*,((\s*--.+)|$)/)) {
       i++;
       group.push(lines[i]);
     }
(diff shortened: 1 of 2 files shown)


src/formatter/formatCommaPositions.ts
```diff
@@ -52,7 +52,7 @@ function groupCommaDelimitedLines(lines: string[]): string[][] {
     // when line ends with comma,
     // gather together all following lines that also end with comma,
     // plus one (which doesn't end with comma)
-    while (lines[i].match(/.*,$/)) {
+    while (lines[i].match(/.*,((\s*--.+)|$)/)) {
```
nene, line 55: For some reason `.+` is used here instead of `.*`. This means the regex won't match empty comments.

Also, it has an extra pair of parenthesis.

src/formatter/formatCommaPositions.ts
```diff
@@ -68,7 +68,11 @@ function formatTabular(commaLines: string[]): string[] {
     if (i === commaLines.length - 1) {
       return line; // do not add comma for last item
     }
-    return line + ' '.repeat(maxLineLength - line.length - 1) + ',';
+    const commentMatch = /\s*--/.exec(line);
+    const endOfLinePosition = commentMatch ? commentMatch.index : line.length;
+    return `${line.slice(0, endOfLinePosition)}${' '.repeat(
+      maxLineLength - line.length - 1
+    )},${line.slice(endOfLinePosition + 1)}`;
```
nene, line 75: I have trouble understanding this code. It attempts to do something with comments in the `commaPosition:tabular` case. But there are no tests that cover this scenario.

I tried to write a test for this, but it produced a pretty strange result:

```sql
SELECT
  alpha         ,
  beta,--comment
  delta         ,
  epsilon       ,
  iota
```

nene (changes requested):
Thanks for the PR.

The `commaPosition:before` case seems to work as expected. But the `commaPosition:tabular` implementation is problematic.

src/formatter/formatCommaPositions.ts
d3x42: Thanks, I'll add fixes

src/formatter/formatCommaPositions.ts
d3x42: >  have trouble understanding this code.
This part was refactored to make it easier to undestand

>  I tried to write a test for this, but it produced a pretty strange result:
This case was added to tests. Also I changed code to make it work as expected.

If you'll find other use cases, I'll add them to tests with corresponding fixes.

nene:
Thanks for the fixes.

The logic in `formatTabular()` function was still unnecessarily complex. You had added comments and long descriptive variable names, which helped, but it was still hard to understand what was actually going on. I extracted it into a separate function and completely rewrote the logic inside.

d3x42:
Thank you!

nene:
Released the fix in 12.0.5

PR #556, opened 2023-01-27
Added pre-commit integration.

slyapustin:
This PR introduces direct [pre-commit](https://pre-commit.com/) compatibility.

Here is an example of how it can be integrated after it will be integrated:
```
repos:
-   repo: https://github.com/sql-formatter-org/sql-formatter
    rev: ''
    hooks:
    -   id: sql-formatter
```

Right now you need to create a separate repository in order to be able to integrate with `sql-formatter`.
Like this one: https://github.com/slyapustin/pre-commit-sql-formatter

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,0 +1,7 @@
+-   id: sql-formatter
+    name: sql-formatter
+    description: "Reformat SQL files with sql-formatter"
+    entry: sql-formatter
+    language: node
+    types: [sql]
+    args: [--fix]
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,7 +1,6 @@
 -   id: sql-formatter
     name: sql-formatter
     description: "Reformat SQL files with sql-formatter"
-    entry: sql-formatter
+    entry: sql-formatter --fix
     language: node
     types: [sql]
-    args: [--fix]


nene:
Thanks for the PR, but I don't quite understand why this pre-commit configuration file should be part of this repository.

There can be many tools which might want to use this library. The library itself should not need to include code to cater for each of these tools. Specifically this library provides two well-established API-s:

- calling the `sql-formatter` tool from command line.
- calling the formatting function through JavaScript API.

So far these have been enough for all kinds of tools to integrate with it.

slyapustin:
It just simplifies `sql-formatter` integration with pre-commit hooks. 

And it will not require everyone to have a separate repository for that (like I created).

There are more than [200+ popular projects](https://pre-commit.com/hooks.html) that added integration with the pre-commit hooks already.

nene:
Sorry, but I don't buy it. To me it looks like a shortcoming of this pre-commit tool.

Also, me being completely unfamiliar with this tool makes me a bad maintainer of that integration. I'd rather not take that responsibility upon me.

nene:
Also, clicking through this [list of hooks](https://pre-commit.com/hooks.html) it seems to me that most of the repositories linked from there are either:

- python-based tools
- shallow wrappers for a non-python tools

Found a few rare exceptions. So it looks like adding of this config file might be a standard practice in the world of Python, but a pretty rare thing outside of Python realm.

caffeinatedgaze:
It would really be nice to have it integrated with pre-commit. I am sadly not able to use this tool otherwise. sqlfluff could become a nice alternative for me - but it seems to be very unstable in the way it parses grammar, which renders it useless.

caffeinatedgaze:
[This thread](https://github.com/pre-commit/pre-commit.com/issues/781#issuecomment-1406906153) has good points pro adding this single config file. This may even encourage @nene to reconsider :smile:

To make it easier for the readers, @asottile mentions [eslint](https://github.com/eslint/eslint) and [ruboco](https://github.com/rubocop/rubocop) as examples of off-Python world projects where pre-commit is very well welcomed.

nene:
Why not simply use [pre-commit-sql-formatter](https://github.com/slyapustin/pre-commit-sql-formatter)?

I guess the problem is that these small wrapper-projects become easily out of date with the original library as they need to reference a specific version of the library, unlike when the configuration file is part of the original library repo (at least it seems that a version number is required in the former case, but not in the latter case).

caffeinatedgaze:
> as they need to reference a specific version

This indeed looks valid. In addition, one needs to maintain an irrelevant version tag to fetch a hook from a third-party repo.

PR #574, opened 2023-03-08
Add FOR UPDATE clause support for Postgres

nwalters512:
Like #421, but for Postgres.

IMO things look better if `OF` (if present) is placed on the same like as `FOR ...` and any specific columns are placed on a separate line. I make this change for PL/SQL as well, but I'd be happy to revert that if you feel differently.

### DIFF
--- a/src/languages/postgresql/postgresql.formatter.ts
+++ b/src/languages/postgresql/postgresql.formatter.ts
@@ -18,6 +18,7 @@ const reservedClauses = expandPhrases([
   'LIMIT',
   'OFFSET',
   'FETCH {FIRST | NEXT}',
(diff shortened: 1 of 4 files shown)


src/languages/postgresql/postgresql.formatter.ts
```diff
@@ -18,6 +18,7 @@ const reservedClauses = expandPhrases([
   'LIMIT',
   'OFFSET',
   'FETCH {FIRST | NEXT}',
+  'FOR {UPDATE | NO KEY UPDATE | SHARE | KEY SHARE} [OF]',
```
nwalters512, line 21: This was taken straight from the Postgres docs: https://www.postgresql.org/docs/current/sql-select.html

nene:
Thanks for the. Makes sense to me.

Issue #606, opened 2023-06-08
[FORMATTING] GO Statement Before Procedure Declaration


grantwforsythe:
**Input data**

```tsql
SET
NOCOUNT ON;

GO

CREATE OR ALTER PROCEDURE dbo.sp_say_hi (@Name VARCHAR(25) = 'World')
AS BEGIN
    PRINT CONCAT('Hello, ', @Name);
END;

```

**Expected Output**

```tsql
SET
NOCOUNT ON;

GO

CREATE
OR ALTER
PROCEDURE dbo.sp_say_hi (@Name VARCHAR(25) = 'World')
AS BEGIN
    PRINT CONCAT('Hello, ', @Name);
END;
```

The `GO` statement should be on it's own line.

**Actual Output**

```tsql
SET
NOCOUNT ON;

GO CREATE
OR ALTER
PROCEDURE dbo.sp_say_hi (@Name VARCHAR(25) = 'World') AS BEGIN
    PRINT CONCAT('Hello, ', @Name);
END;
```
When executed on the database a fatal error occurs (`Incorrect syntax was encountered while CREATE was being parsed`) because of the `GO`.

**Usage**

- How are you calling / using the library?
The library is being called from a Git hook with the below config:
  ```json
  {
    "language": "tsql",
    "tab width": 4,
    "useTabs": false,
    "keywordCase": "upper",
    "indentStyle": "standard",
    "logicalOperatorNewline": "before",
    "expressionWidth": 80,
    "linesBetweenQueries": 2,
    "denseOperators": false,
    "newlineBeforeSemicolon": false
  }
  ```
- What SQL language(s) does this apply to?
Transact-SQL/T-SQL
- Which SQL Formatter version are you using?
Latest (`v12.2.2`)

nene:
Thanks, the fix is now released in 12.2.2

nene:
Thanks for reporting.

[...]

PR #609, opened 2023-06-09
Add INTO Clause and Remove Invalid Data Definition

grantwforsythe:
Reverting a change I had made previously and adding `INTO` as a reserved clause.

### DIFF
--- a/test/transactsql.test.ts
+++ b/test/transactsql.test.ts
@@ -118,15 +118,12 @@ describe('TransactSqlFormatter', () => {
     `);
   });
 
-  it('formats SET NOCOUNT ON; GO CREATE OR ALTER PROCEDURE', () => {
-    const result = format('SET NOCOUNT ON; GO CREATE OR ALTER PROCEDURE p');
+  it('formats GO CREATE OR ALTER PROCEDURE', () => {
+    const result = format('GO CREATE OR ALTER PROCEDURE p');
     expect(result).toBe(dedent`
-    SET
-    NOCOUNT ON;
-
-    GO
-    CREATE OR ALTER PROCEDURE
-      p
+      GO
+      CREATE OR ALTER PROCEDURE
+        p
     `);
   });
 
--- a/test/transactsql.test.ts
+++ b/test/transactsql.test.ts
@@ -118,18 +118,6 @@ describe('TransactSqlFormatter', () => {
     `);
   });
 
-  it('formats SET NOCOUNT ON; GO CREATE OR ALTER TABLE', () => {
-    const result = format('SET NOCOUNT ON; GO CREATE OR ALTER TABLE t');
-    expect(result).toBe(dedent`
-    SET
-    NOCOUNT ON;
-
-    GO
-    CREATE OR ALTER TABLE
-      t
-    `);
-  });
-
   it('formats SET NOCOUNT ON; GO CREATE OR ALTER PROCEDURE', () => {
     const result = format('SET NOCOUNT ON; GO CREATE OR ALTER PROCEDURE p');
     expect(result).toBe(dedent`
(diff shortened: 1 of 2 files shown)


test/transactsql.test.ts
```diff
@@ -118,27 +118,37 @@ describe('TransactSqlFormatter', () => {
...
-      p
+  it('formats INSERT INTO', () => {
+    const result = format("INSERT INTO tbl(firstName, lastName) VALUES('George', 'Orwell')");
+    expect(result).toBe(dedent`
+    INSERT INTO
+      tbl (firstName, lastName)
+    VALUES
+      ('George', 'Orwell')
     `);
   });
```
nene, line 153: This test is not needed. It's already covered by the `supportsInsertInto()` call to a shared test.

test/transactsql.test.ts
```diff
@@ -118,27 +118,37 @@ describe('TransactSqlFormatter', () => {
...
+    const result = format('SELECT col INTO #temp FROM tbl');
     expect(result).toBe(dedent`
-    SET
-    NOCOUNT ON;
+    SELECT
+      col
+    INTO
+      #temp
+    FROM
+      tbl
```
nene, line 141: Should indent the expected formatted code by extra 2 spaces, like done in other tests.

test/transactsql.test.ts
```diff
@@ -118,27 +118,37 @@ describe('TransactSqlFormatter', () => {
...
     expect(result).toBe(dedent`
     SET
     NOCOUNT ON;
 
     GO
-    CREATE OR ALTER TABLE
-      t
+    CREATE OR ALTER PROCEDURE
+      p
     `);
```
nene, line 130: Probably best to concentrate this test to just the `GO` statement.

The preceding `SET` statement isn't really relevant here. All the `;`-separated statements are formatted in isolation anyway, so it can't really have any effect to the test outcome.

In general I haven't written tests for all the combinations of `OR ALTER`, `IF EXISTS` etc. Just too much effort for too little benefit. There are some generic tests that ensure the expansion of the `{blah | blah} [something]` syntax works properly.

Therefor in here I would just concentrate on testing `GO` with some simple statement that could follow it. (I personally don't really know that much about TSQL, and especially about the GO-statement).

nene (commented):
Thanks for the PR.

Generally all good. I wrote some comments about the tests.

You can tackle these if you want to, or I'll do that by myself before merging it in.

grantwforsythe:
Thanks so much!

No need to handle these yourself, I will implement your suggestions tomorrow AM (EST).

test/transactsql.test.ts
grantwforsythe: Confirming redundant test has been removed

test/transactsql.test.ts
grantwforsythe: Confirming the indentation has been updated

test/transactsql.test.ts
grantwforsythe: Fair enough, I have removed the `SET` statement from the test.

The problem I noticed with the `GO` statement was when I was defining the stored procedure with `CREATE OR ALTER` as mentioned in #606. Would this be acceptable?

nene: Ah yeah... I misunderstood things a little.

The actual problem you had wasn't really with the `GO` statement. But I think it looked like a problem with formatting of `GO` because the formatter probably formatted it all on one long line.

The thing is that the formatter doesn't actually detect the `GO` statement specifically. It only detects the `CREATE TABLE` or `CREATE PROCEDURE` and other such statements and adds a newline before them (not a newline after `GO`). It just treats `GO` as some unknown statement it doesn't understand and prints it without line breaks. And similarly it treated the whole `GO CREATE OR ALTER PROCEDURE p`.

This behavior is unfortunate, but on the other hand I'm reluctant to add a rule for formatting the `GO` keyword, as it's such a short one it could also easily occur in various other context of the SQL language, and the formatter is pretty poor at understanding the context.

nene:
Included to 12.2.3 release.

PR #621, opened 2023-07-10
fix EAGAIN error for stdin

seblj:
I got some EAGAIN errors and tracked it down to being a problem of using `fs.readFileSync` or `fs.readSync`. It's really hard to reproduce, but it seems like others also have had this problem:

https://github.com/tatethurston/TwirpScript/issues/208
https://github.com/nodejs/help/issues/2663
https://stackoverflow.com/questions/40362369/stdin-read-fails-on-some-input

This seems to fix the problem.

I am not sure if it's the best way to do it, but it seems to work for me at least.

### DIFF
--- a/bin/sql-formatter-cli.cjs
+++ b/bin/sql-formatter-cli.cjs
@@ -23,6 +23,22 @@ class PrettierSQLArgs {
     });
   }
 
+  async getStdin() {
+    if (process.stdin.isTTY) {
+      return Buffer.alloc(0);
+    }
+
+    const result = [];
+    let length = 0;
+
+    for await (const chunk of process.stdin) {
+      result.push(chunk);
+      length += chunk.length;
+    }
+
+    return Buffer.concat(result, length).toString();
+  }
+
   getParser() {
     const parser = new ArgumentParser({
       add_help: true,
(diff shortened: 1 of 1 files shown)


nene:
Thanks for the PR, but I don't have enough details to evaluate this.

What is the actual problem that this PR is fixing. Like, how to reproduce the bug? I also haven't head from anybody else complaining about any issue that looks similar. Or at least I didn't manage to find anything in the issue tracker.

The fix you have included is pretty elaborate. But it seems to be a copy-paste from stack overflow post, which itself has received only a single upvote, so I'm not too keen to trust that. Given that reading stdin is such a common operation, I'm especially doubtful that one would need so much code to handle such a common task correctly.

seblj:
Yes I understand. I only noticed this issue when combining it with a plugin I have for neovim (https://github.com/seblj/nvim-formatter), and when formatting a lot of files at once.

I will try to come up with an example to make it easy to reproduce without formatting a lot of files and without the plugin.

I just made the PR as an example for now and to make you aware that there is an issue with `readFileSync` and stdin

nene:
Thanks. Your efforts are much appreciated. :)

seblj:
Okay I have a reproducible where I expect that it would format it correctly. However there is two different errors now with the fix I have, and what is currently the way it is handled. Using the `get-stdin` package like `prettier` fixes this problem though.

Create a file `foo.sql`

```sql
        CREATE TABLE
  users (
    id uuid PRIMARY KEY DEFAULT uuid_generate_v4 (),
                    email VARCHAR(50) UNIQUE NOT NULL,
    name VARCHAR(150) NOT NULL
  );
```

Run:

`awk '{print $0; system("sleep .1");}' foo.sql | sql-formatter -l postgresql`

This will give you the error of `Error: no file specified and no data in stdin` which is the EAGAIN error.

However, the fix I have produced currently will give another error since it will try to format only the first line since it will not wait until it's finished receiving input from stdin.

I will switch to using the `get-stdin` package like `prettier` in an upcoming commit where you can try the same reproducible, and you will se that it will then format it correctly

nene:
Thanks for the fix @seblj 

For some reason Github failed to notify me about the comments and improvements you did two weeks ago. But thanks to @bitst0rm this PR got back to my attention.

I sort-of made the `get-stdin` library work by simply importing 8.0.0 version (not the latest 9.0.0). Tried a few things to get it to work with the ESM thingie, but then decided to not waste my time and gave up. Doesn't really matter much for this one-function library.

seblj:
No problem! Thanks for merging!

nene:
Released the fix as 12.2.4

PR #654, opened 2023-10-31
fix #500, handle BETWEEN inside CASE expression

### LINKED ISSUE #500: Crash with BETWEEN inside CASE expression
**Describe the bug**

```
with
  p as (
    SELECT distinct
      s.id,
      case
        when s.end_date between (
          CASE
            WHEN CURRENT_DATE >= 12 THEN 13
          end
        ) and (
          CASE
            WHEN CURRENT_DATE >= 12 THEN 13
          end
        )  then year(current_date) + 1

      end as s_u,
    FROM
      table_name s
      )
```

**Actual behavior**

Crash:
```
token: {
    type: 'BETWEEN',
    raw: 'between',
    text: 'BETWEEN',
    start: 81,
    precedingWhitespace: ' '
  }
}
```

**Usage**
```
npx sql-formatter -l hive provoke.hql
```

Crash does not occur on 10.6.0
Exists in 11.0.2 and 12.0.0-beta.1

QoVoQ:
fix #500

### DIFF
--- a/src/parser/grammar.ne
+++ b/src/parser/grammar.ne
@@ -175,7 +175,6 @@ free_form_sql -> ( asteriskless_free_form_sql | asterisk ) {% unwrap %}
 asteriskless_free_form_sql ->
   ( asteriskless_andless_expression
   | logic_operator
-  | between_predicate
   | comma
   | comment
   | other_keyword ) {% unwrap %}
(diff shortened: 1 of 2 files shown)


QoVoQ:
Hi, @nene could you help to review this PR at your convenience ? thx.

nene (approved):
Thanks a lot. This looks great!

Fiddled around with this and it seems to work as expected.

I'll probably tweak this a bit (to avoid `property_access_prefix` being almost a duplicate of `asteriskless_andless_expression` but not quite).

Thanks again for tackling this long-standing bug.

PR #655, opened 2023-11-01
fix: Update E style escape strings for postgresql

cmanou:
As described in the [wiki](https://github.com/sql-formatter-org/sql-formatter/wiki/strings), E style PostgreSQL strings support both backslash and double single quote escaping. Also added the additional test for E strings into the test util.


nene (approved):
Thanks for spotting and fixing.

nene:
This fix is now released in 13.0.4

Issue #550, opened 2023-01-21
[VSCODE] SQLFlavourOverride: db2  ....  Gets error on "#" character, even though db2 allows for it


moymike:
I'm working with DB2 (iSeries) SQL statements.  One of the primary databases I work with includes "#" within the library name.

Example:
`select * from db#prod.pohead where ponum = 1111;`

Format Document error
`Unable to format SQL: Error: Parse error: Unexpected "#prod.pohead" at line 99 column 99`

If I use the IBM Client Access SQL tool, the format works fine.  It appears that the the "Prettier-SQL.SQLFlavourOverride": "db2" is not allowing character "#" in the library name even though it is valid.

nene:
Thanks for reporting.

Could you provide a link to DB2 documentation that specifies which characters are allowed in database names?

I've been following [this document](https://www.ibm.com/docs/en/db2/9.7?topic=elements-identifiers), which only mentions:

> An ordinary identifier is an uppercase letter followed by zero or more characters, each of which is an uppercase letter, a digit, or the underscore character.

Apparently that's not the full story, but I'm not familiar enough with DB2 to know from where to look for better sources.

moymike:
In DB2 the database tables are stored as LIBRARY.FILE.  (Lib.PF or Lib.LF)  The naming of libraries (and library items) allow for A-Z, @, #, , 0-9, underscores (_), and periods (.).

[IBM i library naming rules](https://jazz.net/sandbox01-clmhelp/index.jsp?topic=%2Fcom.ibm.team.scm.doc%2Ftopics%2Fr_RTCp_libraryparms.html)

Although these special characters are allowed, and a lot of systems contain them, they can be problematic.  This likely was not an issue years ago though, when many systems, such as ours was designed.

[Db2 object naming rules](https://www.ibm.com/docs/en/db2/11.5?topic=rules-db2-object-naming)

I tested another VScode extension "Sql Formatter" which has a "db2" option that works with the "#" library name.

If you update the extension please let me know and I'll try it out.  Your extension contains more formatting options.

nene:
Thanks for the references. I'll see what I can do to fix this.

> I tested another VScode extension "Sql Formatter" which has a "db2" option that works with the "#" library name.

There seem to be many VSCode extensions named "SQL Formatter", all of them being wrappers around this same library, though often using an outdated version. The older versions of this library were more willing to treat all kinds of text as valid SQL.

nene:
[...]

Issue #203, opened 2022-04-26
Language DB2 format with $ failed


derPeterson:
I have specified db2 language and have a select, the values with $ characters, are separated by whitespaces, which leads to SQL error....

`SELECT * FROM TABLENAME AS C WHERE C.CLASSID IN $P{CLASSID} AND C.OID IN $P{OID} $P{INCLUDE}`

Format Result:

`SELECT * FROM TABLENAME AS C WHERE C.CLASSID IN $ P { CLASSID } AND C.OID IN $ P { OID } $ P { INCLUDE }`

I'm not a DB2 expert, but my DBeaver tool formats the $ correctly and leaves them as they were....

Thanks already for the help and for the free library, otherwise it works great.

nene:
This should new be fixed in 6.0 release.

Issue #123, opened 2021-02-21
white space inserted for language db2


huineng:
i have specified language db2 and have a select that needs to replace some hex characters

```sql
replace(AL3.NAME1, x'00', x'20') AS CUSTNAME
```

but get's replace by 

```sql
replace(AL3.NAME1, x '00', x '20') AS CUSTNAME
```

This gives an sql error

```
com.ibm.db2.jcc.am.SqlSyntaxErrorException: ILLEGAL SYMBOL "'00'". 
SOME SYMBOLS THAT MIGHT BE LEGAL ARE: AS * AT OPTION YEAR YEARS 
MONTH MONTHS DAY DAYS HOUR HOURS. SQLCODE=-104, SQLSTATE=42601, DRIVER=4.28.11
```

Let me know how this problem can be solved.. maybe can be added to a list of exceptions

[some doc on hex](https://www.ibm.com/support/knowledgecenter/SSEPEK_10.0.0/sqlref/src/tpc/db2z_charstringconstants.html)

huineng:
I have mad a pull request after testing it out on my code https://github.com/zeroturnaround/sql-formatter/pull/127

inferrinizzard:
Hi, this issue is fixed in my fork here:
https://github.com/inferrinizzard/prettier-sql
https://www.npmjs.com/package/prettier-sql

nene:
Now fixed for both lowecase and uppercase `X`.

PR #658, opened 2023-11-08
Add SQL dialect for DB2 for IBM i

chrjorgensen:
Hello.

We use your great SQL formatter utility in our [Db2 for IBM i tools](https://github.com/codefori/vscode-db2i) extension, but we find that the DB2 dialect in the formatter is for DB2 for Z (mainframe), whereas our extension is for IBM i on IBM Power server. There is quite some differences between the two platforms, which causes some problems in the SQL formatter.

So we offer this PR to add a different DB2 dialect targeted for IBM i - so we and other users of your extension can select this platform instead of IBM Z.

I'm a newbie to your code, but I have tried to copy everything Db2 related into new files for Db2i. Please verify carefully and let me know, if I missed something.

Thanks.


chrjorgensen:
It seems like the tests fails when run for the new Db2i dialect - what have I missed?

image

etc...

src/languages/db2i/db2i.formatter.ts
```diff
@@ -0,0 +1,193 @@
...
+  'OPEN',
+  'PREPARE',
+  'REFRESH TABLE',
+  'RELEASE',
+  'RELEASE SAVEPOINT',
+  'RENAME',
+  'REVOKE',
+  'ROLLBACK',
+  'SAVEPOINT',
+  'SELECT',
```
nene, line 128: The problem is that `SELECT` is listed both here in `onelineClauses` and also above in `selectClause`.

The entry in here happens to have higher precedence and the result is that column names are listed one the same line as `SELECT` keyword.

src/languages/db2i/db2i.formatter.ts
```diff
@@ -0,0 +1,193 @@
...
+    reservedFunctionNames: functions,
+    stringTypes: [
+      { quote: "''-qq", prefixes: ['G', 'N', 'U&'] },
+      { quote: "''-raw", prefixes: ['X', 'BX', 'GX', 'UX'], requirePrefix: true },
+    ],
+    identTypes: [`""-qq`],
+    identChars: { first: '@#$' },
+    paramTypes: { positional: true, named: [':'] },
+    paramChars: { first: '@#$', rest: '@#$' },
+    operators: ['**', '¬=', '¬>', '¬<', '!>', '!<', '||', 'CONCAT'],
```
nene, line 188: Although `CONCAT` is an operator it shouldn't be listed here. Only operators consisting of special characters should be listed here. `CONCAT` should only appear in the general keywords list.

src/languages/db2i/db2i.formatter.ts
```diff
@@ -0,0 +1,193 @@
...
+  'SET {GENERATED ALWAYS | GENERATED BY DEFAULT}', // for alter column
+  'SET NOT NULL', // for alter column
+  'SET {NOT HIDDEN | IMPLICITLY HIDDEN}', // for alter column
+  'SET FIELDPROC', // for alter column
+  'DROP {DEFAULT | NOT NULL | GENERATED | IDENTITY | ROW CHANGE TIMESTAMP | FIELDPROC}', // for alter column
+  // - truncate:
+  'TRUNCATE [TABLE]',
+  // other
+  'SET [CURRENT] SCHEMA',
+  // https://www.ibm.com/docs/en/i/7.5?topic=reference-statements
```
nene, line 65: Lots of entries listed below are likely incorrect. They're also incorrect in the original `DB2` implementation. This list should only contain phrases that can begin a statement or clause.

For example the phrase `SELECT INTO` can not appear in valid SQL, instead there can be something like `SELECT col1, col2 FROM tbl1 INTO tbl2`. So probably better to add an entry for `INTO`, so the `INTO`-part gets formatted on a separate line.

nene (changes requested):
Thanks for the PR. Pointed out a few issues that popped up in my first quick glance.

You should keep in mind that the original DB2 implementation is likely seriously problematic. I have no idea how many people actually use it, if any. It was introduced during some very early days of SQL Formatter development and has remained largely unchanged. I Personally have never used any version of DB2.

nene:
I'm unsure whether having two DB2 dialects is worth it. The fact that the existing dialect targets the mainframe version is most likely just the result of googling DB2 syntax and picking the first result which happened to be the mainframe version.

I would guess that there are less people interested in using this tool with the mainframe version... but really I have no idea.

chrjorgensen:
Thanks for showing me where I made mistakes, I will fix them asap.

I didn't know about the state of the existing DB2 dialect, assumed it was okay and copied it. I'm also not familiar how the formatter works and the rules it applies, but I learn along the way.

FYI, there are three DB2 versions available: DB2 for Z (mainframe), DB2 for IBM i and DB2 for LUW (Linux, Unix, Windows). They are not the same (even though they share some codebase) and each have special functions and keywords...

src/languages/db2i/db2i.formatter.ts
chrjorgensen: Ah, this I forgot to remove from the list, when adding the statements. Fixed...

src/languages/db2i/db2i.formatter.ts
chrjorgensen: Thanks for explaining - fixed...

src/languages/db2i/db2i.formatter.ts
chrjorgensen: Ah, thank you for your explanation and suggestion. I will go through these and change them as required.

chrjorgensen: Issues should be fixed now - all tests run without errors.

nene:
Looking through the few past issues that relate to DB2 I see the following:

- #123 references the Z-version
- #203 seems to be about a problem with using a templating language, not really related to DB2. At least I guess so. It mentions DBeaver tool, which supports all variants of DB2.
- #550 references the i-version

The current implementation of DB2 is really a mess. I have taken some stuff (like list of keywords) from the Z-version and some other stuff (like identifier syntax) from the LUW-version.

I'm currently inclining towards throwing the existing db2 implementation away and just keeping the i-version. At least I know that somebody is actively using it, and the differences between these DB2 versions for now are really minimal.

Just FYI. This doesn't really effect this PR.

test/features/join.ts
```diff
@@ -42,43 +42,43 @@ export default function supportsJoin(
         const result = format(`
           SELECT * FROM customers
           ${join} orders ON customers.customer_id = orders.customer_id
-          ${join} items ON items.id = orders.id;
+          ${join} items ON items.col1 = orders.col1;
```
nene, line 45: Is it really the case that unquoted `id` can't be used as a column name in DB2 i?

When I look at the reserved word lists of [LUW-version](https://www.ibm.com/docs/en/db2/11.5?topic=sql-reserved-schema-names-reserved-words) and [Z-version](https://www.ibm.com/docs/en/db2-for-zos/11?topic=words-reserved#db2z_reservedwords__newresword) the `ID` indeed doesn't appear in reserved words list.

However I still find it surprising that a) one would need to quote such a commonly used column name b) there's such a striking difference between the i-version of DB2 from other DB2 dialects.

chrjorgensen:
Thanks for looking into the DB2 "mess"! 😃 

I've made some updates to the DB2i formatter - and it raised some questions in my head:
- I had to remove `ANY` from the functions, since it collided with the `ANY` set operator. Any (no pun intended) way to distinguish between the two?
- Where are the rules for identifier name specified? Is it the `ident`specifications in `<dialect>.formatter.ts`?
- What is the priority used by the formatter? Order of specifications?

I'm curious and want to learn how the formatter works...

test/features/join.ts
chrjorgensen: The [reserved words list](https://www.ibm.com/docs/en/i/7.5?topic=words-reserved) for DB2 for IBM i shows `ID` as a reserved word - but I haven't tested if `ID` is allowed in the context of a column name.

chrjorgensen: I stumbled into the `ID` problem when I activated `JOIN ... USING` - suddenly the tests failed for DB2i because `ID` was converted to uppercase by the formatter. So I changed the `id` in the test to a more generic column name.

Just tested `id` ad a column name on DB2 for i - and it is indeed allowed... Is it better to remove `ID` from the reserved words list?

nene: Could you please test this?

I remember seeing `ID` in the keywords list of some other SQL dialects, but it has always turned out that it's not really a reserved keyword.

nene:
> I had to remove `ANY` from the functions, since it collided with the `ANY` set operator. Any (no pun intended) way to distinguish between the two?

Not really. I guess in both cases `ANY` will be followed by `(`, which makes distinguishing the two extra difficult.

> Where are the rules for identifier name specified? Is it the `ident`specifications in `<dialect>.formatter.ts`?

Yes, that's correct. The `identChars` option defines additional characters for identifier names. The regex for matching identifiers is constructed here: https://github.com/sql-formatter-org/sql-formatter/blob/master/src/lexer/regexFactory.ts#L162C22-L162C22

> What is the priority used by the formatter? Order of specifications?

The [Tokenizer](https://github.com/sql-formatter-org/sql-formatter/blob/master/src/lexer/Tokenizer.ts) class defines the order of matching different types of tokens. Pretty much reading the rules listed in this class from top-to-bottom... with higher priority items at the top. From there one can see that `reservedClauses` is matched before `reservedSelect`.

But really the main thing is not to repeat the same pattern in multiple places. One notable special case is `reservedPhrases` which sole purpose is to have higher priority than `reservedClauses`, which allows overriding patterns that would otherwise be detected and formatted as separate clauses, like overriding `ON UPDATE` so it's not confused with the `UPDATE` statement.

test/features/join.ts
nene: Yep. Best to just remove it from reserved words list. Given that this list has currently like 500 words, my suspicion is that many more of these keywords are actually not reserved.

nene: I encountered a similar problem when trying to find out which keywords does SingleStoreDB actually reserve. Their documentation only listed all the keywords and didn't distinguish which ones are reserved and which ones were not. I ended up executing hundreds of SQL queries to see which would actually result in a failure.

nene:
I made small tweaks and merged this in.

nene:
This is now released in 13.1.0.

chrjorgensen:
Thank you! And thank you for explaining about the formatter and how it works. 😃

PR #661, opened 2023-11-11
Fix db2 options

chrjorgensen:
A fix to DB2 formatter:
- ALTER TABLE RENAME COLUMN is not supported in DB2i


src/languages/db2/db2.formatter.ts
```diff
@@ -185,7 +185,7 @@ export const db2: DialectOptions = {
     reservedKeywords: keywords,
     reservedFunctionNames: functions,
     stringTypes: [
-      { quote: "''-qq", prefixes: ['G', 'N', 'U&'] },
+      { quote: "''-qq", prefixes: ['G', 'N', 'U'] },
```
nene, line 188: Is this change of intentional?

nene:
Sorry, in the mean time I had already fixed the `U&''` strings difference between DB2 and DB2i. Just had forgotten to push my changes.

src/languages/db2/db2.formatter.ts
nene: That is, you seem to be suggesting that `U&''` strings are not supported in DB2 LUW-variant, but instead `U''` strings are supported.

nene:
Also, if you're further interested in improving DB2 support, you might want to review the [wiki pages](https://github.com/sql-formatter-org/sql-formatter/wiki) which describe the basic syntax of various dialects.

I updated the first section (identifiers, strings, etc) there with references to DB2 LUW-version and DB2i. With the LOW-version of DB2 I can test out the syntax in [dbfiddle](https://dbfiddle.uk/). But for DB2i I have no other sources besides the IBM documentation. The "Queries" sections is also mostly updated, the rest is still in todo status.

src/languages/db2/db2.formatter.ts
chrjorgensen: Sorry, a mistake from my part - my rebase didn't go as expected. I'll remove this change, since you've already fixed the `U&''` issue.

chrjorgensen:
> Sorry, in the mean time I had already fixed the U&'' strings difference between DB2 and DB2i. Just had forgotten to push my changes.

That's alright - it's better that we are two thinking the same than none of us is thinking about this. 👍 
I'll remove the change and force push the `RENAME´ change.

nene:
Thanks.

chrjorgensen:
Thank you for doing all this great work on the DB2 dialects!

test/db2i.test.ts
```diff
@@ -16,6 +17,11 @@ describe('Db2iFormatter', () => {
 
   supportsComments(format, { nestedBlockComments: true });
   supportsCreateTable(format, { orReplace: true });
+  supportsAlterTable(format, {
+    addColumn: true,
+    dropColumn: true,
+    renameColumn: false,
```
nene, line 23: FYI: In general instead of specifying `false` here, better to just skip the option altogether. At least that's the style that the rest of the tests are following for now. Keeps the tests shorter, as like `supportsAlterTable()` takes an entire 5 boolean flags.

This concrete case I already cleaned up: https://github.com/sql-formatter-org/sql-formatter/commit/91e4be051c25df5e2cb29433e70495146b390b24

chrjorgensen: Noted - only did it this way to clarify that it was on purpose in contrast to the DB2, which has `true` in `renameColumn`.

PR #663, opened 2023-11-11
Add option for specifying case for identifiers

chrjorgensen:
Add an option `identifierCase`to specify what case the identifiers should be converted to - like the keyword case. Possible options are `preserver`, `upper` and `lower`.

### DIFF
--- a/docs/identifierCase.md
+++ b/docs/identifierCase.md
@@ -31,11 +31,11 @@ group by Column4
 
 ```
 select
-  count(a.COLUMN1),
-  max(a.COLUMN2 + a.COLUMN3),
-  a.COLUMN4 AS MYCOL
+  count(A.COLUMN1),
+  max(A.COLUMN2 + A.COLUMN3),
+  A.COLUMN4 AS MYCOL
 from
-  TABLE1 as a
+  TABLE1 as A
 where
   COLUMN6
   and COLUMN7
--- a/docs/identifierCase.md
+++ b/docs/identifierCase.md
@@ -1,0 +1,54 @@
+# identifierCase
+
+Converts identifiers to upper or lowercase.
+
+## Options
+
+- `"preserve"` (default) preserves the original case.
+- `"upper"` converts to uppercase.
+- `"lower"` converts to lowercase.
+
+### preserve
+
+```
+select
+  count(a.Column1),
+  max(a.Column2 + a.Column3),
+  a.Column4 AS myCol
+from
+  Table1 as a
+where
+  Column6
+  and Column7
+group by Column4
+```
+
+### upper
+
+```
+select
+  count(a.COLUMN1),
+  max(a.COLUMN2 + a.COLUMN3),
+  a.COLUMN4 AS MYCOL
+from
+  TABLE1 as a
+where
+  COLUMN6
+  and COLUMN7
+group by COLUMN4
+```
+
+### lower
+
+```
+select
+  count(a.column1),
+  max(a.column2 + a.column3),
+  a.column4 AS mycol
+from
+  table1 as a
+where
+  column6
+  and column7
+group by column4
+```
--- a/docs/identifierCase.md
+++ b/docs/identifierCase.md
@@ -2,6 +2,10 @@
 
 Converts identifiers to upper or lowercase.
 
+Note: An identifier is a name of a SQL object.
+There are two types of SQL identifiers: ordinary identifiers and quoted identifiers.
+Only ordinary identifiers are subject to be converted.
+
 ## Options
 
 - `"preserve"` (default) preserves the original case.
--- a/src/formatter/ExpressionFormatter.ts
+++ b/src/formatter/ExpressionFormatter.ts
@@ -286,7 +286,7 @@ export default class ExpressionFormatter {
   }
 
   private formatIdentifier(node: IdentifierNode) {
-    this.layout.add(node.text, WS.SPACE);
+    this.layout.add(this.showIdentifier(node), WS.SPACE);
   }
 
   private formatParameter(node: ParameterNode) {
@@ -504,6 +504,21 @@ export default class ExpressionFormatter {
         return node.text;
       case 'lower':
         return node.text.toLowerCase();
+    }
+  }
+
+  private showIdentifier(node: IdentifierNode): string {
+    if (/['"\\`]/.test(node.text[0]) || node.text.startsWith(`U&`)) {
+      return node.text;
+    } else {
+      switch (this.cfg.identifierCase) {
+        case 'preserve':
+          return node.text;
+        case 'upper':
+          return node.text.toUpperCase();
+        case 'lower':
+          return node.text.toLowerCase();
+      }
     }
   }
 }
--- a/src/formatter/ExpressionFormatter.ts
+++ b/src/formatter/ExpressionFormatter.ts
@@ -145,7 +145,9 @@ export default class ExpressionFormatter {
   private formatArraySubscript(node: ArraySubscriptNode) {
     this.withComments(node.array, () => {
       this.layout.add(
-        node.array.type === NodeType.keyword ? this.showKw(node.array) : node.array.text
+        node.array.type === NodeType.keyword
+          ? this.showKw(node.array)
+          : this.showIdentifier(node.array)
       );
     });
     this.formatNode(node.parenthesis);
@@ -508,9 +510,7 @@ export default class ExpressionFormatter {
   }
 
   private showIdentifier(node: IdentifierNode): string {
-    if (!(node.tokenType === TokenType.IDENTIFIER)) {
-      return node.text;
-    } else {
+    if (node.tokenType === TokenType.IDENTIFIER || node.tokenType === TokenType.ARRAY_IDENTIFIER) {
       switch (this.cfg.identifierCase) {
         case 'preserve':
           return node.text;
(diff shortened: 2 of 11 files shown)


src/formatter/ExpressionFormatter.ts
```diff
@@ -286,7 +286,7 @@ export default class ExpressionFormatter {
   }
 
   private formatIdentifier(node: IdentifierNode) {
-    this.layout.add(node.text, WS.SPACE);
+    this.layout.add(this.showIdentifier(node), WS.SPACE);
```
nene, line 289: In addition to plain identifiers we also have several identifier-like things:

- variables
- parameters

these usually behave just like identifiers and the only thing distinguishing them from normal identifiers is some prefix like `@myvar` or `:myparam`.

As the parser currently treats variables as identifiers, these too end up upper/lowercased. But parameters are kepts separate by the parser, so the case of these doesn't change. Should parameters also change together with identifiers?

test/options/identifierCase.ts
```diff
@@ -0,0 +1,71 @@
+import dedent from 'dedent-js';
+
+import { FormatFn } from '../../src/sqlFormatter.js';
+
+export default function supportsIdentifierCase(format: FormatFn) {
```
nene, line 5: There are likely more tests needed here to cover:

- that all identifiers in expression like `foo.bar.baz` get uppercased,
- variables
- parameters
- array names (treated differently by lexer)
- various types of identifier quoting styles

As most of these things depend on the dialect, it might be better to just add additional tests to places like `supportsIdentifiers()`.

test/options/identifierCase.ts
```diff
@@ -0,0 +1,71 @@
...
+      from
+        tbl1
+        left join tbl2
+      where
+        cola > 1
+        and colb = 3
+    `);
+  });
+
+  it('does not uppercase identifiers inside strings', () => {
```
nene, line 52: nit: I wouldn't really call it an identifier inside a string - it's just a string.

I personally wouldn't add a separate test for this, I'd just include some strings inside the general "converts identifiers to lowercase/uppercase" test cases. But that's really a personal preference. Doesn't matter much either way.

docs/identifierCase.md
```diff
@@ -0,0 +1,54 @@
+# identifierCase
+
+Converts identifiers to upper or lowercase.
```
nene, line 3: This might be a good place to specify exactly what classifies as an identifier and what doesn't.

nene (changes requested):
It's a good start, but needs some more work and thought.

docs/identifierCase.md
```diff
@@ -0,0 +1,54 @@
...
+select
+  count(a.COLUMN1),
+  max(a.COLUMN2 + a.COLUMN3),
+  a.COLUMN4 AS MYCOL
+from
+  TABLE1 as a
+where
+  COLUMN6
+  and COLUMN7
+group by COLUMN4
```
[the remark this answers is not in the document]
chrjorgensen: Fixed - didn't check whether a correlation name was an identifier or not...

docs/identifierCase.md
chrjorgensen: Okay with this text?

src/formatter/ExpressionFormatter.ts
```diff
@@ -506,4 +506,19 @@ export default class ExpressionFormatter {
         return node.text.toLowerCase();
     }
   }
+
+  private showIdentifier(node: IdentifierNode): string {
+    if (/['"\\`]/.test(node.text[0]) || node.text.startsWith(`U&`)) {
```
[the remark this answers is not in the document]
chrjorgensen: Agree - I had some trouble identifying quoted identifiers.
Your mention of IDENTIFIER and QUOTED_IDENTIFIER from the lexer was the key!

test/options/identifierCase.ts
chrjorgensen: String test is now included in identifier case tests and removed as separate test.

test/options/identifierCase.ts
chrjorgensen: My change is only for unquoted identifiers - I think the other tokens types should not be included in this change.

docs/identifierCase.md
chrjorgensen: I added a note in commit [afe5b48](https://github.com/sql-formatter-org/sql-formatter/pull/663/commits/afe5b48b281957e5904b9d6e271d67c1f272e595) - is it okay?

docs/identifierCase.md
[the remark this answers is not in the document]
chrjorgensen: Fixed in commit [0c74c02](https://github.com/sql-formatter-org/sql-formatter/pull/663/commits/0c74c02f5abc81842c5e054d260640c87c4e0ef0)

src/formatter/ExpressionFormatter.ts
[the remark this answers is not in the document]
chrjorgensen: I followed your suggestion and changed the code to store the token type in the node for identifiers - and now only nodes with token type `IDENTIFIER` will be converted.

src/formatter/ExpressionFormatter.ts
chrjorgensen: My change is only for unquoted identifiers - I think variables and parameters are outside the scope and should be handled in another PR...

test/options/identifierCase.ts
chrjorgensen: I added a test for multi-part identifiers in commit [558efa3](https://github.com/sql-formatter-org/sql-formatter/pull/663/commits/558efa36bfa434fdb672408a2fed5ccaaa4d0359)

test/options/identifierCase.ts
chrjorgensen: String test is now included in identifier case tests and removed as separate test.

src/parser/grammar.ne
```diff
@@ -202,7 +208,7 @@ atomic_expression ->
 array_subscript -> %ARRAY_IDENTIFIER _ square_brackets {%
   ([arrayToken, _, brackets]) => ({
     type: NodeType.array_subscript,
-    array: addComments({ type: NodeType.identifier, text: arrayToken.text}, { trailing: _ }),
+    array: addComments({ type: NodeType.identifier, tokenType: TokenType.ARRAY_IDENTIFIER, text: arrayToken.text}, { trailing: _ }),
```
nene, line 211: So, currently `ARRAY_IDENTIFIER` tokens are treated differently from normal identifiers. This results in the following SQL:

```sql
select foo, foo[1] from tbl
```

being formatted as:

```sql
select
  FOO,
  foo[1]
from
  TBL
```

That is, when the `foo` column is used normally, it gets uppercased, but when used together with array-accessor operator it is not uppercased.

This difference between normal identifiers and array-identifiers is really just an internal quirk of SQL Formatter implementation. We should treat both as identifiers and change the case of them.

docs/identifierCase.md
```diff
@@ -0,0 +1,58 @@
+# identifierCase
+
+Converts identifiers to upper or lowercase.
+
+Note: An identifier is a name of a SQL object.
+There are two types of SQL identifiers: ordinary identifiers and quoted identifiers.
+Only ordinary identifiers are subject to be converted.
```
nene, line 7: This is better, but according to this description I would expect the case of variables and parameters to also be converted, because they too are names of SQL objects.

In general I see all these as different kinds of identifiers:

- schema, table and column names
- variable names
- function names
- parameter names

Variables are a particularly tricky case. Some dialects like MySQL have variables with a prefix like `@foo`, and therefore we can easily distinguish them. Other dialects like PostgreSQL also have variables, but there is no special prefix, an idientifier `foo` could refer to a variable, or it could refer to a table or column.

nene (changes requested):
This is much better now.

Should definitely fix the handling of `array[index]`.

Dealing with the casing of variables and parameters is I think out of scope for this PR. While I believe that this `identifierCase` option should also cover these, a further work is needed in the lexer and parser to make it possible to distinguish between quoted and unquoted variants of both.

I'll also have to think how should this feature interact with another feature I'd like to have: specifying the casing of functions. Function names are also identifiers. But I'd like to be able to say `functionCase: "upper", identifierCase: "lower"`.

docs/identifierCase.md
chrjorgensen: Ah, I think I mixed variables and parameters... by variables, do you mean SQL variables defined by `CREATE VARIABLE` (in DB2)? If yes, then I agree that they should also be considered an identifier and converted.

nene: Yes, variables like that.

src/parser/grammar.ne
chrjorgensen: Array identifiers are now being converted together with ordinary identifiers - see commit [0bb7d21](https://github.com/sql-formatter-org/sql-formatter/pull/663/commits/0bb7d2137c5073a217e30f82c3fb120ce9df678b)

nene (approved):
Thanks!

nene:
@karlhorky there is no before or after here. The formatter does not do multiple passes of applying the format. It just detects some words as keywords and some as identifiers. And then these get converted to upper/lower case if configured so. It will never be the case that a word is considered both identifier and keyword by the formatter.

karlhorky:
Ok understood, thanks for the clarification.

PR #670, opened 2023-11-22
Bundle with tsup

### LINKED ISSUE #604: Package cannot be imported for certain TypeScript configurations
**Describe the bug**

I apologize in advance for filing this; ESM/CJS/TypeScript compatibility is a horrific can of worms.

Importing this package in a TypeScript project that uses `"module": "Node16"` fails when the project itself is not configured to use ESM (that is, `package.json` does not contain `"type": "module"`).

I've created a reproduction here: https://github.com/nwalters512/sql-formatter-typescript-repro. Instructions are in the readme.

**Expected behavior**

Given that `sql-formatter` provides both ESM and CJS implementations, I would expect to be able to import the CJS version in this situation.

**Actual behavior**

TypeScript produces the following error:

```
src/index.ts:1:24 - error TS1479: The current file is a CommonJS module whose imports will produce 'require' calls; however, the referenced file is an ECMAScript module and cannot be imported with 'require'. Consider writing a dynamic 'import("sql-formatter")' call instead.
  To convert this file to an ECMAScript module, change its file extension to '.mts', or add the field `"type": "module"` to '/Users/nathan/git/sql-formatter-typescript-repro/package.json'.

1 import { format } from 'sql-formatter';
                         ~~~~~~~~~~~~~~~


Found 1 error in src/index.ts:1
```

Note that things still work correctly at runtime; the presence of the `types` field combined with `"type": "module"` in this package confused TypeScript enough that it doesn't realize that there is a completely valid CJS implementation. I considered filing an issue upstream on TypeScript itself, but this was already discussed on https://github.com/microsoft/TypeScript/issues/54263, which was closed without any indication that TypeScript itself would fix this (even though I think they should).

I believe the way to fix this would be to drop the `"types"` field from `"exports"` and instead place `*.d.ts`/`*.d.cts` files adjacent to the files that are referenced by the `"import"` and `"require"` exports in `package.json`. This way, TypeScript will automatically discover the correct types based on the particular conditions that it's using to resolve the implementation files.

nwalters512:
This PR aims to resolve #604. It uses [`tsup`](https://github.com/egoist/tsup) to bundle code and emit both CJS and ESM (and dedicated types for each). While I don't love bundling code, `tsup` makes it really straightforward to get something that "just works".

An alternative we could use `tsc` directly to emit ESM/CJS builds to `dist/esm` and `dist/cjs`, respectively, then make a second pass with a custom script to rename `dist/cjs/*.js` to `dist/cjs/*.cjs`. Unfortunately `tsc` can't do this natively: https://github.com/microsoft/TypeScript/issues/49462

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,16 +3,14 @@
   "version": "14.0.0",
   "description": "Format whitespace in a SQL query to make it more readable",
   "license": "MIT",
-  "main": "dist/sql-formatter.min.cjs",
+  "main": "dist/index.cjs",
+  "module": "dist/index.js",
   "unpkg": "dist/sql-formatter.min.js",
-  "module": "lib/index.js",
-  "types": "lib/src/index.d.ts",
   "exports": {
     "./package.json": "./package.json",
     ".": {
-      "types": "./lib/src/index.d.ts",
-      "import": "./lib/index.js",
-      "require": "./dist/sql-formatter.min.cjs"
+      "import": "./dist/index.js",
+      "require": "./dist/index.cjs"
     }
   },
   "bin": {
@@ -64,10 +62,9 @@
     "prepare": "yarn clean && yarn grammar && yarn fix && yarn check && yarn build",
     "pre-commit": "npm-run-all --parallel ts:changes lint:changes",
     "grammar": "nearleyc src/parser/grammar.ne -o src/parser/grammar.ts",
-    "build:babel": "babel src --out-dir lib --extensions .ts --source-maps",
-    "build:types": "tsc --emitDeclarationOnly --isolatedModules",
-    "build:minified": "webpack --config webpack.prod.js && cp dist/sql-formatter.min.cjs dist/sql-formatter.min.js",
-    "build": "yarn grammar && npm-run-all --parallel build:babel build:types build:minified",
+    "build:tsup": "tsup src/index.ts --format cjs,esm --sourcemap --dts",
+    "build:webpack": "webpack --config webpack.prod.js && cp dist/sql-formatter.min.cjs dist/sql-formatter.min.js",
+    "build": "yarn grammar && npm-run-all --parallel build:tsup build:webpack",
     "release": "release-it"
   },
   "repository": {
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -3,16 +3,14 @@
   "version": "14.0.0",
   "description": "Format whitespace in a SQL query to make it more readable",
   "license": "MIT",
-  "main": "dist/sql-formatter.min.cjs",
+  "main": "dist/index.cjs",
+  "module": "dist/index.js",
   "unpkg": "dist/sql-formatter.min.js",
-  "module": "lib/index.js",
-  "types": "lib/src/index.d.ts",
```
nwalters512, line 9: Removing the `types` field both here and in `exports` lets TypeScript resolve the correct types by filename.

package.json
```diff
@@ -3,16 +3,14 @@
   "version": "14.0.0",
   "description": "Format whitespace in a SQL query to make it more readable",
   "license": "MIT",
-  "main": "dist/sql-formatter.min.cjs",
+  "main": "dist/index.cjs",
```
nwalters512, line 6: This now uses the un-minified, non-UMD bundle. I think this is ok; `unpkg` will continue using the minified+UMD bundle produced by webpack.

src/index.ts
```diff
@@ -2,9 +2,24 @@ export { supportedDialects, format, formatDialect } from './sqlFormatter.js';
 export { expandPhrases } from './expandPhrases.js';
 export { ConfigError } from './validateConfig.js';
 
-// Intentionally use "export *" syntax here to make sure when adding a new SQL dialect
-// we wouldn't forget to expose it in our public API.
-export * from './allDialects.js';
+// When adding a new dialect, be sure to add it to the list of exports below.
```
nwalters512, line 5: Removing the `export *` was necessary because esbuild transformed it into code that wasn't statically analyzable for tree shaking. If desired, I can look into adding a test that asserts that this file has all the necessary language imports?

nwalters512:
@nene it looks like the latest version of `tsup` is expecting Node 18+. However, it looks like Node is pinned at v14 in CI. Are you willing to use a newer version of Node, or should I look into other solutions?

nene:
No other reason of using Node 14 besides me not having bothered to witch the CI to a newer version.

PR #676, opened 2023-11-29
Switch from `flatKeywordList()` to flat arrays

karlhorky:
Part of preparations for PR #673

Copied from [my original comment](https://github.com/sql-formatter-org/sql-formatter/pull/673#issuecomment-1830518684):

I see there is no distinction in the keyword tokens between `all` and `dataType` / `datatype` / `dataTypes`:

https://github.com/sql-formatter-org/sql-formatter/blob/290f3b767492d5e3204866447f9f4e4c098f825f/src/languages/postgresql/postgresql.keywords.ts#L4-L8

https://github.com/sql-formatter-org/sql-formatter/blob/290f3b767492d5e3204866447f9f4e4c098f825f/src/languages/transactsql/transactsql.functions.ts#L78-L83

https://github.com/sql-formatter-org/sql-formatter/blob/290f3b767492d5e3204866447f9f4e4c098f825f/src/languages/snowflake/snowflake.keywords.ts#L102-L106

https://github.com/sql-formatter-org/sql-formatter/blob/290f3b767492d5e3204866447f9f4e4c098f825f/src/languages/transactsql/transactsql.functions.ts#L303-L309

This is because of the `flatKeywordList()` function, which effectively discards the key information:

https://github.com/sql-formatter-org/sql-formatter/blob/290f3b767492d5e3204866447f9f4e4c098f825f/src/utils.ts#L18-L20

Because of this, I would like to simplify the `src/languages/*/*.keywords.ts` files to export 1 flat array (with empty lines + comments as separators instead of object keys) and delete the `flatKeywordList()` function - this will prevent unusual, inconsistent naming conventions and also clarify that the object keys do nothing.

PR #677, opened 2023-11-29
Separate data types from generic keywords

karlhorky:
Superset of https://github.com/sql-formatter-org/sql-formatter/pull/676/ (first merge #676)


karlhorky:
> I notice that data types are missing from db2 and singlestoredb dialects.

1. Yeah originally I was going to do only a few dialects, but I ended up doing more. Probably worth it to do all of them now.

> I'm not sure whether keywords like DAY, MONTH, HOUR should be considered data types

2. I was also unsure here. It was mentioned in multiple docs pages, so I decided to go for it. But I don't feel strongly about this, so I'll revert this.

> The ARRAY keyword is trickier. It can be used in both data type declaration, but also when writing an `ARRAY[1, 2, 3]` literal. I think we already have some special handling of arrays, but my memory can also be wrong.

3. Right, since `ARRAY` is used in those positions and is more commonly considered a data type (across docs and also programming languages generally), I would suggest keeping `ARRAY` as is, if possible. If it's important to revert it, I can do this too.

I'll get started on 1 and 2

karlhorky:
1. Data types for `db2`, `singlestore` and `transactsql` have also be split out of the keywords  971fcd36
2. `SECOND`, `MINUTE`, `HOUR`, `DAY`, `WEEK`, `MONTH`, `YEAR` keywords are moved back to `keywords` 03f6f9a7
3. I'll wait for your availability + decision on `ARRAY`

karlhorky:
Thanks for the review and merge! I'll rebase #673

PR #689, opened 2023-12-06
Add dataTypeCase & functionCase

nene:
Continuation of PR #673


nene:
I'm still not sure how to best handle the `ARRAY` keyword/type.

In many dialects array literals are written as `ARRAY[1, 2, 3]` - here `ARRAY` is not a type, it's part of the literal syntax.

In BigQuery one can write just `[1, 2, 3]`, but also `ARRAY[1, 2, 3]` and `ARRAY<INT64>[1, 2, 3]` - here `ARRAY` is a type.

At the moment in this PR the `ARRAY[]` is formatted as keyword, but `ARRAY<INT64>[]` as data type. Which isn't ideal.

Will need to do more research into how the `ARRAY` keyword is used in other dialects.

PR #692, opened 2023-12-09
Add IDENTIFIED in mysql keywords

### LINKED ISSUE #691: [FORMATTING] `IDENTIFIED` recognized as identifier
**Input data**

```sql
create user a1@'%' identified by 'passwd';
```

**Expected Output**

```sql
CREATE USER a1 @'%' IDENTIFIED BY 'passwd';
```

**Actual Output**

```sql
CREATE USER a1 @'%' identified BY 'passwd';
```

**Usage**

- How are you calling / using the library? `sql-formatter --config config.sql file.sql`

```json
{
  "language": "mysql",
  "keywordCase": "upper",
  "dataTypeCase": "upper",
  "functionCase": "upper"
}
```

If I add `"identifierCase": "upper"` in config file, the result would be

```sql
CREATE USER A1 @'%' IDENTIFIED BY 'passwd';
```

I think `identified` should be a keyword.

- What SQL language(s) does this apply to? `mysql`
- Which SQL Formatter version are you using? `15.0.0`

eeeXun:
Fix #691

### DIFF
--- a/src/languages/mariadb/mariadb.formatter.ts
+++ b/src/languages/mariadb/mariadb.formatter.ts
@@ -261,6 +261,7 @@ const reservedPhrases = expandPhrases([
   'ON {UPDATE | DELETE} [SET NULL | SET DEFAULT]',
   'CHARACTER SET',
   '{ROWS | RANGE} BETWEEN',
+  'IDENTIFIED BY',
 ]);
 
 // For reference: https://mariadb.com/kb/en/sql-statements-structure/
--- a/src/languages/mariadb/mariadb.keywords.ts
+++ b/src/languages/mariadb/mariadb.keywords.ts
@@ -76,7 +76,6 @@ export const keywords: string[] = [
   'HOUR_MICROSECOND',
   'HOUR_MINUTE',
   'HOUR_SECOND',
-  'IDENTIFIED',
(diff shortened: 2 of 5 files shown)


eeeXun:
I see `IDENTIFIED` is in [mysql keywords page](https://dev.mysql.com/doc/refman/8.0/en/keywords.html). But not in [mariadb keywords page](https://mariadb.com/kb/en/reserved-words/).
Should I remove the change for mariadb?

nene:
MySQL docs list all keywords, but MariaDB docs only the reserved ones (marked by "(R)" in MySQL docs).

nene:
As `IDENTIFIED` is not a reserved word, it can be freely used as an identifier. Simply adding it to the keywords array will result in a query like `SELECT identified FROM foo` to be formatted as `SELECT IDENTIFIED FROM foo`.

A better approach would be to add `"IDENTIFIED BY"` to the `reservedPhrases` array in `*.formatter.ts` files.

This also applies to [SingleStoreDB](https://docs.singlestore.com/cloud/reference/sql-reference/security-management-commands/create-user/) dialect (which is also a close relative of MySQL).

nene:
Thanks. Merged.

Issue #709, opened 2024-01-27
[FORMATTING] Field named "password" converted to uppercase when "keywordCase" is "upper"


vitorwdson:
**Input data**

Which SQL and options did you provide as input?

```sql
select
    id, username, password, email
from
    users
where
    username = 'foo' and password = 'bar';
```

**Expected Output**

```sql
SELECT
    id,
    username,
    password,
    email
FROM
    users
WHERE
    username = 'foo'
    AND password = 'bar';
```

**Actual Output**

```sql
SELECT
    id,
    username,
    PASSWORD,
    email
FROM
    users
WHERE
    username = 'foo'
    AND PASSWORD = 'bar';

```

**Usage**

- How are you calling / using the library?
```bash
sql-formatter -c ./sql-formatter.json
```
And the config is this
```json
{
  "language": "postgresql",
  "tabWidth": 4,
  "keywordCase": "upper",
  "dataTypeCase": "upper",
  "identifierCase": "lower",
  "linesBetweenQueries": 2
}
```

- What SQL language(s) does this apply to?
PostgreSQL

- Which SQL Formatter version are you using?
15.1.3

nene:
Thanks for reporting.

The PostgreSQL keywords list currently contains too many non-reserved keywords, like `PASSWORD`. Should likely remove most of them. There's a similar problem with `DB2` and `DB2i` dialects. But with PostgeSQL it should be simpler to fix.

nene:
Related to #156

nene:
Fixed and released in 15.2.0

karlhorky:
@nene this is now causing the following, is this expected?

```sql
CREATE TABLE animals (
  id integer PRIMARY key generated always AS identity
);
```

This is an unusual formatting of `KEY`, `GENERATED`, `ALWAYS`, and `IDENTITY`.

Before this change, it was the common style:

```sql
CREATE TABLE animals (
  id integer PRIMARY KEY GENERATED ALWAYS AS IDENTITY
);
```

nene:
Yep, that's an unfortunate side-effect of this change. Basically there are two choices in SQL Formatter:

- Detect too many words as keywords and end up uppercasing column names with `keywordCase: upper`.
- Detect too few words as keywords and end up not uppercasing these keywords with `keywordCase: upper`.

[...]

PR #714, opened 2024-02-05
Add primary key, generated column reserved phrases

karlhorky:
Add reserved phrases for PostgreSQL for primary keys and generated columns

After this change, the formatting results in the common style in the ecosystem:

```sql
CREATE TABLE animals (
  id integer PRIMARY KEY GENERATED ALWAYS AS IDENTITY
);
```

Before, the following configuration would result in the mixed lowercase for `KEY`, `GENERATED`, `ALWAYS`, and `IDENTITY`:

```sql
CREATE TABLE animals (
  id integer PRIMARY key generated always AS identity
);
```

Config (attempt to match style of official PostgreSQL docs):

```js
{
  language: 'postgresql',
  keywordCase: 'upper',
  identifierCase: 'lower',
  dataTypeCase: 'lower',
  functionCase: 'lower',
}
```

This was caused by the PostgreSQL keywords being removed in:

- https://github.com/sql-formatter-org/sql-formatter/issues/709

### DIFF
--- a/src/languages/postgresql/postgresql.formatter.ts
+++ b/src/languages/postgresql/postgresql.formatter.ts
@@ -246,6 +246,8 @@ const reservedJoins = expandPhrases([
 ]);
 
 const reservedPhrases = expandPhrases([
+  'PRIMARY KEY',
+  'GENERATED {ALWAYS | BY DEFAULT} AS IDENTITY',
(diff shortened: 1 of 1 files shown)


nene:
Thanks.

karlhorky:
Glad to help, thanks for the review and merge 🙌

PR #720, opened 2024-02-20
Format DEFAULT VALUES clause for Postgres

### LINKED ISSUE #718: [FORMATTING] `DEFAULT VALUES` not properly supported in INSERT statement
**Input data**

```sql
INSERT INTO workspaces DEFAULT VALUES RETURNING id;
```

options:
```js
{ language: 'postgresql', paramTypes: { named: ['$'] } }
```

**Expected Output**

The main issue is that `DEFAULT VALUES` [should be treated as a unit](https://www.postgresql.org/docs/current/sql-insert.html), as it is distinct from the usual `VALUES` or the query that typically follows `INSERT INTO`.

```sql
INSERT INTO
  workspaces
DEFAULT VALUES
RETURNING
  id;
```

**Actual Output**

```sql
INSERT INTO
  workspaces DEFAULT
VALUES
RETURNING
  id;
```

**Usage**

    - I'm using prettier, and we created a [custom prettier plugin](https://github.com/PrairieLearn/PrairieLearn/blob/master/packages/prettier-plugin-sql/src/index.ts) that uses sql-formatter to format the file.
    - This is specifically for Postgres.
    - 15.2.0

nwalters512:
Closes #718.

### DIFF
--- a/src/languages/postgresql/postgresql.formatter.ts
+++ b/src/languages/postgresql/postgresql.formatter.ts
@@ -23,6 +23,7 @@ const reservedClauses = expandPhrases([
   // - insert:
   'INSERT INTO',
   'VALUES',
+  'DEFAULT VALUES',
   // - update:
   'SET',
   // other
--- a/test/postgresql.test.ts
+++ b/test/postgresql.test.ts
@@ -331,4 +331,18 @@ describe('PostgreSqlFormatter', () => {
         release;
     `);
   });
+
+  it('formats DEFAULT VALUES clause', () => {
+    expect(
+      format(`INSERT INTO items default values RETURNING id;`, {
+        keywordCase: 'upper',
+      })
+    ).toBe(dedent`
+      INSERT INTO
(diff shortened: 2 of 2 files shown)


src/languages/postgresql/postgresql.formatter.ts
```diff
@@ -23,6 +23,7 @@ const reservedClauses = expandPhrases([
   // - insert:
   'INSERT INTO',
   'VALUES',
+  'DEFAULT VALUES',
```
nwalters512, line 26: I wasn't sure exactly which of the "clauses" arrays this belongs in, though it seems to not matter for the tests. If you'd rather this be moved to another array, please let me know.

nene (approved):

nene:
Thanks. There's indeed not much of a difference into which *clauses array to place this as there won't be no items following `DEFAULT VALUES` that should be nested inside it.

Issue #747, opened 2024-06-14
[FORMATTING] TransactSQL incorrectly formats some block comments, leading to broken statements


wylieconlon:
**Input data**

```sql
/** block comment **/
```

**Expected Output**

```sql
/** block comment **/
```

**Actual Output**

```sql
/ * * block comment * * /
```

**Usage**

This example reproduces on the sql-formatter playground. The only option that fixes the bug is enabling "dense operators", which produces this output:

```sql
/**block comment**/
```

nene:
Thanks for reporting.

Looks like this bug is related to the parsing of nested block-comments. Affects Transact-SQL, Postgres, DB2 for IMBi.

The the final `**/` seems to trigger it:

```sql
/* block comment **/
```

results in:

```sql
/ * block comment ** /
```

nene:
The fix for this is now out in 15.3.2

PR #748, opened 2024-06-14
fix formatting of block quotes with **/ closing characters

wylieconlon:
This is a common bug with Regex-based tokenizers. The previous regex was capturing too many characters in the "MIDDLE" regex, which was causing expressions like `/** comment **/` to be treated as individual operators instead of a comment.

The fix implemented here is to use a stack-based tokenizer which to parse nested comments instead. This appears to work as expected for TransactSQL dialects according to the [reference doc](https://learn.microsoft.com/en-us/sql/t-sql/language-elements/slash-star-comment-transact-sql?view=sql-server-ver16). 

Closes https://github.com/sql-formatter-org/sql-formatter/issues/747

### DIFF
--- a/src/lexer/NestedComment.ts
+++ b/src/lexer/NestedComment.ts
@@ -2,8 +2,6 @@
 import { RegExpLike } from './TokenizerEngine.js';
 
 const START = /\/\*/uy; // matches: /*
-const MIDDLE = /([^/*]|\*[^/]|\/[^*])+/uy; // matches text NOT containing /* or */
-const END = /\*\//uy; // matches: */
 
 /**
  * An object mimicking a regular expression,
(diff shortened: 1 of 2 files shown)


nene:
Hey, thanks for the patch. 👍 👍 👍 

It's rare to get a fix for a non-trivial issue like this.

I'll take some time to read the code (I'm super tired right now), but will most likely merge this in.

wylieconlon:
Please let me know if you have any feedback about the implementation! Thank you for making it easy to contribute :)

nene (commented):
Hey, after reviewing this fix, I decided to make a simpler one instead. See #751

I found this solution to be a bit too complex to read and understand. Wrote some comments about the problems I see with this code.

I did end up using the test you wrote, and included you accordingly to list of contributors.

Thanks.

PR #758, opened 2024-07-18
Add `GREATEST_IGNORE_NULLS` to `snowflake.functions.ts`

mxr:
See https://docs.snowflake.com/en/sql-reference/functions/greatest_ignore_nulls

Let me know if I am missing some steps. Thank you!


nene:
Thanks!

PR #766, opened 2024-08-09
Support named config file

### LINKED ISSUE #760: [VSCode] Feature Request: Support JSON config file
**Describe the Feature**
It would be nice if the VSCode extension could support reading a json config file in the same format expected by the CLI

I would like a single source of truth for SQL formatting configuration, rather than having to specify it once in workspace VSCode config and once in a json file for use with CLI in our CI / pre-commit autoformatter

schusj:
Closes #760

### DIFF
--- a/bin/sql-formatter-cli.cjs
+++ b/bin/sql-formatter-cli.cjs
@@ -132,17 +132,12 @@ class SqlFormatterCli {
 
   async readFile(filename) {
     try {
-
-
-    return promisify(fs.readFile)(filename, { encoding: 'utf-8' })
-    }  catch (e) {
-
-
+      return promisify(fs.readFile)(filename, { encoding: 'utf-8' });
+    } catch (e) {
       this.exitWhenIOError(e, filename);
       console.error('An unknown error has occurred, please file a bug report at:');
       console.log('https://github.com/sql-formatter-org/sql-formatter/issues\n');
       throw e;
-
     }
   }
 
--- a/bin/sql-formatter-cli.cjs
+++ b/bin/sql-formatter-cli.cjs
@@ -99,14 +99,14 @@ class SqlFormatterCli {
     return this.parseFile(localConfig);
   }
 
-  async findConfig(dir = __dirname) {
+  findConfig(dir = __dirname) {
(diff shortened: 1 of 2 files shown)


bin/sql-formatter-cli.cjs
```diff
@@ -72,57 +73,73 @@ class SqlFormatterCli {
...
+    // Otherwise find a local config file
+    const localConfig = await this.findConfig();
+    if (!localConfig) {
+      return null;
+    }
+
+    return this.parseFile(localConfig);
+  }
+
+  async findConfig(dir = __dirname) {
```
nene, line 102: It seems that this is not really an async function. At least the only `await` inside of it is calling the same function recursively.

bin/sql-formatter-cli.cjs
```diff
@@ -72,57 +73,73 @@ class SqlFormatterCli {
...
+      return JSON.parse(await this.readFile(filename));
+    } catch (e) {
+      console.error(`Error: unable to parse as JSON or treat as JSON file: ${filename}`);
+      process.exit(1);
+    }
+  }
+
   async readFile(filename) {
-    return promisify(fs.readFile)(filename, { encoding: 'utf-8' });
+    return promisify(fs.readFile)(filename, { encoding: 'utf-8' }).catch(e => {
```
nene, line 134: It would be more consistent with the rest of the code when this was written with `try {} catch` and `await` rather than calling `.catch()`.

nene (approved):
Thanks.

Looks like a nice little addition. I've added a few small comments, feel free to fix them if you feel like it. Otherwise I'll merge it as-is.

I should also point out that this does not actually resolve issue #760. That one is about VSCode extension, but the logic in `bin/sql-formatter-cli.cjs` is used only by the sql-formatter command line tool and won't in any way impact the VSCode extension which only uses the `sql-formatter` library (source of which resides in `src/` dir).

schusj:
Oh okay, I use the CLI with neovim and just assumed #760 was related / could be solved with this. Have resolved your comments.

nene:
This is now released in 15.4.0

PR #794, opened 2024-10-25
added support for decimal values without leading digits

jagmitswami:
Encountered issue while handling decimal values without leading digits (for example .00457). Lexer was breaking, updated number token regex to support the similar cases and added a corresponding test case.

Sample SQL query:  `SELECT employee_id FROM employees WHERE salary > .456 * 1000000 AND bonus < .0000239 * salary;`
```
An Unexpected Error Occurred
Parse error at token: 456 at line 1 column 51 Unexpected NUMBER token: {"type":"NUMBER","raw":"456","text":"456","start":50}
```

### DIFF
--- a/src/lexer/Tokenizer.ts
+++ b/src/lexer/Tokenizer.ts
@@ -51,7 +51,7 @@ export default class Tokenizer {
       {
         type: TokenType.NUMBER,
         regex:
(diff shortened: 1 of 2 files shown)


nene:
Thanks a lot!

Issue #809, opened 2024-12-19
[FORMATTING] a few cases of unexpected identifierCase "lower" behavior


gajus:
These are the unexpected diffs from PostgreSQL codebase after using the experiment `{ identifierCase: 'lower' }`:

```diff
- user_job_posting AS MATERIALIZED (
+ user_job_posting AS materialized (
```

```diff
- ON CONFLICT DO NOTHING
+ ON CONFLICT DO nothing
```

```diff
- ON CONFLICT (foreign_cloudinary_asset_id) DO NOTHING
+ ON CONFLICT (foreign_cloudinary_asset_id) DO nothing
```

```diff
- WITH ORDINALITY nids
+ WITH ordinality nids
```

```diff
- ORDER BY job_posting_approved_at DESC NULLS LAST,
+ ORDER BY job_posting_approved_at DESC nulls last,
```

```diff
- DROP MATERIALIZED VIEW IF EXISTS discover_independents_user_account_v3 CASCADE;
+ DROP MATERIALIZED VIEW if EXISTS discover_independents_user_account_v3 CASCADE;
```

```diff
- RESET ROLE;
+ RESET role;
```

```diff
- CREATE DATABASE "foo" TEMPLATE contra_template
+ CREATE DATABASE "foo" template contra_template
```

Otherwise, it works great.

Thanks for the awesome lib!

nene:
Well, there's a reason why the [identifierCase](https://github.com/sql-formatter-org/sql-formatter/blob/master/docs/identifierCase.md) option is marked as experimental.

Though really, all of the other case options (`keywordCase`, `dataTypeCase`, `functionCase`) are problematic as well. The root of all this is that SQL Formatter doesn't really fully parse the SQL. It only uses some heuristics to do it's formatting. But without proper parsing it can't really distinguish whether something like `last` is a keyword or identifier, because it depends on the context.

PostgreSQL in particular is pretty relaxed with regards to keywords - only a few keywords in PostgreSQL are fully reserved and can never be used as identifiers. (Though there are SQL dialects that are even worse in this regard.)

The approach SQL Formatter currently uses is to find some sort of sweetspot between two extremes:

- Detect anything that could be a keyword always as a keyword - results in lots of normal identifiers detected as keywords.
- Detect only the keywords that can never be used as identifiers as keywords - results in lots of keywords being detected as identifiers.

[...]

PR #820, opened 2025-01-05
Add DROP IDENTITY, OVERRIDING SYSTEM VALUE to PostgreSQL clauses

karlhorky:
Hey @nene, hope you're great! 👋 Happy New Year!

Quick PR to recognize `DROP IDENTITY` and `OVERRIDING SYSTEM VALUE` as one-line clauses

With the configuration:

```js
{
  keywordCase: 'upper',
  identifierCase: 'lower',
  dataTypeCase: 'lower',
  functionCase: 'lower',
}
```

This makes the following change:

`DROP IDENTITY`

```SQL
-- Before PR #820
ALTER TABLE events
ALTER COLUMN id
DROP identity;

-- After PR #820
ALTER TABLE events
ALTER COLUMN id
DROP IDENTITY;
```

`OVERRIDING SYSTEM VALUE`

```SQL
-- Before PR #820
INSERT INTO
  events (
    id,
    title,
    description
  ) overriding system value
VALUES
  (
    1,
    'abc',
    'def'
  );

-- After PR #820
INSERT INTO
  events (
    id,
    title,
    description
  )
OVERRIDING SYSTEM VALUE
VALUES
  (
    1,
    'abc',
    'def'
  );
```


nene:
Thanks. This is useful.

However I should point out that using `identifierCase: lower` is really just asking for trouble. See the [docs](https://github.com/sql-formatter-org/sql-formatter/blob/master/docs/identifierCase.md) for clarification. There's a reason this configuration option is marked as experimental.

karlhorky:
Interesting, I see that the commit f4cda2182f8705d317b2894eeb8f838e69cccc71 adding this explanation was motivated by #809 (hi again @gajus 👋)

Personally, I want this behavior - everything by default lowercase, and then the recognized keywords as uppercase. As such, I would personally evaluate each of these cases as (minor) bugs in SQL Formatter's dialects and keyword detection.

Opened a new PR for these particular cases mentioned by @gajus:

- https://github.com/sql-formatter-org/sql-formatter/pull/821

karlhorky:
Thanks also for the release in [`sql-formatter@15.4.9`](https://github.com/sql-formatter-org/sql-formatter/releases/tag/v15.4.9), can confirm it's working 👍

PR #821, opened 2025-01-05
Add more PostgreSQL keywords and phrases

### LINKED ISSUE #809: [FORMATTING] a few cases of unexpected identifierCase "lower" behavior

karlhorky:
Closes #809

```diff
- user_job_posting AS materialized (
+ user_job_posting AS MATERIALIZED (
```
>
```diff
- ON CONFLICT DO nothing
+ ON CONFLICT DO NOTHING
```
  
```diff
- ON CONFLICT (foreign_cloudinary_asset_id) DO nothing
+ ON CONFLICT (foreign_cloudinary_asset_id) DO NOTHING
```
  
```diff
- WITH ordinality nids
+ WITH ORDINALITY nids
```
  
```diff
- ORDER BY job_posting_approved_at DESC nulls last,
+ ORDER BY job_posting_approved_at DESC NULLS LAST,
```
  
```diff
+ DROP MATERIALIZED VIEW if EXISTS discover_independents_user_account_v3 CASCADE;
+ DROP MATERIALIZED VIEW IF EXISTS discover_independents_user_account_v3 CASCADE;
```
  
```diff
- RESET role;
+ RESET ROLE;
```

Some specific examples:

```sql
-- Before
SELECT
  fruit,
  pos 
FROM
  unnest(
    ARRAY['apple', 'banana', 'cherry']
  )
WITH
  ordinality AS fruits (fruit, pos);

-- After
SELECT
  fruit,
  pos 
FROM
  unnest(
    ARRAY['apple', 'banana', 'cherry']
  )
WITH ORDINALITY AS fruits (fruit, pos);
```

### DIFF
--- a/src/languages/postgresql/postgresql.formatter.ts
+++ b/src/languages/postgresql/postgresql.formatter.ts
@@ -172,7 +172,7 @@ const tabularOnelineClauses = expandPhrases([
   'DROP IDENTITY',
   'DROP INDEX',
   'DROP LANGUAGE',
-  'DROP MATERIALIZED VIEW',
+  'DROP MATERIALIZED VIEW [IF EXISTS]',
   'DROP OPERATOR',
   'DROP OPERATOR CLASS',
   'DROP OPERATOR FAMILY',
@@ -216,7 +216,7 @@ const tabularOnelineClauses = expandPhrases([
   'REFRESH MATERIALIZED VIEW',
   'REINDEX',
   'RELEASE SAVEPOINT',
-  'RESET',
+  'RESET [ALL|ROLE|SESSION AUTHORIZATION]',
   'REVOKE',
   'ROLLBACK',
   'ROLLBACK PREPARED',
@@ -252,11 +252,15 @@ const reservedPhrases = expandPhrases([
   'PRIMARY KEY',
   'GENERATED {ALWAYS | BY DEFAULT} AS IDENTITY',
   'ON {UPDATE | DELETE} [SET NULL | SET DEFAULT]',
+  'DO {NOTHING | UPDATE}',
+  'AS MATERIALIZED',
   '{ROWS | RANGE | GROUPS} BETWEEN',
   // https://www.postgresql.org/docs/current/datatype-datetime.html
   '[TIMESTAMP | TIME] {WITH | WITHOUT} TIME ZONE',
   // comparison operator
   'IS [NOT] DISTINCT FROM',
+  'NULLS {FIRST | LAST}',
+  'WITH ORDINALITY',
 ]);
 
 // https://www.postgresql.org/docs/14/index.html
--- a/src/languages/postgresql/postgresql.keywords.ts
+++ b/src/languages/postgresql/postgresql.keywords.ts
@@ -98,7 +98,6 @@ export const keywords: string[] = [
   'SYMMETRIC', // reserved
   'TABLE', // reserved
   'TABLESAMPLE', // reserved (can be function or type)
-  'TEMPLATE', // non-reserved, used in CREATE DATABASE to specify a template database
   'THEN', // reserved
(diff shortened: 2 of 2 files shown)


nene:
Thanks for the patch. Replying also to your comment in #820.

I agree that this using `identifierCase: lower` and `keywordCase: upper` is something that most people would want. It's just that with it's not possible to fully achieve that with the architecture of SQL Formatter. Some of these issues can be patched, like you've done in this PR, and I'm happy to accept such patches, but that can only take us so far. In particular it's mostly safe to introduce more multi-keyword syntax to be detected. However, when it comes to single keywords, we're getting into a problematic territory. Like in here, you've added `TEMPLATE` to keywords list, but this is not a reserved keyword - it could equally well be used as a name of a table or column, in which case it will now end up being uppercased.

```sql
SELECT template FROM tbl;

CREATE DATABASE foo TEMPLATE my_template;
```

Unfortunately SQL Formatter doesn't understand the context to know when `TEMPLATE` is used as a keyword and when it's used as an identifier.

karlhorky:
Right, I was considering whether this `'TEMPLATE'` change would be rejected for that reason. 

What I would really want is such a syntax (inspired by regular expressions), to allow for multi-keyword syntax with non-keyword parts in between:

```
"CREATE DATABASE '[^']+' TEMPLATE"
```

But I'm guessing that would be a larger change. 

For now, I could remove the `'TEMPLATE'` keyword instead, if that's a desirable path forward.

nene:
Yeah, that would be a major undertaking. There are several problems:

- These items between keywords can have quite a bit of syntax by themselves. Like the name of database could be either a quoted or unquoted identifier.
- The actual syntax supported by the SQL dialect might be more complex than what one might guess from a bug report. See [PostgreSQL docs for CREATE DATABASE](https://www.postgresql.org/docs/current/sql-createdatabase.html).
- There would also need to be a mechanism to distinguish which part of this expression are the keywords (that should e.g. be uppercased) and which parts are not.

For now, yeah, I see no better approach than just dropping `TEMPLATE` from keywords list.

karlhorky:
@nene makes sense, thanks for the consideration and guidance :) 

I'm guessing this and #820 will be included in a new published release in either the next patch or minor:

- next patch: [`sql-formatter@15.4.9`](https://github.com/sql-formatter-org/sql-formatter/releases/tag/v15.4.9)
- next minor: [`sql-formatter@15.5.0`](https://github.com/sql-formatter-org/sql-formatter/releases/tag/v15.5.0)

nene:
I released these in [15.4.9](https://github.com/sql-formatter-org/sql-formatter/releases/tag/v15.4.9)

PR #832, opened 2025-02-25
feat: suuport '$' as named param type

DepingMeng:


nene:
Thanks for the effort of making a pull request, but I can't accept this in its current form.

Which exact problem are you trying to solve? As the title of the pull request doesn't match with the code, I'm not sure what your actual intention here is.

If you think this syntax should be supported for PostgreSQL, please provide a link to official PostgreSQL documentation (or some other relevant source).

PR #836, opened 2025-03-05
feat: improve code splitting by introducing subpackage entry points

ruijdacd:
This PR introduces a solution outlined in [this comment](https://github.com/sql-formatter-org/sql-formatter/issues/452#issuecomment-1253353270), which enables more granular control over the bundle output.

### Current Problem

When utilizing the `formatDialect` API, the entire package is bundled together instead of just the specific `dialect` provided by the user. This results in unnecessary bloat in the bundle size.

To illustrate this issue, here is a simple reproduction of a Vite app. In this example, both `format(SQL_QUERY, { language: "postgresql" })` and `formatDialect(SQL_QUERY, { dialect: postgresql })` produce the same bundle size, demonstrating the problem clearly.

You can view the example here:
- [StackBlitz](https://stackblitz.com/edit/vitejs-vite-fkuwcjek?file=index.html)
- [Netlify](https://imaginative-dusk-b5d8a7.netlify.app/)


nene:
Thanks for the pull request, but I'm not sure I understand what you are trying to achieve here and how.

You're referencing my own comment... in an already solved issue. I don't quite get it.

You seem to claim that it's possible to detect at compile time which strings are used as the `language` option and only include sources for those SQL dialects. I don't understand how that's really possible. Would you care to explain.

ruijdacd:
> Thanks for the pull request, but I'm not sure I understand what you are trying to achieve here and how.
> 
> You're referencing my own comment... in an already solved issue. I don't quite get it.
> 
> You seem to claim that it's possible to detect at compile time which strings are used as the `language` option and only include sources for those SQL dialects. I don't understand how that's really possible. Would you care to explain.

I apologize for the lack of detail, but I noticed that tree-shaking wasn't functioning as intended, particularly in Vite. I attempted to use `formatDialect()`, but it ended up bundling everything, behaving similarly to `format()`.

The only way I managed to achieve a smaller bundle was by separating `formatDialect` from `format` and offering sub-imports for each language.

Let me know your thoughts and if you have any feedback!

ruijdacd:
I've made some changes to the Stackblitz to separate the reproduction into separate pages to visualize the JS size differences better.

nene:
Thanks for clarifying.

I should point out though that the fix you're proposing involves a major
breaking change to the API, which I won't be merging lightly. Before I do
merge it I would really like to find out why the current implementation
isn't working and if there's some way to make it work. If you would give a
hand in investigating that, I would very much appreciate it. If no, that's
fine as well, but then I'll have to do it myself.

PS. Haven't yet tried to reproduce the problem you're reporting or tried
your fix. Just my initial general thoughts about this.

ruijdacd:
Yeah, totally! I didn't except this to get released quickly. I'd be happy to help out wherever needed

nene:
So, I analyzed this thing and found out that this code splitting did work as expected in the 12.0.0 version where it was first introduced. It was still working in `14.0.0`, but it broke in `15.0.0`.

Not sure which of [the changes from 14 to 15](https://github.com/sql-formatter-org/sql-formatter/compare/v14.0.0...v15.0.0) might have caused it.

nene:
So, the cause seems to be #670. Strangely there's a comment that says:

> Removing the export * was necessary because esbuild transformed it into code that wasn't statically analyzable for tree shaking

However the end result seems to be the opposite. Hmm...

nwalters512:
@nene I opened #842 to resolve the code-splitting issue.

nene:
@ruijdacd I've released `15.5.0-beta.2` with a [fix for this issue](https://github.com/sql-formatter-org/sql-formatter/pull/842). Would be nice if you could give it a try to see if it also fixes the problem for you.

nene:
I'm closing this as the problem has been solved by different approach in [15.5.0](https://github.com/sql-formatter-org/sql-formatter/releases/tag/v15.5.0).

@ruijdacd feel free to still get back to me if this turns out to not fix the problem for you.

PR #842, opened 2025-03-16
Build with tsc instead of tsup

nwalters512:
As reported in https://github.com/sql-formatter-org/sql-formatter/pull/836#issuecomment-2727273274, my previous changes from #670 broke the ability for this package to be tree-shaken.

This PR takes the other approach I had proposed in #604: it runs `tsc` twice, once to produce a CJS build and once to produce an ESM one. It writes a `package.json` file containing `{"type": "commonjs"}` to `dist/cjs/package.json` so that the files in that directory are treated as CJS.

Since #670 introduced an `exports` field, the internal changes to paths shouldn't be visible to consumers, so this should be safe to release as a patch or minor version. That said, I wouldn't blame you if you wanted to release this as a major version to play it safe.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,14 +3,14 @@
   "version": "15.4.11",
   "description": "Format whitespace in a SQL query to make it more readable",
   "license": "MIT",
-  "main": "dist/index.cjs",
-  "module": "dist/index.js",
+  "main": "dist/cjs/index.cjs",
+  "module": "dist/esm/index.js",
   "unpkg": "dist/sql-formatter.min.js",
   "exports": {
     "./package.json": "./package.json",
     ".": {
-      "import": "./dist/index.js",
-      "require": "./dist/index.cjs"
+      "import": "./dist/esm/index.js",
+      "require": "./dist/cjs/index.cjs"
     }
   },
   "bin": {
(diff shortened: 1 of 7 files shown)


nene:
Thanks a whole lot. This looks mighty promising. Too tired today to check
it out. Hopefully will manage tomorrow.

package.json
```diff
@@ -63,9 +63,10 @@
     "prepare": "yarn clean && yarn grammar && yarn fix && yarn check && yarn build",
     "pre-commit": "npm-run-all --parallel ts:changes lint:changes",
     "grammar": "nearleyc src/parser/grammar.ne -o src/parser/grammar.ts",
-    "build:tsup": "tsup src/index.ts --format cjs,esm --sourcemap --dts",
+    "build:cjs": "tsc -p tsconfig.cjs.json && echo \"{\"type\": \"commonjs\"}\" > dist/cjs/package.json",
```
nene, line 66: Hmm... I think this doesn't work quite as intended. I generates a file containing:

```
{type: commonjs}
```

which isn't valid JSON. I tweaked it a bit: https://github.com/sql-formatter-org/sql-formatter/commit/14e5daff45d1c257d1d6d68da90aa95a1a0c37c1

nene:
So I merged this in and released as `15.5.0-beta.1`. Then fixed the package.json issue and released `15.5.0-beta.2`.

PR #843, opened 2025-03-16
Run typechecking in CI

nwalters512:
The upgrades to Jest-related packages were the easiest solution to fix this type error that occurred when `yarn ts:check` was run on `master`:

```
node_modules/@jest/environment/build/index.d.ts:327:26 - error TS2430: Interface 'JestImportMeta' incorrectly extends interface 'ImportMeta'.
  The types returned by 'jest.createMockFromModule(...)' are incompatible between these types.
    Type 'unknown' is not assignable to type 'T'.
      'T' could be instantiated with an arbitrary type which could be unrelated to 'unknown'.

327 export declare interface JestImportMeta extends ImportMeta {
                             ~~~~~~~~~~~~~~


Found 1 error in node_modules/@jest/environment/build/index.d.ts:327
```

Jest 29 changed the default snapshot version (see https://jestjs.io/docs/upgrading-to-jest29#snapshot-format). I updated the existing snapshots to use that. I'd also be happy to change the Jest config if you wanted to minimize the diff.


nene:
Thanks for that. Works great. ❤️

PR #844, opened 2025-03-20
chore: replace get-stdin with native stream/consumers

outslept:
This follows the recommendation from [Sindre Sorhus (author of get-stdin)](https://github.com/sindresorhus/get-stdin/tree/main?tab=readme-ov-file#tip) to use the built-in streamConsumers.text() in modern Node.js versions. Reduces external dependencies while maintaining identical functionality.


nene:
Thanks. This is now released in [15.5.0](https://github.com/sql-formatter-org/sql-formatter/releases/tag/v15.5.0).

PR #851, opened 2025-03-25
fix(postgresql): add options to create statements

jamesarosen:
Many of the `CREATE ...` statements in the postgres dialect were missing options, especially the `OR REPLACE` option that is common to many commands.

Fixes #850


src/languages/postgresql/postgresql.formatter.ts
```diff
@@ -128,18 +128,18 @@ const tabularOnelineClauses = expandPhrases([
...
-  'CREATE PROCEDURE',
+  'CREATE [OR REPLACE] PROCEDURE',
   'CREATE PUBLICATION',
   'CREATE ROLE',
-  'CREATE RULE',
-  'CREATE SCHEMA',
-  'CREATE SEQUENCE',
+  'CREATE [OR REPLACE] RULE',
+  'CREATE SCHEMA [AUTHORIZATION]',
+  'CREATE {TEMPORARY | TEMP | UNLOGGED} SEQUENCE',
```
jamesarosen, line 142: I _think_ this is the right syntax. Please compare with https://www.postgresql.org/docs/17/sql-createsequence.html

nene: Not quite. The `[ { TEMPORARY | TEMP } | UNLOGGED ]` in postgres docs is really equivalent to just: `[ TEMPORARY | TEMP | UNLOGGED ]`.

nene:
Thanks for putting in effort to fix more things than just your immediate problem 👍

nene:
Released in [15.5.2](https://github.com/sql-formatter-org/sql-formatter/releases/tag/v15.5.2)

PR #896, opened 2025-09-02
support timezone phrase as datatype not as a keyword

eenagy:
Fixed SQL formatting for timezone data types like `TIMESTAMP WITH TIME ZONE` - these were incorrectly treated as keywords instead of data types, so `dataTypeCase: 'lower'` settings now work correctly for timestamp phrases.


nene:
Released in v15.6.7

Issue #614, opened 2023-06-22
Feature Request: ClickHouse dialect


derN3rd:
**Describe the Feature**

Hey there,
I'm currently very happy with sql-formatter in my code base, but I've just started to use ClickHouse for an analytics project and figured out that the formatting is breaking the SQL files (resulting in invalid queries).

Don't exactly know if one of the existing languages is similar to ClickHouse, but it seems not, as they have added a few custom keywords.

Their SQL reference is here: https://clickhouse.com/docs/en/sql-reference

Thanks in advance

**Why do you want this feature?**
Dialect of ClickHouse seems to be different from already provided languages

nene:
Thanks for the feature request.

From my first glance it indeed doesn't look very similar to any other SQL dialect I know of. Notably it has the `$foo$heredoc$foo$` strings like Postgres, while supporting both `"double quotes"` and `` `backticks` `` for quoting identifiers like SQLite.

If you're interested in getting a support for it, you unfortunately need to put in some effort to make it happen. I'm open for a pull request, but I'm not really interested in implementing it by myself. Luckily for you, adding a support for a new dialect to sql-formatter is not that hard. You don't even need to fork sql-formatter to do so, you can just provide it a [custom dialect definition](https://github.com/sql-formatter-org/sql-formatter/blob/master/docs/dialect.md#custom-dialect-configuration-experimental).

Well, admittedly there's not that much of documentation there, but just copying an existing dialect implementation from sql-formatter source code and modifying it slightly should be pretty simple.

You can also look at the [wiki](https://github.com/sql-formatter-org/sql-formatter/wiki), which documents syntax details of all the currently supported SQL dialects.

garysassano:
Any update on this?

nene:
I'm personally not going to implement any new dialects. Submit a pull request if you're interested. The implementation part is not really hard, the hard part is reading all the documentation and figuring out all the details about the syntax of the dialect.

nene:
Clickhouse support is now available in [15.7.0](https://github.com/sql-formatter-org/sql-formatter/releases/tag/v15.7.0)

PR #922, opened 2025-11-14
[WIP] Add preliminary Clickhouse formatter

mattbasta:
Addresses https://github.com/sql-formatter-org/sql-formatter/issues/614

I've started putting together a Clickhouse formatter. Before I start getting into the weeds with tests and updating all of the docs/tests/playground, I wanted to check that things are directionally correct. Please let me know if there's anything you'd like to see done differently before I proceed.

### DIFF
--- a/.github/workflows/webpack.yaml
+++ b/.github/workflows/webpack.yaml
@@ -22,6 +22,12 @@ jobs:
       - name: Install
         run: yarn install --ignore-scripts
 
+      - name: Format
+        run: yarn pretty:check
+
+      - name: Lint
+        run: yarn lint
+
       - name: Test
         run: yarn test
 


src/languages/clickhouse/clickhouse.formatter.ts
```diff
@@ -0,0 +1,161 @@
...
+    reservedFunctionNames: functions,
+    nestedBlockComments: false,
+    underscoresInNumbers: true,
+    stringTypes: ['$$', "''-qq", "''-qq-bs"],
+    identTypes: ['""-qq', '``'],
+    paramTypes: {
+      // https://clickhouse.com/docs/sql-reference/syntax#defining-and-using-query-parameters
+      custom: [
+        {
+          regex: String.raw`\{\s*[a-zA-Z0-9_]+\s*:\s*[a-zA-Z0-9_]+\s*\}`,
```
nene: That regex would fail when a more complex type is used in the parameter, like this:

```sql
SELECT {d: Map(String, Array(UInt8))};
```

src/languages/clickhouse/clickhouse.formatter.ts
```diff
@@ -0,0 +1,161 @@
...
+]);
+
+const standardOnelineClauses = expandPhrases([
+  // https://clickhouse.com/docs/sql-reference/statements/create
+  'CREATE [OR REPLACE] [TEMPORARY] TABLE [IF NOT EXISTS]',
+  // https://clickhouse.com/docs/sql-reference/statements/update
+  'UPDATE',
+  // https://clickhouse.com/docs/sql-reference/statements/system
+  'SYSTEM RELOAD {DICTIONARIES | DICTIONARY | FUNCTIONS | FUNCTION | ASYNCHRONOUS METRICS} [ON CLUSTER]',
+  'SYSTEM DROP {DNS CACHE | MARK CACHE | ICEBERG METADATA CACHE | TEXT INDEX DICTIONARY CACHE | TEXT INDEX HEADER CACHE | TEXT INDEX POSTINGS CACHE | REPLICA | DATABASE REPLICA | UNCOMPRESSED CACHE | COMPILED EXPRESSION CACHE | QUERY CONDITION CACHE | QUERY CACHE | FORMAT SCHEMA CACHE | DROP FILESYSTEM CACHE}',
```
nene: `SYSTEM DROP {... | DROP FILESYSTEM CACHE}`

will result in:

`SYSTEM DROP DROP FILESYSTEM CACHE`

src/languages/clickhouse/clickhouse.formatter.ts
```diff
@@ -0,0 +1,161 @@
...
+  'SHOW [EXTENDED] [FULL] COLUMNS {FROM | IN}',
+  // https://clickhouse.com/docs/sql-reference/statements/attach
+  'ATTACH {TABLE | DICTIONARY | DATABASE} [IF NOT EXISTS]',
+  // https://clickhouse.com/docs/sql-reference/statements/detach
+  'DETACH {TABLE | DICTIONARY | DATABASE} [IF EXISTS]',
+  // https://clickhouse.com/docs/sql-reference/statements/drop
+  'DROP {DICTIONARY | DATABASE | USER | ROLE | QUOTA | PROFILE | SETTINGS PROFILE | VIEW | FUNCTION | NAMED COLLECTION | ROW POLICY | POLICY} [IF EXISTS]',
+  'DROP [TEMPORARY] TABLE [IF EXISTS] [IF EMPTY]',
+  // https://clickhouse.com/docs/sql-reference/statements/exists
+  'EXISTS [TEMPORARY] {TABLE | DICTIONARY | DATABASE}',
```
nene: The linked docs say that the `TABLE | DICTIONARY | DATABASE` part is optional.

But making all that optional would result in `EXISTS (...)` expressions being detected as `EXISTS` statements.

So I guess you made a deliberate choice here.

src/languages/clickhouse/clickhouse.keywords.ts
```diff
@@ -0,0 +1,782 @@
...
+  'YEAR',
+  'YEARS',
+  'YY',
+  'YYYY',
+  'ZKPATH',
+];
+
+export const keywordPhrases: string[] = [
+  // See documentation of `keywords` above
+  'ADD COLUMN',
```
nene: These should really be just part of keywords.

The -phrases variables should be used for the cases where you're using the `expandPhrases()` function to expand expressions like `TIMESTAMP {WITH | WITHOUT} TIME ZONE`.

When you just need multi-word keywords, they work just fine out of the box. Just like you're using multi-word data type names further down.

src/languages/clickhouse/clickhouse.formatter.ts
```diff
@@ -0,0 +1,161 @@
...
+import { expandPhrases } from '../../expandPhrases.js';
+import { functions } from './clickhouse.functions.js';
+import { dataTypes, keywords, keywordPhrases } from './clickhouse.keywords.js';
+
+const reservedSelect = expandPhrases(['SELECT [DISTINCT]']);
+
+const reservedClauses = expandPhrases([
+  // https://clickhouse.com/docs/sql-reference/statements/explain
+  'EXPLAIN [AST | SYNTAX | QUERY TREE | PLAN | PIPELINE | ESTIMATE | TABLE OVERRIDE]',
+]);
```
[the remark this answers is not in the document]
nene: Also, I don't expect to see the `EXPLAIN` statement in here - for all other dialects the `EXPLAIN` statement has been listed under `tabularOnelineClauses`.

mattbasta:
@nene I appreciate you taking the time to look! I'm very aware that it's not nearly ready to land and that it needs quite a bit of love, I just didn't want to invest a ton more time if it's completely off the mark. I have a commit in progress with the tests; I'm trying to aim for nearly all of the example queries from the docs to get handled correctly. I'll ping you when I have something that's ready for review.

src/languages/clickhouse/clickhouse.formatter.ts
```diff
@@ -0,0 +1,362 @@
...
+    tabularOnelineClauses,
+  },
+};
+
+/**
+ * Converts IN and ANY from RESERVED_FUNCTION_NAME to RESERVED_KEYWORD
+ * when they are used as operators (not function calls).
+ *
+ * IN operator: foo IN (1, 2, 3) - IN comes after an identifier/expression
+ * IN function: IN(foo, 1, 2, 3) - IN comes at start or after operators/keywords
```
nene: I see `in` listed in section titled [Functions for Implementing the IN Operator](https://clickhouse.com/docs/sql-reference/functions/in-functions)

This gives me the impression that `in` being available as a function is more of an implementation detail of Clickhouse and one normally wouldn't use it like so.

src/languages/clickhouse/clickhouse.formatter.ts
```diff
@@ -0,0 +1,362 @@
...
+function postProcess(tokens: Token[]): Token[] {
+  return tokens.map((token, i) => {
+    const nextToken = tokens[i + 1] || EOF_TOKEN;
+    const prevToken = tokens[i - 1] || EOF_TOKEN;
+
+    // Only process IN and ANY that are currently RESERVED_FUNCTION_NAME
+    // Check text (uppercase canonical form) for matching, but preserve raw (original casing)
+    if (
+      token.type === TokenType.RESERVED_FUNCTION_NAME &&
+      (token.text === 'IN' || token.text === 'ANY')
```
nene: I see `IN` and `ANY` being listed in both `clickhouse.keywords.ts` and `clickhouse.functions.ts`. There seems to be many more. Like there's `AND`, `OR`, `NOT`, `CAST`, `DATE` in both functions and keywords.

Having them listed in both places creates confusion, as it's not obvious which one takes priority. It happens to be that function names are detected before keywords by the tokenizer, so these all end up being classified as function names. But it would be better to avoid this ambiguity by only listing them in one place.

I would suggest listing them in the category that's the more common use case. Like I would move all these things to the keywords list and remove them completely from functions. This way it also aligns better with other dialects.

PS.  `AND` and `OR` won't really have any effect in either list as these are hard-coded to the tokenizer and will always get detected as special type of tokens.

mattbasta:
Hey @nene, apologies for the delay in following up. It's been a busy month here.

I think I've addressed all of your comments. Please let me know if there's anything that you'd like changed or that you aren't happy with in the current implementation.

nene:
Thanks for all the work. Sorry for the delay in responding, but it looks all good now.

Also thanks for updating the Github workflow file. (FYI, I didn't at all expect you to do this, but it's great that you did.)

PR #924, opened 2025-12-09
Add PostgreSQL COMMENT keyword

rpaterson:
https://www.postgresql.org/docs/14/sql-comment.html

### DIFF
--- a/src/languages/postgresql/postgresql.keywords.ts
+++ b/src/languages/postgresql/postgresql.keywords.ts
@@ -18,6 +18,7 @@ export const keywords: string[] = [
   'COLLATE', // reserved
   'COLLATION', // reserved (can be function or type)
   'COLUMN', // reserved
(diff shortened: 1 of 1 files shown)


nene:
Thanks for the pull request.

Please describe what sort of issue you are trying to solve with this change.

FYI, this list of keywords that you edited is supposed to only include reserved keywords, and `COMMENT` is not a reserved keyword in PostgreSQL.

rpaterson:
I have sql-formatter configured for PostgreSQL and to capitalize keywords,
but it's not capitalising COMMENT.

nene:
I'd be happy to accept a PR with that change. For today I'm going to sleep :)

FYI, if you're looking for a formatter that doesn't have this sort of limitation, I suggest you check out [prettier-plugin-sql-cst](https://github.com/nene/prettier-plugin-sql-cst)

PR #938, opened 2026-03-27
Add missing PostgreSQL operators and fix >^ geometric operator

rathboma:
## Summary
- Adds support for 5 PostgreSQL operators from core and popular extensions that were incorrectly tokenized (split into shorter operators): `<@>` (earthdistance), `&&&` (PostGIS 3D overlap), `|=|` (PostGIS trajectory distance), `~>` (cube), `#=` (hstore)
- Fixes the geometric "is above" operator which was defined as `^>` instead of `>^` per [PostgreSQL docs](https://www.postgresql.org/docs/current/functions-geometry.html)
- Adds unit tests for each operator in realistic SQL contexts, with links to relevant PostgreSQL/PostGIS documentation

## Detail

The formatter's tokenizer was greedily matching shorter known operators instead of the full operator. For example `<@>` was being split into `<@` + `>`, producing `bar <@ > point(1, 2)` instead of `bar <@> point(1, 2)`.

The fix is simply adding these operators to the PostgreSQL dialect's operator list. The existing regex builder in `regexFactory.ts` already sorts operators by length descending, so longer operators automatically take priority.

### Operators added

| Operator | Source | Docs |
|----------|--------|------|
| `<@>` | earthdistance extension | [earthdistance](https://www.postgresql.org/docs/current/earthdistance.html) |
| `&&&` | PostGIS | [geometry_overlaps_nd](https://postgis.net/docs/geometry_overlaps_nd.html) |
| `\|=\|` | PostGIS | [geometry_distance_cpa](https://postgis.net/docs/geometry_distance_cpa.html) |
| `~>` | cube extension | [cube](https://www.postgresql.org/docs/current/cube.html) |
| `#=` | hstore extension | [hstore](https://www.postgresql.org/docs/current/hstore.html) |
| `>^` (fix) | geometric | [functions-geometry](https://www.postgresql.org/docs/current/functions-geometry.html) |

## Test plan
- [x] All new operators tested via `supportsOperators()` for basic spacing and dense mode
- [x] All new operators tested in realistic SQL contexts (WHERE, ORDER BY, SELECT)
- [x] Full PostgreSQL test suite passes (460 tests)

### DIFF
--- a/src/languages/postgresql/postgresql.formatter.ts
+++ b/src/languages/postgresql/postgresql.formatter.ts
@@ -319,6 +319,7 @@ export const postgresql: DialectOptions = {
       '##',
       '<->',
       '&&',
+      '&&&',
       '&<',
       '&>',
       '<<|',
@@ -326,7 +327,7 @@ export const postgresql: DialectOptions = {
       '|>>',
       '|&>',
       '<^',
-      '^>',
+      '>^',
       '?#',
       '?-',
       '?|',
@@ -334,7 +335,10 @@ export const postgresql: DialectOptions = {
       '?||',
       '@>',
       '<@',
+      '<@>',
       '~=',
+      // PostGIS
+      '|=|',
       // JSON
       '?',
       '@?',
@@ -376,6 +380,10 @@ export const postgresql: DialectOptions = {
       '<->>',
       '<<<->',
       '<->>>',
+      // Cube
+      '~>',
+      // Hstore
+      '#=',
       // Type cast
       '::',
       ':',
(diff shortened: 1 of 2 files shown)


rathboma:
@nene You're right I should have been more up-front in my use of AI. I was in a rush so omitted it. Please accept my apologies.

I also run a big open source project so I understand the frustration :-(.

Anyway -- thank you for taking your time to maintain this library, it is much appreciated! You're doing great work.

nene:
Thanks. It's all good :)

PR #944, opened 2026-05-12
fix(sqlite): support Tcl-style $param(...) syntax

### LINKED ISSUE #943: Problem with $p() syntax in SQLite
**Input data**

```sql
select $p(x=y);
```

**Expected Output**

```sql
select $p(x=y);
```

**Actual Output**

```sql
select $p (x=y);
```

**Usage**

cli
sqlite
- Which SQL Formatter version are you using? (The exact version number.)
15.6.10

I'm not 100% sure but I think in the tcl `$()` param syntax sqlite supports requires the variable name and following parens to be touching (no space). At least, from manual testing in the cli that seemed to be the case.

SAY-5:
Closes #943.

SQLite's `$`-prefixed parameter follows [Tcl variable syntax](https://www.sqlite.org/lang_expr.html#parameters), which allows one or more `::`-separated suffixes and an optional `(...)` trailer. Previously the formatter only accepted a bare identifier after `$`, so inputs like `$p(x=y)` were tokenized as `$p` followed by a parenthesized expression and re-emitted with an unwanted space.

This moves SQLite's `$`-parameter into a dialect-specific custom rule that matches the full Tcl syntax, while `:` and `@` continue to use the default named-parameter path.

### Before
```
> format('select $p(x=y);', { language: 'sqlite' });
'select\n  $p (x = y);'
```

### After
```
> format('select $p(x=y);', { language: 'sqlite' });
'select\n  $p(x=y);'
```

Tests cover the existing `$name` behaviour as well as `$p(x=y)` and `$foo::bar(extra)`.

### DIFF
--- a/src/languages/sqlite/sqlite.formatter.ts
+++ b/src/languages/sqlite/sqlite.formatter.ts
@@ -90,7 +90,19 @@ export const sqlite: DialectOptions = {
     ],
     identTypes: [`""-qq`, '``', '[]'],
     // https://www.sqlite.org/lang_expr.html#parameters
-    paramTypes: { positional: true, numbered: ['?'], named: [':', '@', '$'] },
+    // Note: the $-prefixed form follows Tcl variable syntax and may include
+    // one or more "::"-separated suffixes and an optional "(...)" trailer.
+    paramTypes: {
+      positional: true,
+      numbered: ['?'],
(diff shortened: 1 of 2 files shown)


nene (approved):
Awesome!

PR #946, opened 2026-05-14
Remove config for CodeSandbox CI

karlhorky:
[CodeSandbox CI is being discontinued](https://news.ycombinator.com/item?id=47418491#:~:text=CodeSandbox%20CI%3A%20Shutting%20down%20on%20April%2015%2C%202026.%20Migrate%20to%20an%20alternative%20CI%20provider%20before%20then) and its environment is stuck on Node.js v20, which is ancient and incompatible with an increasing range of new libraries.

- [x] Remove CodeSandbox CI config
- [x] Remove CodeSandbox CI app (in [`sql-formatter-org` GitHub Apps settings](https://github.com/organizations/sql-formatter-org/settings/installations), probably needs @nene)
- [ ] Decide on a replacement, if any

## Retaining the functionality

For Preflight, we switched to Git exotic sources (eg. `pnpm add "upleveled/preflight#switch-codesandbox-to-pnpm-github-urls"`), but this was possible because we have no build step:

- https://github.com/upleveled/preflight/pull/739

I'm guessing this is not possible for `sql-formatter`, at least how it is currently set up.

So another option would be `pkg.pr.new` from StackBlitz:

- https://pkg.pr.new/


nene:
Suits me. I've never really used these codesandbox builds anyway.

nene:
Uninstalled the CodeSandbox app.

nene:
I'm fine with not having a replacement.

PR #945, opened 2026-05-14
Add `devEngines.packageManager` field to package.json

karlhorky:
Hi @nene 👋 Hope thing are good 🙂 

## Problem

I was cloning the repo again and when I tried to use `yarn` to install dependencies (because I saw the `yarn.lock` file, Yarn 4.13.0 was selected by Corepack on my machine:

```bash
➜  sql-formatter git:(master) yarn
➤ [YN0065](https://yarnpkg.com/advanced/error-codes#yn0065---telemetry_notice): Yarn will periodically gather anonymous telemetry: https://yarnpkg.com/advanced/telemetry
➤ [YN0065](https://yarnpkg.com/advanced/error-codes#yn0065---telemetry_notice): Run yarn config set --home enableTelemetry 0 to disable

➤ [YN0087](https://yarnpkg.com/advanced/error-codes#yn0087---migration_success): Migrated your project to the latest Yarn version 🚀

➤ YN0000: · Yarn 4.13.0
➤ YN0000: ┌ Resolution step
➤ [YN0085](https://yarnpkg.com/advanced/error-codes#yn0085---updated_resolution_record): │ + @babel/cli@npm:7.25.9, @babel/core@npm:7.26.0, @babel/plugin-proposal-class-properties@npm:7.18.6, @babel/preset-env@npm:7.26.0, and 931 more.
➤ YN0000: └ Completed in 2s 905ms
➤ YN0000: ┌ Post-resolution validation
➤ [YN0086](https://yarnpkg.com/advanced/error-codes#yn0086---explain_peer_dependencies_cta): │ Some peer dependencies are incorrectly met by dependencies; run yarn explain peer-requirements for details.
➤ YN0000: └ Completed
➤ YN0000: ┌ Fetch step
➤ [YN0013](https://yarnpkg.com/advanced/error-codes#yn0013---fetch_not_cached): │ 935 packages were added to the project (+ 237.88 MiB).
➤ YN0000: └ Completed in 3s 341ms
➤ YN0000: ┌ Link step
➤ YN0000: └ Completed in 2s 915ms
➤ YN0000: · Done with warnings in 9s 283ms
```

I think this is probably not the correct package manager - the project intends to use Yarn v1, right?

## Solution

This PR specifies `yarn@1.22.22` in [the `devEngines.packageManager` field](https://docs.npmjs.com/cli/v11/configuring-npm/package-json#devengines) of `package.json, to enable the following behavior:

1. If Corepack is installed:
   1. block usage of incorrect package manager
[...]

PR #949, opened 2026-05-14
Switch from yarn@1.22.22 to pnpm@11.0.8

karlhorky:
Switch from yarn@1.22.22 to pnpm@11.0.8, as per https://github.com/sql-formatter-org/sql-formatter/pull/945#issuecomment-4450261762.

This also pins `@types/node@22.9.0`, matching the version previously resolved through `yarn.lock`. Without the pin, pnpm resolves `@types/node@25.6.0`, which TypeScript 4.9 cannot check. The observed diagnostics:

- `error TS1165: A computed property name in an ambient context must refer to an expression whose type is a literal type or a 'unique symbol' type`
- `error TS2339: Property 'dispose' does not exist on type 'SymbolConstructor'`
- `error TS2726: Cannot find lib definition for 'esnext.disposable'`

Install pnpm via the methods on the installation page:

- https://pnpm.io/installation

TODO:

- [x] Delete `yarn.lock`
- [x] Generate `pnpm-lock.yaml`
- [x] Update package scripts from Yarn commands to pnpm commands
- [x] Update GitHub Actions from Yarn commands to pnpm commands
- [x] Update contributor docs from Yarn commands to pnpm commands
- [x] Remove the README install example with Yarn
- [x] Update CI from Node 18 to Node 22 for pnpm 11 compatibility
- [x] Enable pnpm dependency caching in CI
- [x] Pin `devEngines.packageManager` to `pnpm@11.0.8`
- [x] Pin `@types/node@22.9.0`
- [x] Verify dependencies before scripts with `verifyDepsBeforeRun: error`
- [x] Declare the root workspace package with `packages: ['.']`
- [x] Block packages newer than seven days with `minimumReleaseAge: 10080`
- [x] Apply release-age checks strictly with `minimumReleaseAgeStrict: true`
- [x] Treat missing publish times as unsafe with `minimumReleaseAgeIgnoreMissingTime: false`
- [x] Prevent trust downgrades with `trustPolicy: no-downgrade`
- [x] Allow the trust-policy exception `eslint-config-prettier@8.10.2`
- [x] Allow the trust-policy exception `eslint-plugin-prettier@4.2.5`
- [x] Allow the trust-policy exception `semver@5.7.2 || 6.3.1`
- [x] Allow the trust-policy exception `undici-types@6.19.8`
- [x] Block exotic subdependencies with `blockExoticSubdeps: true`

## Summary by CodeRabbit

* **Chores**
  * Updated CI and project tooling to Node.js 22.x.
  * Migrated package management from Yarn to pnpm across CI, scripts, and project config.
  * Added a pnpm workspace with stricter dependency and verification policies.
  * Updated developer scripts and bumped dev tooling/type dependency versions.

* **Documentation**
  * Updated install and contribution instructions to reflect pnpm and Node.js 22.x.

[Review Change Stack](https://app.coderabbit.ai/change-stack/sql-formatter-org/sql-formatter/pull/949)


karlhorky:
@coderabbitai review

.github/workflows/coveralls.yaml
```diff
@@ -10,7 +10,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [18.x]
+        node-version: [22.x]
```
karlhorky, line 13: Node.js 18 and 20 are EOL

.github/workflows/webpack.yaml
```diff
@@ -18,21 +18,25 @@ jobs:
         uses: actions/setup-node@v4
         with:
           node-version: ${{ matrix.node-version }}
+          cache: pnpm
+
+      - name: Enable Corepack
+        run: corepack enable
```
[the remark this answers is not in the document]
karlhorky: done in 2f7e7a9c

pnpm-lock.yaml
```diff
@@ -0,0 +1,11614 @@
...
+      {
+        integrity: sha512-vuyHg81vvWA1Z1ELfvLko2c8f34gyA0zaic0+Rllc5lbCnbSyuvb2Oxpm6TAUAC/2xZN3QGqxBNggD1nNR2AfQ==,
+      }
+
+  '@types/node@25.6.0':
+    resolution:
+      {
+        integrity: sha512-+qIYRKdNYJwY3vRCZMdJbPLJAtGjQBudzZzdzwQYkEPQd+PJGixUL5QfvCLDaULoLv+RhT3LDkwEfKaAkgSmNQ==,
+      }
+
```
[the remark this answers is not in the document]
karlhorky: invalid - CI does not fail on newer transitive `@types/node` versions

.github/workflows/coveralls.yaml
```diff
@@ -19,12 +19,16 @@ jobs:
         uses: actions/setup-node@v4
         with:
           node-version: ${{ matrix.node-version }}
+          cache: pnpm
+
+      - name: Enable Corepack
+        run: corepack enable
```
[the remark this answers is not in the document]
karlhorky: done in 2f7e7a9c

package.json
```diff
@@ -92,6 +92,7 @@
     "@types/babel__core": "^7.1.15",
     "@types/jest": "^29.5.14",
     "@types/nearley": "^2.11.2",
+    "@types/node": "22.9.0",
```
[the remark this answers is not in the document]
karlhorky: invalid - CI does not fail on newer transitive `@types/node` versions

.github/workflows/coveralls.yaml
```diff
@@ -10,21 +10,24 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [18.x]
+        node-version: [22.x]
 
     steps:
       - uses: actions/checkout@v4
 
+      - uses: pnpm/action-setup@v6
```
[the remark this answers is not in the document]
karlhorky: addressed in 2c29cec71fa3aaaf966a8d965f365051a717b2a6 and a9cdd38b65d5e3ff00ddf660e863e5729aa7a236

package.json
```diff
@@ -133,8 +134,8 @@
   },
   "devEngines": {
     "packageManager": {
-      "name": "yarn",
-      "version": "1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e"
+      "name": "pnpm",
+      "version": "11.0.8+sha512.4c4097e1dd2d42372c4e7fa5a791ff28fc75a484c7ac192e64b1df0fdef17594ba982f9b4fed9adfb3c757846f565b799b2763fb3733d1de1bcb82cf46684912"
```
nene, line 138: Do we really need to lock ourselves to a very specific version? Why not just `version: ^11.0.8`

Like, I installed `pnpm` and now it complains that my `pnpm` version is too new.

The reason I was OK with `yarn` being a specific version was because that version has been the latest `1.x` release for a long time now and is unlikely to get any updates.

.github/workflows/webpack.yaml
```diff
@@ -9,30 +9,41 @@ jobs:
...
+        node-version: [22.x]
 
     steps:
       - uses: actions/checkout@v4
 
+      # Work around pnpm/action-setup v6 resolving the wrong version from package.json
+      # https://github.com/pnpm/action-setup/issues/227
+      - name: Read pnpm version
+        id: pnpm-version
+        run: 'echo "version=$(jq --raw-output ''.devEngines.packageManager.version | split("+")[0]'' package.json)" >> "$GITHUB_OUTPUT"'
```
nene, line 21: This seems related to that very specific version. Will the need for this workaround still stay when we relax the pnpm version requirements?

pnpm-workspace.yaml
```diff
@@ -0,0 +1,37 @@
...
+
+# Prevent installing packages published less than 7 days ago
+# https://pnpm.io/settings#minimumreleaseage
+minimumReleaseAge: 10080
+minimumReleaseAgeStrict: true
+minimumReleaseAgeIgnoreMissingTime: false
+
+# Fail when another package manager is used
+# https://pnpm.io/settings#pmonfail
+pmOnFail: error
```
nene: I guess that's another place that enforces this strict pnpm version check.

The comment is misleading though, other package managers won't be respecting this pnpm config file.

pnpm-workspace.yaml
```diff
@@ -0,0 +1,37 @@
...
+trustPolicy: no-downgrade
+trustPolicyExclude:
+  - eslint-config-prettier@8.10.2
+  - eslint-plugin-prettier@4.2.5
+  - semver@5.7.2 || 6.3.1
+  - undici-types@6.19.8
+
+# Fail before scripts when dependencies are out of sync
+# https://pnpm.io/settings#verifydepsbeforerun
+verifyDepsBeforeRun: error
```
nene: There are lots of settings here. The comments aren't really telling me though what I'd really want to know. What I'm really interested in is why have these specific settings been chosen in this repository?

For example `sideEffectsCache: false`. The docs tell me that:

> You may want to disable this setting if:
> -  The install scripts modify files outside the package directory (pnpm cannot track or cache these changes).
> -  The scripts perform side effects that are unrelated to building the package.

I don't think either of these are true. But perhaps I'm mistaken.

nene: In general, maybe let's not try too hard with all these additional pnpm configurations.

The main goal of this PR should be to switch from yarn to pnpm, while keeping everything working as close as possible to how it was before. Additional pnpm configurations can be layered on top of that later on.

package.json
karlhorky: > Like, I installed `pnpm` and now it complains that my `pnpm` version is too new.

What error message did you get? I think I probably misconfigured [`pmOnFail`](https://pnpm.io/settings#pmonfail) to be `'error'` instead of just leaving it with the default `'download'` - I reverted this now:

- 4f4df0255b4d7a8d89de268bd7508d2622942cc0

The default `'download'` will just download and use the correct version of pnpm.

> Do we really need to lock ourselves to a very specific version? Why not just `version: ^11.0.8`

Mostly for security / exact reproducibility across all environments

In my projects, I also configure pinned versions. To avoid updating all versions myself, what I do is set up [Renovate](https://github.com/marketplace/renovate) to create PRs with upgrades for my dependencies (similar to Dependabot, but better). This also upgrades the package manager version.

.github/workflows/webpack.yaml
karlhorky: The issue https://github.com/pnpm/action-setup/issues/227 appears to be related to `pnpm/action-setup@6` sometimes not reading the version correctly

Why a workaround?

1. I used the workaround instead of downgrading to avoid upgrade churn in future, since v6 does some things differently
2. I used the workaround instead of relaxing the version because of [the reproducibility argument](https://github.com/sql-formatter-org/sql-formatter/pull/949#:~:text=Mostly%20for%20security%20/%20exact%20reproducibility%20across%20all%20environments)

pnpm-workspace.yaml
karlhorky: Yeah I think this config setting was a mistake, I reverted it:

- https://github.com/sql-formatter-org/sql-formatter/pull/949#discussion_r3244174768

pnpm-workspace.yaml
karlhorky: I somehow missed this `sideEffectsCache: false` - it must have slipped in during iteration, in the name of security (to avoid reusing cached built artifacts and instead rebuild them). But as far as I can tell, this is unrelated to security.

I've removed `sideEffectsCache: false` and `strictDepBuilds: true` (this is default now in pnpm v11):

- 249ea2ba

I've reviewed all of the settings again now:

- `blockExoticSubdeps`: block other non-registry sources of transitive deps 👍 
- `minimumReleaseAge` + friends: min 7 days old, prevents supply chain attacks 👍 
- `trustPolicy` + friends: prevent package trust downgrades (eg. provenance disappearing in following releases) 👍 
- `verifyDepsBeforeRun`: error if `node_modules` not up to date when running scripts 👍

package.json
nene, line 138: I'm still not succeeding in `pnpm install`. I'm getting the following error:

```
[ERROR] This project is configured to use 11.0.8+sha512.4c4097e1dd2d42372c4e7fa5a791ff28fc75a484c7ac192e64b1df0fdef17594ba982f9b4fed9adfb3c757846f565b799b2763fb3733d1de1bcb82cf46684912 of pnpm. Your current pnpm is v11.1.2
If you want to bypass this version check, you can set the "pmOnFail" configuration to "warn" or "ignore" (e.g. via --pm-on-fail=ignore). If using "devEngines.packageManager", you can set its "onFail" to "warn" or "ignore"
```

It seems to me that contrary to what pnpm docs say the `download` is not actually the default, because it works when I do:

```sh
pnpm install --pm-on-fail=download
```

Similarly when I explicitly set `pmOnFail: download` in the config file.

nene: Additionally, when I run with this `download` option enabled, I get the `pnpm-lock.yaml` file updated, because the pnpm 11.0.8 gets added to it as a dependency.

pnpm-workspace.yaml
```diff
@@ -0,0 +1,25 @@
...
+minimumReleaseAgeIgnoreMissingTime: false
+
+# Prevent package trust metadata from downgrading on updates
+# https://pnpm.io/settings#trustpolicy
+trustPolicy: no-downgrade
+trustPolicyExclude:
+  - eslint-config-prettier@8.10.2
+  - eslint-plugin-prettier@4.2.5
+  - semver@5.7.2 || 6.3.1
+  - undici-types@6.19.8
```
nene: I wonder about these specific packages. What's the deal with them?

When I remove these entries, the install works fine. But for some reason we want to allow these packages to be replaced with lower trust versions in the future. Why?

package.json
karlhorky: oh, interesting. I wonder why the default `'download'` behavior is not being used here. maybe this is a pnpm bug

if you give me the following details, I'll open a pnpm issue:

1. what is the output of `pnpm config get pmOnFail`? (\*without\* the config setting present in `pnpm-workspace.yaml`)
2. are there any relevant environment variables set? eg. `pnpm_config_pm_on_fail` or `COREPACK_*`
3. How was pnpm 11.1.2 installed or invoked? eg. npm global install, standalone installer, Corepack, Volta, mise, Homebrew. even if not installed through Corepack,  what are the outputs of `corepack --version`, `command -v pnpm` and `ls -al "$(command -v pnpm)"`?
4. which Node.js version and OS are you using? if OS is Linux, what distro / version?

once I open the issue, I can also configure `pmOnFail: download` in `pnpm-workspace.yaml` and mark it as a workaround while the pnpm bug is fixed, with a link to the issue

pnpm-workspace.yaml
karlhorky: I added comments (which I also add in my own projects) to be clear why these versions are there:

- b1d7dc65

```yaml
# Prevent package trust metadata from downgrading on updates
# https://pnpm.io/settings#trustpolicy
trustPolicy: no-downgrade
trustPolicyExclude:
  # first version with provenance: eslint-config-prettier@10.1.4 https://github.com/prettier/eslint-config-prettier/issues/339#issuecomment-3090250555
  - eslint-config-prettier@8.10.2
  # first version with provenance: eslint-plugin-prettier@5.3.1 https://github.com/prettier/eslint-plugin-prettier/pull/757
  - eslint-plugin-prettier@4.2.5
  # first version with provenance: semver@7.5.1 https://github.com/npm/node-semver/issues/838#issuecomment-3698312268
  - semver@5.7.2 || 6.3.1
  # first version with provenance: undici-types@6.13.0 https://github.com/nodejs/undici/issues/4666
  - undici-types@6.19.8
```

the short version: previous versions before these versions had provenance

pnpm-workspace.yaml
```diff
@@ -0,0 +1,29 @@
...
+trustPolicy: no-downgrade
+trustPolicyExclude:
+  # first version with provenance: eslint-config-prettier@10.1.4 https://github.com/prettier/eslint-config-prettier/issues/339#issuecomment-3090250555
+  - eslint-config-prettier@8.10.2
+  # first version with provenance: eslint-plugin-prettier@5.3.1 https://github.com/prettier/eslint-plugin-prettier/pull/757
+  - eslint-plugin-prettier@4.2.5
+  # first version with provenance: semver@7.5.1 https://github.com/npm/node-semver/issues/838#issuecomment-3698312268
+  - semver@5.7.2 || 6.3.1
+  # first version with provenance: undici-types@6.13.0 https://github.com/nodejs/undici/issues/4666
+  - undici-types@6.19.8
```
karlhorky: @nene an alternative for all of these packages (since they're all older than 60 days) is to ignore the trust policy downgrade ([provenance disappearing on new versions](https://github.com/pnpm/pnpm/issues/10202#issuecomment-3552462881)) using [`trustPolicyIgnoreAfter`](https://pnpm.io/settings#trustpolicyignoreafter)

I'm using this on some of our projects:

```suggestion
# Prevent trust level decreases with packages newer than 60 days
# to mitigate supply chain risks
# https://pnpm.io/settings#trustpolicy
trustPolicy: no-downgrade
trustPolicyIgnoreAfter: 86400
```

pnpm-workspace.yaml
karlhorky: an alternative would be using [`trustPolicyIgnoreAfter`](https://pnpm.io/settings#trustpolicyignoreafter):

- https://github.com/sql-formatter-org/sql-formatter/pull/949#discussion_r3247475428

package.json
nene: > what is the output of `pnpm config get pmOnFail`?

Well, the output is the same error:

```
[ERROR] This project is configured to use 11.0.8+sha512.4c4097e1dd2d42372c4e7fa5a791ff28fc75a484c7ac192e64b1df0fdef17594ba982f9b4fed9adfb3c757846f565b799b2763fb3733d1de1bcb82cf46684912 of pnpm. Your current pnpm is v11.1.2
If you want to bypass this version check, you can set the "pmOnFail" configuration to "warn" or "ignore" (e.g. via --pm-on-fail=ignore). If using "devEngines.packageManager", you can set its "onFail" to "warn" or "ignore"
```

Even calling `pnpm help` gives me this same error.

> are there any relevant environment variables set? eg. `pnpm_config_pm_on_fail` or `COREPACK_*`

No.

> How was pnpm 11.1.2 installed or invoked? eg. npm global install, standalone installer, Corepack, Volta, mise, Homebrew. even if not installed through Corepack,  what are the outputs of `corepack --version`, `command -v pnpm` and `ls -al "$(command -v pnpm)"`?

Using `brew install pnpm`

> which Node.js version and OS are you using?

- Node 22.22.3
- MacOS 26.3

I also tried whether the Fish shell might be the culprit, but no, same error with Bash.

pnpm-workspace.yaml
nene: BTW, what is "provenance"? It's a terminology I haven't come across before.

package.json
karlhorky: thanks!

2 things still open:

1. try running the `pnpm config get pmOnFail` in a different directory (or temporarily remove `devEngines.packageManager` from `package.json`) 

and this part of number 3:

> even if not installed through Corepack, what are the outputs of `corepack --version`, `command -v pnpm` and `ls -al "$(command -v pnpm)"`?

pnpm-workspace.yaml
karlhorky: it's an npm security mechanism, to allow verifying how a package was built:

- https://docs.npmjs.com/generating-provenance-statements

package.json
nene: If I run it one some random directory I get as result:

```
$ pnpm config get pmOnFail
undefined
```

Same, when I remove the `devEngines.packageManager` config.

```
$ corepack --version
0.34.6

$ command -v pnpm
/opt/homebrew/bin/pnpm

$ ls -al "$(command -v pnpm)"
lrwxr-xr-x  1 nene  admin  30 May 14 22:26 /opt/homebrew/bin/pnpm@ -> ../Cellar/pnpm/11.1.2/bin/pnpm
```

pnpm-workspace.yaml
nene: I don't really understand this.

If we're currently using `eslint-config-prettier@8.10.2` and that doesn't have provenance, how would it be possible to downgrade the provenance of this package? Is there some way of having negative provenance? And why would we want that? I'm not sure I really understand this provenance thing at all.

karlhorky: pnpm's `trustPolicy` is time-based, not semver-based

as soon as a version with provenance is published (in this case `eslint-config-prettier@10.1.4`), then any future versions that are published (in this case `eslint-config-prettier@8.10.2`) are marked as a trust downgrade

it does make sense, if you think through a scenario where: 

1. attacker gets publish access to popular library, which has provenance on latest major
2. attacker intentionally publishes older version (also with high download count), on a major version which has never had provenance
3. to a casual observer, this new version may not appear dangerous
4. pnpm `trustPolicy` errors about it

some community members find this counterintuitive (as did I when I first heard it). there's more discussion over in the issue I linked on the "provenance disappearing on new versions" link above: 

- https://github.com/pnpm/pnpm/issues/10202#issuecomment-3552462881

karlhorky: `trustPolicyIgnoreAfter` works around it by ignoring such violations of new versions after a certain period of time. I set it to 60 days in the code suggestion above, which is probably plenty of time for the static analysis security scanners and the community to find real malicious versions and for those versions to be removed from npm

package.json
karlhorky: Ok, opened the issue:

- https://github.com/pnpm/pnpm/issues/11676

pnpm-workspace.yaml
nene: Thanks for the explanation.

In general I agree it's better to use this `trustPolicyIgnoreAfter` than singling out those specific versions.

nene (commented):
In general I'm all good with this.

The single showstopper for me is the `pmOnFail` issue. If we could just add to the config file:

```
pmOnFail: download
```

Then I would be fine with merging this. Even when this gets fixed in pnpm, the old pnpm versions with this issue will still be around for a while, troubling people who would like to check out this library.

karlhorky:
@nene Good point, I forgot that I wanted to add that as a workaround - done:

- dcee63f96d576ed828dd5af67d26ee9a3728e2db

nene (approved):
Awesome.

karlhorky:
> > The single showstopper for me is the `pmOnFail` issue. If we could just add to the config file:
> > 
> > ```
> > pmOnFail: download
> > ```
> 
> Good point, I forgot that I wanted to add that as a workaround - done:

I've upgraded to pnpm 11.9.0 and reverted the workaround here:

- https://github.com/sql-formatter-org/sql-formatter/pull/957

PR #952, opened 2026-06-03
fix: keep leading block comment of clause item on its own line

sarathfrancis90:
A block comment before the first item of a clause (e.g. `SELECT /* c */ foo FROM tbl`) was formatted inline on the first pass but moved to its own line on the second — so formatting wasn't idempotent.

The inline-vs-standalone decision used only the comment's `precedingWhitespace`. Since the formatter emits a newline before a clause's first item, the second pass sees that structural newline and flips the decision. Now a block comment is also treated as standalone when it's already at the start of the line in the output being built, so the first pass already matches the fixpoint. Genuinely-inline comments are unaffected.

Added tests covering comments in more positions (between items, after FROM/JOIN, in UPDATE/SET) to confirm it generalizes.

## Release Notes

* **Bug Fixes**
  * Improved block comment formatting to intelligently determine placement based on content and position context. Block comments now consistently appear on their own lines when appropriate, with enhanced support for idempotent formatting to ensure results remain stable across multiple formatting passes.

### DIFF
--- a/src/formatter/ExpressionFormatter.ts
+++ b/src/formatter/ExpressionFormatter.ts
@@ -378,7 +378,7 @@ export default class ExpressionFormatter {
   }
 
   private formatBlockComment(node: BlockCommentNode | DisableCommentNode) {
-    if (node.type === NodeType.block_comment && this.isMultilineBlockComment(node)) {
+    if (node.type === NodeType.block_comment && this.isStandaloneBlockComment(node)) {
       this.splitBlockComment(node.text).forEach(line => {
         this.layout.add(WS.NEWLINE, WS.INDENT, line);
       });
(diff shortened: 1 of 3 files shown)


nene:
Thanks. This looks like a reasonable fix. At least on first look. I need to ponder it some more before merging.

I would definitely appreciate not reading through a massive amount of AI-generated descriptions of this whole thing.

I also would like to have some more tests for this. Currently we're only checking for one scenario: a comment right after `SELECT` keyword. But how about comments in other places that should either behave the same or not behave the same, for example:

```sql
SELECT foo /*comment */, /* comment */ bar FROM /*comment */ tbl1 JOIN /* comment */ tbl2;
UPDATE /* comment */ tbl /*comment */ SET /* comment */ x = 10;
```

These are just random examples. I'm trying to come up with scenarios that could possibly cause problems. But mainly we should just have a few more tests to ensure that this fix does indeed work in general, not just for one single case.

sarathfrancis90:
Good call — added a test that runs comments in more positions (between items, after FROM/JOIN, in UPDATE/SET) through formatting twice to confirm they're all idempotent; full suite passes. Also trimmed the description. Thanks for taking a look!

nene:
OK. Thanks. It's merged now, and released in 15.8.1

PR #953, opened 2026-06-12
fix: keep spaces around - operator in dialects with dashed identifiers

sarathfrancis90:
With `denseOperators: true` on BigQuery, a subtraction like `a - b` comes out as `a-b`. That's not just ugly: BigQuery allows dashes inside identifiers, so the densed output re-parses as a single identifier `a-b`, and `format(format(sql))` no longer matches `format(sql)`. `a - foo(y)` is worse — it becomes `a-foo (y)`, turning the expression into an identifier followed by a call.

The fix is to leave the `-` operator spaced in dialects that allow dashed identifiers (currently just BigQuery). I derived that flag from the existing `identChars.dashes` tokenizer option so there's a single source of truth. Numeric cases like `1 - 2` were already safe (the `-` is parsed into the literal), so this only affects the identifier case.

The shared dense-operator test in `operators.ts` was actually asserting the broken `foo-bar` output for BigQuery; I scoped that to the dashed-identifier dialects and added a BigQuery regression test.

## Release Notes

* **Bug Fixes**
  * Fixed formatting of the `-` operator in dialects that support dashes in identifiers, such as BigQuery. When dense operator formatting is enabled, spaces around the `-` operator are now properly preserved to prevent expressions from being incorrectly merged into dashed identifiers.

### DIFF
--- a/src/dialect.ts
+++ b/src/dialect.ts
@@ -34,15 +34,17 @@ export const createDialect = (options: DialectOptions): Dialect => {
 
 const dialectFromOptions = (dialectOptions: DialectOptions): Dialect => ({
   tokenizer: new Tokenizer(dialectOptions.tokenizerOptions, dialectOptions.name),
-  formatOptions: processDialectFormatOptions(dialectOptions.formatOptions),
+  formatOptions: processDialectFormatOptions(dialectOptions),
 });
 
-const processDialectFormatOptions = (
-  options: DialectFormatOptions
-): ProcessedDialectFormatOptions => ({
+const processDialectFormatOptions = ({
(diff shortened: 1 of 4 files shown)


nene:
Hi, sorry for the delay.

This is probably the best pull request I've received over the years. Very well thought through fix for this quirky bug. Thanks a lot.

nene:
Released in 15.8.2

karlhorky:
@coderabbitai why did you write this?

> Docstring coverage is 0.00% which is insufficient. The required threshold is 80.00%.

and can you suppress it for all future reviews?

PR #956, opened 2026-06-26
fix: keep a space when densing "-" before a negative operand

sarathfrancis90:
While fuzzing `format(format(x)) === format(x)`, I found a case where `denseOperators` turns valid SQL into something that means something different.

When a `-` operator is immediately followed by a negative operand, dense mode drops both spaces and collapses them into `--`:

```js
format("SELECT 1 - -1 AS x", { denseOperators: true })
// => "SELECT\n  1--1 AS x"
```

`--` starts a line comment, so `--1 AS x` is now a comment — the negation and the alias are silently dropped, and reformatting the output changes it again to `1 --1 AS x`. It happens both for a unary minus (`a - -b`) and for negative number literals (`1 - -1`). (BigQuery already keeps `-` spaced after #953, so it is unaffected.)

The fix is in the layout: if a token starting with `-` would land directly after one ending with `-`, keep a single space so the `--` marker can never form. Everything else stays dense (`a- -b`) and formatting is idempotent again.

Added a regression test to the shared operators feature test. Full test suite, typecheck, lint, prettier and build all pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed SQL formatting in dense mode to prevent `- -` sequences (e.g. `a - -b`, `1 - -1`) from being accidentally turned into `--` line comments.
  * Ensured `denseOperators: true` keeps the needed spacing boundaries so formatting is stable across re-runs. 
* **Tests**
  * Added coverage to verify the behavior and formatting idempotency for `a - -b` and `1 - -1`.

### DIFF
--- a/src/formatter/Layout.ts
+++ b/src/formatter/Layout.ts
@@ -57,9 +57,20 @@ export default class Layout {
           this.items.push(WS.SINGLE_INDENT);
           break;
         default:
+          // Don't glue a token starting with "-" directly onto one ending with
+          // "-": that forms "--", which re-parses as a line comment and
+          // swallows the rest of the line (e.g. densing "a - -b" into "a--b").
+          if (typeof item === 'string' && item.startsWith('-') && this.lastTokenEndsWith('-')) {
+            this.items.push(WS.SPACE);
+          }
           this.items.push(item);
       }
     }
+  }
+
+  private lastTokenEndsWith(suffix: string): boolean {
+    const lastItem = last(this.items);
+    return typeof lastItem === 'string' && lastItem.endsWith(suffix);
   }
 
   private trimHorizontalWhitespace() {
--- a/test/features/operators.ts
+++ b/test/features/operators.ts
@@ -40,6 +40,17 @@ export default function supportsOperators(
     });
   });
 
+  it('does not glue a - in front of a negative operand into a line comment in dense mode', () => {
+    // "a - -b" / "1 - -1" must not be densed into "a--b" / "1--1": the "--"
+    // would re-parse as a line comment and silently swallow the rest of the line.
+    ['SELECT a - -b', 'SELECT 1 - -1'].forEach(sql => {
+      const result = format(sql, { denseOperators: true });
+      expect(result).not.toContain('--');
+      // ...and formatting stays idempotent.
+      expect(format(result, { denseOperators: true })).toBe(result);
+    });
+  });
+
   (cfg.logicalOperators || ['AND', 'OR']).forEach(op => {
     it(`supports ${op} operator`, () => {
       const result = format(`SELECT true ${op} false AS foo;`);
--- a/test/sqlFormatter.test.ts
+++ b/test/sqlFormatter.test.ts
@@ -83,6 +83,19 @@ describe('sqlFormatter', () => {
     }).toThrow('commaPosition config is no more supported.');
   });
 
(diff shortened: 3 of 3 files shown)


test/features/operators.ts
```diff
@@ -40,6 +40,17 @@ export default function supportsOperators(
...
   });
 
+  it('does not glue a - in front of a negative operand into a line comment in dense mode', () => {
+    // "a - -b" / "1 - -1" must not be densed into "a--b" / "1--1": the "--"
+    // would re-parse as a line comment and silently swallow the rest of the line.
+    ['SELECT a - -b', 'SELECT 1 - -1'].forEach(sql => {
+      const result = format(sql, { denseOperators: true });
+      expect(result).not.toContain('--');
+      // ...and formatting stays idempotent.
+      expect(format(result, { denseOperators: true })).toBe(result);
```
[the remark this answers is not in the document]
nene: In general I agree with this Coderabbit suggestion. It's better to be explicit what the output should be (instead of just saying "it shouldn't contain --". And when we're testing the exact output, then there's no need to also add `.not.toContain('--')`.

Additionally, I would just write this as two separate tests instead of doing a loop. For just two test-cases the added complexity of the loop is IMHO not worth it. Better to keep the test code as straight-forward as possible, even at the cost of some verbosity/duplication.

src/formatter/Layout.ts
```diff
@@ -57,11 +57,22 @@ export default class Layout {
...
+          // swallows the rest of the line (e.g. densing "a - -b" into "a--b").
+          if (typeof item === 'string' && item.startsWith('-') && this.lastTokenEndsWith('-')) {
+            this.items.push(WS.SPACE);
+          }
           this.items.push(item);
       }
     }
   }
 
+  private lastTokenEndsWith(suffix: string): boolean {
```
nene, line 71: For naming consistency, should name this to `lastItemEndsWith()`. This class doesn't deal with tokens, it deals with LayoutItems.

src/formatter/Layout.ts
```diff
@@ -57,11 +57,22 @@ export default class Layout {
           this.items.push(WS.SINGLE_INDENT);
           break;
         default:
+          // Don't glue a token starting with "-" directly onto one ending with
+          // "-": that forms "--", which re-parses as a line comment and
+          // swallows the rest of the line (e.g. densing "a - -b" into "a--b").
+          if (typeof item === 'string' && item.startsWith('-') && this.lastTokenEndsWith('-')) {
```
nene, line 63: Isn't the type of `item` always a string in this default block?

nene (commented):
Looks like a good fix.

sarathfrancis90:
Thanks! Tightened the coverage along those lines — added an exact-output assertion (`SELECT\n  a- -b` / `1- -1`) so it also proves the left side stays dense, not just that no `--` forms. I put it in the default-dialect test rather than the shared cross-dialect one on purpose: dialects with dashed identifiers (BigQuery) intentionally keep the `-` spaced, so a fixed expected output there would be dialect-specific — the shared check stays on the `no --` + idempotence invariant that holds for every dialect.

nene:
Thanks.

I reorganized the tests a bit, as I think placing this test to the generic `sqlFormatter.test.ts` isn't really the right place. Instead I change the operators tests so that I can exclude the `-` operator from the list, and won't need to have a special treatment of BigQuery inside these generic tests. Instead the BigQuery specific tests for `-` operator now live inside `bigquery.test.ts`.

PR #958, opened 2026-06-30
fix(postgresql): keep spaces around OPERATOR() with denseOperators

sarathfrancis90:
I was trying out `denseOperators` on some Postgres queries and noticed it mangles the `OPERATOR(schema.op)` syntax:

```
format('SELECT foo OPERATOR(public.===) bar;', { language: 'postgresql', denseOperators: true })
// =>
// SELECT
//   fooOPERATOR(public.===)bar;
```

The spaces around the operator get stripped, so `foo` runs into the `OPERATOR` keyword and the result no longer parses. Unlike `+`/`=`/etc., this one is keyword-like and can't be glued to its operands, so I made it keep its surrounding spaces even in dense mode — same idea as the existing `-` exception for dialects with dashed identifiers.

Added a test under the existing OPERATOR() case in postgresql.test.ts.

### DIFF
--- a/src/formatter/ExpressionFormatter.ts
+++ b/src/formatter/ExpressionFormatter.ts
@@ -339,6 +339,11 @@ export default class ExpressionFormatter {
     // would otherwise re-parse as a single dashed identifier.
     if (text === '-' && this.dialectCfg.identifierDashes) {
       this.layout.add(text, WS.SPACE);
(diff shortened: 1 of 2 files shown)


nene (approved):
Great. Thanks.

nene:
Also, I invited you to be a full-time collaborator. No pressure :)

nene:
@sarathfrancis90 you should now have push access to the repository.

PR #959, opened 2026-07-18
fix: remove extra space after BETWEEN _ AND _ expression

sarathfrancis90:
I noticed the formatter emits two spaces after a BETWEEN predicate whenever something follows it on the same line. With default options:

```js
format('SELECT CASE WHEN foo BETWEEN 1 AND 2 THEN 3 END');
```

gives `WHEN foo BETWEEN 1 AND 2  THEN 3`. Same for `logicalOperatorNewline: 'after'` (`b BETWEEN 1 AND 2  AND`) and for `b BETWEEN 1 AND 2  IS TRUE`.

The cause is in `formatBetweenPredicate`: the sub-expression for the upper bound already ends with a trailing space, and the method then adds another `WS.SPACE` on top. It usually goes unnoticed because with the default `logicalOperatorNewline: 'before'` the following `WS.NEWLINE` trims the doubled space away. The lower bound already normalizes this with `WS.NO_SPACE, WS.SPACE`, so I did the same for the upper bound.

Two existing expectations had the extra space baked in (`test/features/between.ts` and `test/features/case.ts`); both are updated to the intended single space.

Tested with `pnpm run check` (typecheck, prettier, eslint, full jest suite) — all green. I also re-ran a format/re-format idempotency sweep over the repo's own test queries across every dialect and the option matrix; nothing changed apart from the removed space.

### DIFF
--- a/src/formatter/ExpressionFormatter.ts
+++ b/src/formatter/ExpressionFormatter.ts
@@ -222,7 +222,7 @@ export default class ExpressionFormatter {
     this.layout = this.formatSubExpression(node.expr1);
     this.layout.add(WS.NO_SPACE, WS.SPACE, this.showNonTabularKw(node.andKw), WS.SPACE);
     this.layout = this.formatSubExpression(node.expr2);
-    this.layout.add(WS.SPACE);
+    this.layout.add(WS.NO_SPACE, WS.SPACE);
   }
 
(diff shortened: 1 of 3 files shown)


nene (approved):
Awesome! Thanks.

PR #960, opened 2026-07-27
fix: keep DuckDB array-slice colon dense inside [ ]

binggao1230:
DuckDB's array-slice operator formats with a stray space: `foo[:5]` comes out as `foo[: 5]` and `baz[1:5]` as `baz[1: 5]`.

## Root cause

Commit f9575c0 ("Support ':' as struct key-value separator") removed `':'` from DuckDB's `alwaysDenseOperators`, so struct literals like `{'a': 1}` keep the space after the colon. As a side effect the array-slice colon, which travels the same generic operator-spacing path, also became spaced. The identical test still passes for PostgreSQL (which keeps `':'` dense and has no struct literals), and the DuckDB copy was skipped with a `// TODO: This currently conflicts with ":"-operator in struct literals`.

The colon is genuinely context-dependent in DuckDB — dense inside `[...]` (array slice), spaced inside `{...}` (struct key) — but `alwaysDenseOperators` is a flat, dialect-global list with no notion of context, so neither setting is correct for both.

## Fix

Make the `':'` spacing decision depend on the nearest enclosing bracket rather than being dialect-global: dense inside `[...]`, spaced otherwise. The bracket context flows through the recursive `ExpressionFormatter` (each parenthesis passes its own `openParen` down to its children), so a struct nested in a list — `[{'a': 1}]` — still gets the spaced colon because its nearest bracket is `{`.

Only the colon-followed-by-a-value case was visibly wrong (`foo[:5]`, `baz[1:5]`); `bar[1:]` and `zap[:]` already rendered correctly because the trailing space was absorbed by the closing bracket.

## Tests

- Un-skips the pre-existing `supports array slice operator` test (all four forms: `foo[:5]`, `bar[1:]`, `baz[1:5]`, `zap[:]`).
- Adds a test pinning both directions in one statement — `{'k': list[1:5]}` — where the struct colon stays spaced while the slice colon is dense.

The four `formats {} struct literal ...` and `formats prefix aliases` tests double as an over-fix guard: making `:` dense unconditionally passes the slice test but breaks all of them. Full `pnpm test` is green across every dialect.

## Note

`alwaysDenseOperators` has no context mechanism today, so I threaded the bracket context through the formatter rather than widening that config surface. If you'd prefer a different shape (e.g. recognising the slice colon at the grammar level), happy to adjust.

### DIFF
--- a/src/formatter/ExpressionFormatter.ts
+++ b/src/formatter/ExpressionFormatter.ts
@@ -43,6 +43,8 @@ interface ExpressionFormatterParams {
   params: Params;
   layout: Layout;
   inline?: boolean;
+  // openParen of the nearest enclosing bracket ("(", "[" or "{"), if any
+  enclosingBracket?: string;
 }
 
 export interface DialectFormatOptions {
@@ -76,13 +78,22 @@ export default class ExpressionFormatter {
   private inline = false;
   private nodes: AstNode[] = [];
   private index = -1;
+  private enclosingBracket?: string;
 
-  constructor({ cfg, dialectCfg, params, layout, inline = false }: ExpressionFormatterParams) {
+  constructor({
+    cfg,
+    dialectCfg,
+    params,
+    layout,
+    inline = false,
+    enclosingBracket,
+  }: ExpressionFormatterParams) {
     this.cfg = cfg;
     this.dialectCfg = dialectCfg;
     this.inline = inline;
     this.params = params;
     this.layout = layout;
+    this.enclosingBracket = enclosingBracket;
   }
 
   public format(nodes: AstNode[]): Layout {
@@ -194,7 +205,7 @@ export default class ExpressionFormatter {
   }
 
   private formatParenthesis(node: ParenthesisNode) {
-    const inlineLayout = this.formatInlineExpression(node.children);
+    const inlineLayout = this.formatInlineExpression(node.children, node.openParen);
 
     if (inlineLayout) {
       this.layout.add(node.openParen);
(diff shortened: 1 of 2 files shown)


nene:
Thanks for the PR. Sorry for the lack of attention. I'm currently pretty busy. Maybe I'll get to reviewing this on Sunday or next week.

src/formatter/ExpressionFormatter.ts
```diff
@@ -43,6 +43,8 @@ interface ExpressionFormatterParams {
   params: Params;
   layout: Layout;
   inline?: boolean;
+  // openParen of the nearest enclosing bracket ("(", "[" or "{"), if any
+  enclosingBracket?: string;
```
nene, line 47: Lets rename this to `enclosingParenthesis`. This way it aligns with the existing terminology of e.g. `ParenthesisNode.openParen`.

src/formatter/ExpressionFormatter.ts
```diff
@@ -346,6 +357,9 @@ export default class ExpressionFormatter {
       this.layout.add(text, WS.SPACE);
     } else if (this.cfg.denseOperators || this.dialectCfg.alwaysDenseOperators.includes(text)) {
       this.layout.add(WS.NO_SPACE, text);
+    } else if (text === ':' && this.enclosingBracket === '[') {
+      // Array slice colon (e.g. arr[1:5]) is dense; the key-value ":" keeps its space.
+      this.layout.add(WS.NO_SPACE, text);
```
nene, line 362: Nit: The comment `the key-value ":" keeps its space.` is a bit out of place here, as this line of code doesn't do anything with the key-value part. I would just delete it or move that comment inside the next section that deals with this.

nene (approved):
This looks good.

I added a few small suggestions. Would be nice if you could address them. If not, I'll tackle them by myself before merging in.

Thanks.

PR #961, opened 2026-07-28
fix: keep a block comment leading a later statement on its own line

MildlyMeticulous:
A block comment before a statement that follows another is formatted inline on the first pass but moved onto its own line on the second, so formatting isn't idempotent:

```js
format('SELECT 1; /* c */ COMMIT;');
// => "SELECT\n  1;\n\n/* c */ COMMIT;"
// reformatted => "SELECT\n  1;\n\n/* c */\nCOMMIT;"
```

Same shape as #952, one level up. That fix made a block comment standalone when `layout.isAtStartOfLine()`, but `formatAst` gives each statement its own `Layout`, so for a leading comment the layout is still empty and `isAtStartOfLine()` falls through to `return false`. Statements are joined with newlines, so every statement after the first does start on a line — the second pass sees that newline as the comment's `precedingWhitespace` and flips the decision.

`Layout` now takes whether its statement is preceded by others, and an empty layout reports `isAtStartOfLine()` accordingly. The first statement is unaffected, so a leading comment on the whole query still formats inline as before.

Found by fuzzing `format(format(x)) === format(x)` with comments injected at every token boundary: 14,427 variants over 14 dialects, 10 failures before, 0 after. It reproduces in 11 of 21 dialects and for any keyword-only statement — `COMMIT`, `ROLLBACK`, `BEGIN`, `END`, `GO`, `USE`, `VACUUM`, `ANALYZE`, `SAVEPOINT`. Statements with clauses are stable because the following keyword forces the newline anyway.

5,861 tests pass; tsc, eslint and prettier clean.


nene:
Thanks for the PR. Sorry for the lack of attention. I'm currently pretty busy. Maybe I'll get to reviewing this on Sunday or next week.

src/formatter/Layout.ts
```diff
@@ -135,7 +137,7 @@ export default class Layout {
       }
       return item === WS.NEWLINE || item === WS.MANDATORY_NEWLINE;
     }
-    return false;
+    return this.startsOnNewLine;
```
nene, line 140: Looks like the same fix can be achieved by just replacing this with:

```ts
return true;
```

But we're lacking on tests that would demonstrate a need for something more complex than that.

### Layout case

This is definitely OK fix inside the `Layout` class, which gets constructed to format every separate SQL statement. And each statement will start from a new line, including the first one.

We might want to add a test that checks what happens when formatting a block-comment before a first SQL statement, like:

```ts
  it('keeps first block comment on its own line', () => {
    const sql = '/* c1 */ SELECT 1;';
    const result = dedent`
      /* c1 */
      SELECT
        1;
    `;
    expect(format(sql)).toBe(result);
  });
```

### InlineLayout case

It's not so OK with the `InlineLayout` class, because that one is used to format expressions inside parenthesis, which never form a new line. But currently our tests don't cover this scenario well. We're missing a test like this:

```ts
  it('preserves inline block-comment', () => {
    const sql = 'SELECT (/* c */ 1 + 2);';
    const result = dedent`
      SELECT
        (/* c */ 1 + 2);
    `;
    expect(format(sql)).toBe(result);
  });
```

To make that test pass, we should just override the `isAtStartOfLine()`method in `InlineLayout` with one that always returns false.

nene (changes requested):
We don't really need to perform statement-counting to achieve this fix.

And we're lacking tests.

See the comment that outlines a simpler way to fix this problem.

PR #962, opened 2026-08-04
fix: keep a space between an operator and a following sign with denseOperators

spokodev:
## Problem

With `denseOperators`, a binary operator immediately followed by a unary `+`/`-` is glued onto it:

```js
format('SELECT 5 % -2', { language: 'postgresql', denseOperators: true })
// => "SELECT\n  5%-2"
```

PostgreSQL lexes a run of operator characters greedily, and an operator containing one of `~ ! @ # % ^ & |` keeps a trailing `+`/`-` as part of the operator. So `5%-2` is tokenized as `5 %- 2`, and `%-` is not a defined operator:

```
SELECT 5 % -2;   -- 1
SELECT 5%-2;     -- ERROR: operator does not exist: integer %- integer
```

The same happens for `^ # & | ~ ! @` and multi-character operators that contain them (for example `@>` becomes `@>-`). So `denseOperators` can turn a valid query into one that fails to parse or changes meaning.

The formatter already keeps a space for the analogous `-` before `-` case (which would otherwise form a `--` line comment and swallow the rest of the line). This is the same class of problem, so this change generalizes that guard.

## Fix

In `Layout`, when appending an item that starts with `+`/`-`, if the preceding item ends in an operator run that either ends in `-` (the existing `--` case) or contains one of `~ ! @ # % ^ & |`, keep a space. Operators that do not merge with a following sign (`* / + - << >> = < >`) are unaffected, so dense output like `5*-2` and `a=-1` is unchanged.

## Test

Added to `test/postgresql.test.ts`. The full suite passes (5842 tests).


src/formatter/Layout.ts
```diff
@@ -73,6 +75,21 @@ export default class Layout {
...
+    if (typeof lastItem !== 'string') {
+      return false;
+    }
+    const run = /[-+*/<>=~!@#%^&|]+$/u.exec(lastItem)?.[0];
+    if (!run) {
+      return false;
+    }
+    return (item.startsWith('-') && run.endsWith('-')) || /[~!@#%^&|]/u.test(run);
+  }
+
```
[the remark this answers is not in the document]
spokodev: Added `?` and backtick to both the trailing-operator run and the sensitive-character test in the latest push, and extended the test with the built-in jsonb `?` case (`data ? -1` now formats as `data? -1`). Verified on PostgreSQL 17 that an operator ending in `?` or containing a backtick merges with a following sign (`5 <?-2` -> `operator does not exist: integer <?- integer`). Full suite stays green.

src/formatter/Layout.ts
```diff
@@ -57,10 +57,12 @@ export default class Layout {
...
-          // Don't glue a layout item starting with "-" directly onto one ending with
-          // "-": that forms "--", which re-parses as a line comment and
-          // swallows the rest of the line (e.g. densing "a - -b" into "a--b").
-          if (item.startsWith('-') && this.lastItemEndsWith('-')) {
+          // Don't glue an item starting with "-"/"+" onto a preceding operator when
+          // the two would re-lex as one token: "-" onto "-" forms "--" (a line
+          // comment that swallows the rest of the line), and a sign onto an operator
+          // containing ~!@#%^&|`? forms a merged operator like "%-" or "@>-" that parses
+          // differently (e.g. densing "5 % -2" into "5%-2").
+          if (this.wouldMergeIntoOperator(item)) {
```
nene, line 66: This long comment would really be better rewritten as a comment on the `wouldMergeIntoOperator()` method, describing what that method does.

src/formatter/Layout.ts
```diff
@@ -73,6 +75,21 @@ export default class Layout {
     return typeof lastItem === 'string' && lastItem.endsWith(suffix);
   }
 
+  private wouldMergeIntoOperator(item: string): boolean {
```
nene, line 79: I find the name of this method to be kinda awkward. Definitely this code base doesn't contain any other function names starting with `would-` prefix.

I'd suggest inverting the boolean return value of this function and naming it something like `isItemSafeToAppend()`.

src/formatter/Layout.ts
```diff
@@ -73,6 +75,21 @@ export default class Layout {
...
+      return false;
+    }
+    const lastItem = last(this.items);
+    if (typeof lastItem !== 'string') {
+      return false;
+    }
+    const run = /[-+*/<>=~!@#%^&|`?]+$/u.exec(lastItem)?.[0];
+    if (!run) {
+      return false;
+    }
```
nene, line 90: I have no idea what's going on in here. What does the `run` variable mean?

I guess it's some sort of run of characters. But that doesn't really help me in understanding its purpose.

test/postgresql.test.ts
```diff
@@ -234,6 +234,25 @@ describe('PostgreSqlFormatter', () => {
...
+    `);
+    expect(format(`SELECT '[1,2]'::jsonb @> -1`, { denseOperators: true })).toBe(dedent`
+      SELECT
+        '[1,2]'::jsonb@> -1
+    `);
+    expect(format(`SELECT data ? -1 FROM t`, { denseOperators: true })).toBe(dedent`
+      SELECT
+        data? -1
+      FROM
+        t
```
nene, line 252: I counted 17 special characters in the regular expression. In this test we're only checking a few of them.

nene (changes requested):
A major problem with this pull request is that it's tackling an issue that's specific to PostgreSQL, but the implementation is written for all dialects.

To my knowledge this operator concatenation is really mainly an issue in PostgreSQL. Most SQL dialects don't support all these fancy operators. It really would be best if this fix was constrained to only target PostgreSQL.

spokodev:
You're right, thanks. Scoped it to the dialects that lex a run of operator characters as a single operator (PostgreSQL and Redshift), where `5 % -2` densed to `5%-2` re-parses as the operator `%-`. It's behind a new `operatorsCombine` dialect option; MySQL, standard SQL and the rest keep a fixed operator set, so they now dense `5 % -2` to `5%-2` unchanged (added a MySQL test covering that). The `--` line-comment guard is untouched and stays universal, so `a - -b` keeps its space everywhere. Full suite green.

spokodev:
All four points addressed.

**Naming.** `wouldMergeIntoOperator` is now `isItemSafeToAppend` with the boolean inverted, so the call site reads `if (!this.isItemSafeToAppend(item))`.

**The comment** moved off the `default:` branch and onto the method as a doc comment describing what it decides.

**`run`** is now `precedingOperatorChars` — the operator characters the new item would be written against.

**Coverage.** The test now runs over every character that makes PostgreSQL lex a run as one operator, rather than a sample: `~ !~ @> # % ^ & | ?`. Each is checked as `SELECT a <op> -1`; on `aa8efaef` all nine format as `a~-1`, `a%-1`, `a?-1` and so on, and with the fix all nine keep the space.

That is nine of the ten characters in the inner class. The tenth is the backtick: it is legal in a `CREATE OPERATOR` name, which is why it is in the regex, but the lexer never yields it as an operator (`SELECT a \` -1` fails to parse), so there is no way to exercise it from a format test. I left it in the character class and said so in a comment above the test rather than dropping it silently.

Also removed `lastItemEndsWith` — my first commit was its only caller and left it dead.

5852 passing, prettier, eslint and `tsc --noEmit` clean.

test/postgresql.test.ts
spokodev: Answering here too — I replied in the conversation on 7 Sep and this thread stayed open, so the signal never reached you.

The test now runs over every character that makes PostgreSQL lex a run as a single operator rather than a sample: `~ !~ @> # % ^ & | ?`, each as `SELECT a <op> -1`.

Measured just now by disabling the guard (`isItemSafeToAppend` forced to `true`) and running `test/postgresql.test.ts`: **all nine fail**, and 10 with the general case. With the guard in place all 11 pass.

That is nine of the ten characters in the inner class. The tenth is the backtick: it is legal in a `CREATE OPERATOR` name, which is why it is in the class, but the lexer never yields it as an operator — `format('SELECT a ` -1')` gives `Parse error: Unexpected "` -1" at line 1 column 10`, so it cannot be reached from a format test. I left it in the class and said why above the test rather than dropping it silently.

src/formatter/Layout.ts
```diff
@@ -57,20 +57,42 @@ export default class Layout {
...
+    // The operator characters the new item would be written against.
+    const precedingOperatorChars = /[-+*/<>=~!@#%^&|`?]+$/u.exec(lastItem)?.[0];
+    if (!precedingOperatorChars) {
+      return true;
+    }
+    if (item.startsWith('-') && precedingOperatorChars.endsWith('-')) {
+      return false;
+    }
+    return !(this.operatorsCombine && /[~!@#%^&|`?]/u.test(precedingOperatorChars));
   }
```
nene, line 96: I have trouble understanding the logic in this function. The function has a long comment, which might be a sign that a lot of explanation is needed to explain what this code does.

Some observations:

- it starts with check `item.startsWith('+')` that wasn't part of the original code and gets applied to all dialects. When I comment this out, no tests break.
- the logic dealing with `operatorsCombine === true` case is not clearly separated from the logic for the base case.
- There are lots of conditional statements, and the later ones depend on the earlier ones. So to understand what happens on the last line, one really needs to grasp what happened before on the preceding lines.

src/formatter/Layout.ts
```diff
@@ -25,7 +25,7 @@ export type LayoutItem = WS.SPACE | WS.SINGLE_INDENT | WS.NEWLINE | WS.MANDATORY
 export default class Layout {
   private items: LayoutItem[] = [];
 
-  constructor(public indentation: Indentation) {}
+  constructor(public indentation: Indentation, private operatorsCombine = false) {}
```
nene: There's now this new flag `operatorsCombine` on the `Layout` class.

But there is `InlineLayout` class which extends from `Layout`. It seems that the context of `InlineLayout` that flag is now always false. I have a suspicion that this might cause problems. Maybe you have done it intentionally so. Maybe not. I don't know. But it looks suspicious.

FYI: `InlineLayout` is used for formatting a sequence of tokens between a pair of parenthesis, trying to fit them on a single line.

test/mysql.test.ts
```diff
@@ -114,4 +114,12 @@ describe('MySqlFormatter', () => {
...
     `);
   });
+
+  it('does not space a sign after an operator in dense mode', () => {
+    expect(format('SELECT 5 % -2, 5 & -2', { denseOperators: true })).toBe(dedent`
+      SELECT
+        5%-2,
+        5&-2
+    `);
+  });
```
nene: Why this test for MySQL in particular?

I think what you really intended was to write a test for other dialects that don't have the operator-combining problem. But for some reason you just added one test for one specific other dialect.

We should do better than that.

test/postgresql.test.ts
```diff
@@ -234,6 +234,39 @@ describe('PostgreSqlFormatter', () => {
...
+  // as an operator, so it is not reachable from here.
+  it.each(['~', '!~', '@>', '#', '%', '^', '&', '|', '?'])(
+    'keeps a space between the operator %s and a following sign with denseOperators',
+    operator => {
+      expect(format(`SELECT a ${operator} -1`, { denseOperators: true })).toBe(dedent`
+        SELECT
+          a${operator} -1
+      `);
+    }
+  );
```
nene: Like with a previous case of single-test-for-MySQL, here's a similar single test for PostgreSQL. But you enable the `operatorsCombine` flag also for Redshift. So this thing is not only specific for PostgreSQL.

Interestingly Redshift doesn't quite support such a wide range of operators as PostgreSQL. (Or at least SQL Formatter doesn't support many.)

That brings to my mind that inside `isItemSafeToAppend()` method we have these hard-coded regexes to match a wide range of characters. Some of these might not be relevant for Redshift. They might not be harmful, but they might not cause conflicts in the same way. Frankly I really don't know. I'd like to hear your thoughts on this.

Alternative proposal: We could leave Redshift out of this pull request, and focus solely on PostgreSQL. You could then come back with another pull request to solve the problem also for Redshift. But up to you.

test/postgresql.test.ts
```diff
@@ -234,6 +234,39 @@ describe('PostgreSqlFormatter', () => {
...
+      SELECT
+        '[1,2]'::jsonb@> -1
+    `);
+    expect(format(`SELECT data ? -1 FROM t`, { denseOperators: true })).toBe(dedent`
+      SELECT
+        data? -1
+      FROM
+        t
+    `);
+  });
```
nene: Here we're testing the `#`, `%`, `^`, `@>` and `?` operators twice with two separate tests.

nene (changes requested):
Thanks for the improvements.

This is better now. However I still have several concerns.

src/formatter/Layout.ts
spokodev: Rewritten as three independent checks, with each hazard explained next to the regex that describes it. No condition depends on the one before it, and `operatorsCombine` is its own branch.

On `item.startsWith('+')`: it only ever fired under `operatorsCombine`, but you had to read down to the last line to see that, which is a fair complaint. It is reachable rather than dead, though: `%+` is as valid an operator name as `%-`, so `5 % +2` densed to `5%+2` re-parses as the operator `%+`. It was simply untested. It is now -- narrowing the regex to `/^[-]/u` fails 15 tests.

src/formatter/Layout.ts
spokodev: Passed through now, and `Layout` no longer defaults it, so neither construction site can get `false` by accident.

It doesn't change any output, though, and I'd rather say that than imply it fixed a bug: `formatParenthesis` splices the inline layout's items back into the outer layout, which applies the guard again. Setting the flag back to `false` after this change leaves all 6233 tests green.

There is a real gap next to it that the flag does not reach. `addToLength` measures the items passed to `add()`, so it never counts the space the base class inserts. On master, with `denseOperators` and the default `expressionWidth: 50`, `SELECT * FROM t WHERE (aaa...aaa - -1 - -2)` (43 a's, mysql) renders 51 characters inside the parens. This PR makes PostgreSQL reach the same path. Fixing it means measuring what was appended rather than what was passed in, which reflows more than this PR should -- separate PR or an issue, whichever you prefer.

test/mysql.test.ts
spokodev: Agreed, one arbitrary dialect proved nothing. Moved into `supportsOperators()`, so every dialect that doesn't combine operators now asserts it for each operator it declares: 363 cases across 18 dialects, suite time unchanged.

Control that they bite: setting `operatorsCombine: true` on MySQL turns 9 of them red.

test/postgresql.test.ts
spokodev: Taking your proposal -- Redshift is out, the dialect file is identical to master again, and it gets its own pull request.

To answer the question: 9 of the 12 operators sql-formatter declares for Redshift contain one of ``~!@#%^&|`?`` (``^ % @ |/ ||/ & | ~ ||``), so for those the hazard is the same as in PostgreSQL; only `<<`, `>>` and `::` are untouched by the regex. Redshift forked from PostgreSQL 8.0.2, which already had that lexer rule, so the character set isn't wrong for Redshift -- just wider than its tokenizer can produce.

One thing worth flagging: the new cross-dialect test would otherwise assert `1 % -2` -> `1%-2` for Redshift, writing the behaviour I believe is wrong into the suite. So Redshift is marked as sign-absorbing there and asserts nothing until the follow-up.

test/postgresql.test.ts
spokodev: Removed -- it was a subset of the other test. There is now one parameterized test for the operators that absorb a sign and one for the operators that don't, with each operator appearing once.

spokodev:
Pushed, rebased onto master: the guard rewrite, Redshift dropped, and the tests.

`pnpm run check` is green locally, 6233 tests. The CI run on this push shows "action required", so it needs a click from you to start.

The rewrite is behaviour-preserving, and I measured that rather than assuming it: formatting the same 11392 queries before and after (77 operator strings x both signs x 4 shapes x `denseOperators` on and off x two `expressionWidth`s, in postgresql, redshift, mysql and standard sql) gives byte-identical output in every dialect except Redshift, where all 512 differences are the ones removed on purpose.

PR #963, opened 2026-08-08
Avoid quadratic accumulation when parsing long SQL expression lists

### LINKED ISSUE #840: Memory issue with large queries
**Describe the bug**
When formatting a large query containing thousands of ids (5,000-10,000+), the formatter uses a very large amount of memory, often OOMing the process.

**Expected behavior**
While this is a rather large query to deal with, many query engines can handle it fine. The memory consumption to format a <100kB query should not consume GBs of RAM

**Actual behavior**
The formatting function consumed RAM of nearly 50,000x the original query length

**Usage**

Here's an example script which generates two similar types of queries which both run into the issue:
```
const values = Array(10000).fill('myid')

// Large "in" clause
const sql = `SELECT * FROM my_table WHERE col in (${values.map((v) => "'" + v + "'").join(",")})`
// Large number of "or" clauses
// const sql = `SELECT * FROM my_table WHERE (${values.map((v) => "col = '" + v + "'").join(" OR ")})`
const {format} = require('sql-formatter')
const formatted = format(sql, {language: 'sql'})
console.log(Math.round(process.memoryUsage().heapUsed / 1024 / 1024) + " MB")
```

The first query uses 1.5GB of memory, and the second OOMs with 4GB of memory

arc-oai:
## Summary

- Replace repeated array concatenation in SQL expression lists with persistent linked lists and defer array materialization until completed expressions are needed.
- Preserve grammar ambiguity handling and formatted output while covering large `IN` lists, long `OR` chains, and wide `SELECT` lists.
- Re-enable the existing memory regression for #840.

## Why

Nearley expands `free_form_sql:*` into an append postprocessor equivalent to `items.concat([item])`. Repeatedly copying every prefix produces quadratic allocation. For a representative 1,000-value `IN` query, these generated repetition postprocessors copied 1,997,014 prefix entries before this change and 10 afterward. Final AST construction still performs the necessary linear work: 3,004 list entries are materialized and another 3,000 entries are copied into completed arrays. Mutating the shared array with `.push()` is unsafe because alternative parses share prefixes.

`free_form_sql` remains the existing rule for one SQL element. The new `expression_list` recognizes exactly the same zero-or-more sequence as `free_form_sql:*`:

```nearley
expression_list -> null
expression_list -> expression_list free_form_sql
```

Each append instead creates an immutable `{ previous, value }` node. Consequently, `expressions_or_clauses -> expression_list clause:*` preserves the original grammar: free-form elements followed by structured clauses. Clause materialization is deferred because eagerly converting every intermediate parser candidate would reintroduce quadratic copying.

## Benchmarks

Apple M4 Max, Node.js 24.12.0. Values show median formatting time and whole-process peak RSS. Stress processes used a 384 MB V8 old-space limit.

| Input | Before | After |
| --- | --- | --- |
| Simple 9 KB query | 4.19 ms / 108 MB | 4.08 ms / 111 MB |
| 1,000-value `IN` | 21.35 ms / 190 MB | 13.07 ms / 90 MB |
| 3,000-value `IN` | 107.25 ms / 444 MB | 34.47 ms / 109 MB |
| 10,000-value `IN` | Out of memory | 97.98 ms / 157 MB |
| 5,000-condition `OR` chain | Out of memory | 104.61 ms / 157 MB |
| 10,000-column `SELECT` | Out of memory | 110.04 ms / 229 MB |

## Validation

- `pnpm run grammar`
- `pnpm run ts:check`
- `pnpm run pretty:check`
- `pnpm run lint`
- `pnpm run build`
- `pnpm exec jest --runInBand --silent --coverage=false`: 27 suites, 5,841 tests, and 63 snapshots passed.
- Re-enabled issue #840 performance regression passed.
- 1,000 generated differential SQL cases preserved existing output and rejection behavior.

Closes #840.

### DIFF
--- a/src/parser/grammar.ne
+++ b/src/parser/grammar.ne
@@ -34,6 +34,20 @@ interface CommentAttachments {
   trailing?: CommentNode[];
 }
 
+interface ExpressionList {
+  previous?: ExpressionList;
+  value: AstNode;
+}
+
+const materializeExpressionList = (expressions?: ExpressionList): AstNode[] => {
+  const result: AstNode[] = [];
+  for (let current = expressions; current; current = current.previous) {
+    result.push(current.value);
+  }
+  result.reverse();
+  return result;
+};
+
 const addComments = (node: AstNode, { leading, trailing }: CommentAttachments): AstNode => {
   if (leading?.length) {
     node = { ...node, leadingComments: leading };
@@ -83,14 +97,24 @@ main -> statement:* {%
 statement -> expressions_or_clauses (%DELIMITER | %EOF) {%
   ([children, [delimiter]]) => ({
     type: NodeType.statement,
-    children,
+    children: materializeExpressionList(children),
     hasSemicolon: delimiter.type === TokenType.DELIMITER,
   })
 %}
 
 # To avoid ambiguity, plain expressions can only come before clauses
-expressions_or_clauses -> free_form_sql:* clause:* {%
-  ([expressions, clauses]) => [...expressions, ...clauses]
+# For performance, keep both in one persistent list until the surrounding rule is complete.
+expressions_or_clauses -> expression_list {% id %}
+expressions_or_clauses -> expressions_or_clauses clause {%
+  ([previous, value]) => ({ previous, value })
+%}
+
+# Avoid free_form_sql:*: Nearley implements it with array concatenation, which
+# copies every shared prefix and makes long expression lists quadratic.
+# Nearley's null matches no input; undefined represents an empty linked list.
+expression_list -> null {% () => undefined %}
+expression_list -> expression_list free_form_sql {%
+  ([previous, value]) => ({ previous, value })
 %}
 
 clause ->
@@ -119,11 +143,14 @@ limit_clause -> %LIMIT _ expression_chain_ (%COMMA free_form_sql:+):? {%
   }
 %}
 
-select_clause -> %RESERVED_SELECT (all_columns_asterisk free_form_sql:* | asteriskless_free_form_sql free_form_sql:*) {%
+select_clause -> %RESERVED_SELECT (all_columns_asterisk expression_list | asteriskless_free_form_sql expression_list) {%
   ([nameToken, [exp, expressions]]) => ({
     type: NodeType.clause,
     nameKw: toKeywordNode(nameToken),
-    children: [exp, ...expressions],
+    // Nearley completes every prefix; only materialize the surviving clause's children.
+    get children(): AstNode[] {
+      return [exp, ...materializeExpressionList(expressions)];
+    },
   })
 %}
 select_clause -> %RESERVED_SELECT {%
@@ -138,19 +165,23 @@ all_columns_asterisk -> %ASTERISK {%
   () => ({ type: NodeType.all_columns_asterisk })
 %}
 
-other_clause -> %RESERVED_CLAUSE free_form_sql:* {%
+other_clause -> %RESERVED_CLAUSE expression_list {%
   ([nameToken, children]) => ({
     type: NodeType.clause,
     nameKw: toKeywordNode(nameToken),
-    children,
+    get children(): AstNode[] {
+      return materializeExpressionList(children);
+    },
   })
 %}
 
-set_operation -> %RESERVED_SET_OPERATION free_form_sql:* {%
+set_operation -> %RESERVED_SET_OPERATION expression_list {%
   ([nameToken, children]) => ({
     type: NodeType.set_operation,
     nameKw: toKeywordNode(nameToken),
-    children,
+    get children(): AstNode[] {
+      return materializeExpressionList(children);
+    },
   })
 %}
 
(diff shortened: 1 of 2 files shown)


src/parser/grammar.ne
```diff
@@ -34,6 +34,28 @@ interface CommentAttachments {
...
 
+interface ExpressionList {
+  previous: ExpressionList | null;
+  value: AstNode;
+}
+
+interface PendingExpressions {
+  expressions: ExpressionList | null;
+  clauses: AstNode[];
+}
```
nene: The name `PendingExpressions` makes no sense to me. It's the return value of the `expressions_or_clauses` rule.

src/parser/grammar.ne
```diff
@@ -34,6 +34,28 @@ interface CommentAttachments {
...
+  const result: AstNode[] = [];
+  for (let current = expressions; current !== null; current = current.previous) {
+    result.push(current.value);
+  }
+  result.reverse();
+  return result;
+};
+
+const materializeExpressions = ({ expressions, clauses }: PendingExpressions): AstNode[] =>
+  materializeExpressionList(expressions).concat(clauses);
```
nene: The name `materializeExpressions` is similarly an odd one. It doesn't align with the input data type name, nor does it align with the name of the rule `expressions_or_clauses`, together with which it is used.

src/parser/grammar.ne
```diff
@@ -119,11 +146,13 @@ limit_clause -> %LIMIT _ expression_chain_ (%COMMA free_form_sql:+):? {%
...
 %}
 
-select_clause -> %RESERVED_SELECT (all_columns_asterisk free_form_sql:* | asteriskless_free_form_sql free_form_sql:*) {%
+select_clause -> %RESERVED_SELECT (all_columns_asterisk expression_list | asteriskless_free_form_sql expression_list) {%
   ([nameToken, [exp, expressions]]) => ({
     type: NodeType.clause,
     nameKw: toKeywordNode(nameToken),
-    children: [exp, ...expressions],
+    get children(): AstNode[] {
+      return [exp, ...materializeExpressionList(expressions)];
```
nene, line 152: What's the purpose of this wrapping of `children` property into a getter function?

It looks like some an additional performance optimization. But not sure it's really adding anything. At least when I removed these, the code still seemed to perform similarly.

If this is a separate unrelated optimization, then please make a separate PR with it. If it's tightly related to the current optimization and really needed for it, please provide some explanation.

src/parser/grammar.ne
```diff
@@ -83,14 +105,19 @@ main -> statement:* {%
...
 
 # To avoid ambiguity, plain expressions can only come before clauses
-expressions_or_clauses -> free_form_sql:* clause:* {%
-  ([expressions, clauses]) => [...expressions, ...clauses]
+expressions_or_clauses -> expression_list clause:* {%
+  ([expressions, clauses]) => ({ expressions, clauses })
+%}
+
+expression_list -> null {% () => null %}
+expression_list -> expression_list free_form_sql {%
```
nene: This definitely needs some documentation, as to why we're doing it this way. Otherwise somebody reading this code in the future will just refactor this into `expression_list:*`

src/parser/grammar.ne
```diff
@@ -83,14 +105,19 @@ main -> statement:* {%
...
     hasSemicolon: delimiter.type === TokenType.DELIMITER,
   })
 %}
 
 # To avoid ambiguity, plain expressions can only come before clauses
-expressions_or_clauses -> free_form_sql:* clause:* {%
-  ([expressions, clauses]) => [...expressions, ...clauses]
+expressions_or_clauses -> expression_list clause:* {%
+  ([expressions, clauses]) => ({ expressions, clauses })
+%}
```
nene: This also seems to be an additional optimization. Is this really needed? It doesn't seem to have much of an effect on the overall performance.

Again, I would prefer leaving this out and just doing concentrating on the core optimization proposed in this PR.

src/parser/grammar.ne
```diff
@@ -34,6 +34,28 @@ interface CommentAttachments {
   trailing?: CommentNode[];
 }
 
+interface ExpressionList {
+  previous: ExpressionList | null;
```
nene: Note: Generally in SQL Formatter codebase we avoid the use of `null`, instead preferring `undefined`.

Like in here, better to use:
```suggestion
  previous?: ExpressionList;
```

nene (changes requested):
Thanks for the PR. The fix seems to work, but the code is somewhat cryptic and seems to contain various extra stuff that's not really part of the core optimization proposed in this PR.

src/parser/grammar.ne
arc-oai: Yeah, it was "Pending" because they hadn't been materialized yet, but agreed its confusing (more of an optimization detail), will try and clean this up

src/parser/grammar.ne
arc-oai: Yeah this does help in cases like `SELECT  column_1, column_2, column_3, ..., column_1000 FROM my_table` or `SELECT * FROM my_table WHERE id = 1 OR id = 2 OR id = 3 ... OR id = 1000;`. Nearley creates an intermediate clause for every prefix, so eagerly constructing children repeatedly copies the entire prefix and makes this quadratic.

The test from https://github.com/sql-formatter-org/sql-formatter/commit/eb187feaf30239e64bcfeacc603b4049fc7df4de mostly demonstrates this case. It does "pass" without this change because peak/final memory is lowered  (by only the linkedlist change), but it still takes ~20x longer without this (using quadratic _total_ memory, even if not _peak_ memory)

I could split this into a separate follow-up PR if that helps, was just broadly tackling the several related issues of "quadratic costs"

src/parser/grammar.ne
arc-oai: Related to the [other comment](https://github.com/sql-formatter-org/sql-formatter/pull/963#discussion_r3835993274), but this is needed for expressions inside parentheses, such as `WHERE id IN (1, 2, ..., 1000)` or `WHERE (id = 1 OR id = 2 OR ... id = 1000)` (I can add an extra test to cover this related scenario, and split into a followup if that makes more sense)

src/parser/grammar.ne
nene: OK. Thanks for the explanation. Makes sense now.

nene (approved):
Thanks for the cleanup and clarifications.

PR #964, opened 2026-08-18
Do not glue a property-access dot onto an integer literal

dngr2:
`1 . x` formats to `1.x`. In MySQL/MariaDB/TiDB `1.` re-lexes as a number literal,
which swallows the following property-access operator, so the formatted output
parses to a different expression than the input -- formatting isn't idempotent.

Insert a space when a `.` would attach directly to a bare integer literal, the same
way the existing guard keeps `- -` from becoming `--`. Identifiers that merely end
in a digit (`t1.x`) and decimals are unaffected, since the guard only triggers on a
pure integer literal.

Added a mysql test that formats `1 . x` and checks the result formats back to itself.


nene:
Could you provide a full executable example of the code. When I simply try running the following in MySQL:

```sql
SELECT 1 . /*x*/ 5e
```

I just get an error.

Is the number `1` a table name or alias? I tried that, but wasn't able to use plain number as a table name. So it must be something else, but I have no idea.

dngr2:
You're right that it isn't runnable MySQL — I should have led with that. It's not a query, it's a minimal token sequence. The formatter tokenizes rather than validates, so it still lays it out, and the point is only that its own output isn't stable.

On 15.8.2, formatting it twice gives two different results:

```js
const { format } = require('sql-formatter');
const a = format('SELECT 1 . /*x*/ 5e', { language: 'mysql' });
const b = format(a, { language: 'mysql' });
// a === "SELECT\n  1./*x*/ 5e"
// b === "SELECT\n  1. /*x*/ 5e"   <- a space appears after "1."
console.log(a === b); // false
```

First pass glues the `.` onto the `1`; second pass reads `1.` as a number literal and re-spaces it. That double-format divergence is the real symptom.

I overstated the description: the plain `1 . x` and `1 . a` cases are actually idempotent (they settle to `1.x` / `1.a`), so "swallows the property-access operator" was wrong. The honest scope is just that a `.` glued onto a bare integer literal isn't a stable layout.

If you don't think that edge earns a guard, feel free to close — no argument here. If you do, I'll rewrite the description and the test around the double-format case above instead of the misleading framing.

nene:
Thanks for the quick explanation. In that case I'll close this PR.

There are enough problems with the formatter not handling valid SQL. No need to complicate things further with trying to handle invalid SQL.

PR #965, opened 2026-08-22
Remove coderabbit

nene:
We have been testing out Coderabbit. My conclusion is that it hasn't really provided any real help. On the contrary, I find these Coderabbit comments annoying: they just add noise. It's like in every pull request some developer jumps in to report: hey! I'm here, I don't have anything useful to add, but I say it anyway.

On the other hand, the real problem I'm having with pull request is that the majority of them are written by AI in the first place. Like currently there's 5 pull requests open, and I have a strong suspicion that they're all AI-written. Of course it's hard to say so for sure, but I very much feel the AI-written vibe. Like I have written some feedbeck to all of them, and then it's simply been silence from the author of the PR. Like in one case I wrote a comment about how that whole 20-line PR could be achieved with just a single line, and then the author just marked it as resolved, without doing anything.

Given that the problem is in noise created by AI, I think adding more AI noise from our side isn't really helping.

I disabled Coderabbit for now. Will likely completely uninstall it later.

@karlhorky


karlhorky:
@nene ok, thanks for giving it a shot!

Wonder if [Greptile](https://www.greptile.com/) or any others would be good enough in future.

nene:
I think in the SQL Formatter repository we don't really have the problem that these tools are targeting.

Issue #578, opened 2023-03-12
[FORMATTING] Modulo Operator Formatting issue


Syntax-Read3r:
**Input data**

Which SQL and options did you provide as input?

SELECT title, released_year
FROM books
where
    released_year >= 2000
    AND (released_year % 2) != 0;

Returned Error
Unable to format SQL: Error: Parse error: Unexpected "% 2) != 0;" at line 87 column 24

- How are you calling / using the library?
- What SQL language(s) does this apply to? MySQL
- Which SQL Formatter version are you using? MySQL v8

nene:
See: https://github.com/sql-formatter-org/sql-formatter#parse-error-unexpected--at-line-

Syntax-Read3r:
First and foremost thanks for creating this extension. 
Prior to raising the bug issue, I had attempted multiple things to resolve it
such as the code below. The issue I ran into was, the problem was fixed BUT
your extension was no longer be called upon by the IDE.

`	
"[sql]": {
		"editor.defaultFormatter": "mtxr.sqltools"
		//	"editor.defaultFormatter": "inferrinizzard.prettier-sql-vscode"
	},
	"Prettier-SQL.SQLFlavourOverride": "sql",

`

I don't know what the issue, is. If you can spot it, That would be great.

nene:
Sorry, I don't quite understand your description of the issue.

What I can spot is that you're using `"Prettier-SQL.SQLFlavourOverride": "sql"` but you really should set this to MySQL: `"Prettier-SQL.SQLFlavourOverride": "mysql"`, to be able to properly parse MySQL code.

nene:
You can also try to format your code here: https://sql-formatter-org.github.io/sql-formatter/

If the problem is also present there, it's an issue in the core library (though, make sure to also choose the proper SQL dialect first). If not, then it's a problem in the VSCode extension or its configuration.

Syntax-Read3r:
I will tell you something. I don't know how I fixed the bug. But it's now fixed. It process involved a lot of uninstalling extensions and reinstalling them. Thanks for the help.

PR #971, opened 2026-09-07
fix(singlestoredb): support % as the modulo operator

MaxFreedomPollard:
Formatting a SingleStoreDB query that uses `%` throws a parse error instead of returning SQL. `MOD(a, b)` and `a MOD b` format fine, but `a % b` does not, even though SingleStore documents all three as equivalent ([MOD](https://docs.singlestore.com/cloud/reference/sql-reference/aggregate-functions/mod/) gives `expression1 % expression2` and the example `SELECT 25 % 0;`).

```js
format('SELECT 25 % 7', { language: 'singlestoredb' });
// Parse error: Unexpected "% 7" at line 1 column 11.
// SQL dialect used: "singlestoredb".
```

This is the same symptom #578 reported against MySQL, where `(released_year % 2) != 0` failed to format. MySQL was fixed; SingleStoreDB still fails on the identical query.

## Root cause

`src/lexer/Tokenizer.ts` builds the `OPERATOR` rule from a fixed standard set (`+ - / > < = <> <= >= !=`) plus whatever the dialect puts in `cfg.operators`. `%` is not in the standard set, and the `operators` array at `src/languages/singlestoredb/singlestoredb.formatter.ts:267` omits it, so nothing matches the character and `TokenizerEngine` raises the parse error.

Every MySQL-family sibling lists `'%'` first in that same array: `src/languages/mysql/mysql.formatter.ts`, `src/languages/mariadb/mariadb.formatter.ts` and `src/languages/tidb/tidb.formatter.ts`. The SingleStoreDB list, which otherwise tracks them, just dropped it.

## Fix

Add `'%'` to `singlestoredb.formatter.ts`, in the same leading position the siblings use.

`regexFactory.operator` sorts the alternatives by length descending, so SingleStore's three-character `::%` conversion path-operator still matches before the bare one-character `%`. The existing `formats '::%' conversion path-operator without spaces` test in `test/singlestoredb.test.ts` pins that and still passes.

I checked the other 19 dialect files under `src/languages/` for the same omission and found none: every dialect whose documentation gives a `%` operator already lists it. Standard `sql` and `bigquery` are correct to leave it out, since GoogleSQL has `MOD(X, Y)` and no `%` operator. `db2i` also leaves it out, and I left that as it is.

## Tests

`'%'` added to the `supportsOperators(...)` list in `test/singlestoredb.test.ts`, which covers both default spacing (`foo % bar % zap`) and `denseOperators: true` (`foo%bar`). Prettier reflowed the array onto one entry per line, matching how `test/mysql.test.ts` already writes it.

Both new cases fail on master with `Parse error: Unexpected "% bar %zap"` and pass with the fix. `pnpm run pretty:check`, `pnpm run lint`, `pnpm run ts:check`, `pnpm run test` and `pnpm run build` are all green.


nene:
Thanks for the fix.

But please, don't write a wall of text to describe a trivial single-line fix like this one. Or don't use AI to to generate such an epic poem. I really don't know what I should do with it. I suppose you didn't read it. I also didn't read it. For who was it necessary then?

Perhaps if you had not written such a long description I would have even reviewed you change earlier.

MaxFreedomPollard:
Ok, fair

Issue #801, opened 2024-11-14
[FORMATTING] Reserved words can't be used as aliases


pgiraud:
**Input data**

```sql
SELECT set.foo FROM settings AS set
```

Here's a more complex example (the error is a bit different):

```sql
SELECT id AS srvid,
  CASE WHEN id = $1 THEN $2
  ELSE
     COALESCE(alias, hostname || $3 || port)
  END AS alias,
  CASE WHEN id = $4 THEN $5 ELSE hostname END as hostname,
  CASE WHEN id = $6 THEN $7 ELSE port END AS port,
  CASE WHEN id = $8 THEN set.setting
      ELSE s.version::text
  END AS version
FROM public.powa_servers s
LEFT JOIN pg_settings set ON set.name = $9 AND s.id = $10
```

**Usage**

- How are you calling / using the library?
```
import { formatDialect, postgresql } from "sql-formatter";
formatDialect(value, { dialect: postgresql })
```
Tested in the online demo as well.

- What SQL language(s) does this apply to?
PostgreSQL, but the first example fails with any dialect with the same error

- Which SQL Formatter version are you using?
15.4.6

nene:
Thanks for reporting.

Although this error occurs with other dialects, it's more of a PostgreSQL-specific issue as most SQL dialects don't allow reserved keywords to be used as aliases. PostgreSQL also only allows some, like the following won't work:

```sql
SELECT select.foo FROM settings AS select;
```

However this syntax is valid in DB2, where one can even write:

```sql
SELECT select FROM from;
```

Isn't SQL a wonderful language :D

Usually I would recommend checking out [prettier-plugin-sql-cst](https://github.com/nene/prettier-plugin-sql-cst), which does a much better job at formatting PostgreSQL than this library, but turns out that [the same issue](https://github.com/nene/sql-parser-cst/issues/95) is also present in there. Though my hopes of fixing this issue in there are much higher than fixing it in here.

nene:
This has now been improved in 15.9.0 release. Thing should be better now if you use `AS` keyword to declare aliases. Without the `AS`, there will still be problems though.

PR #972, opened 2026-09-14
Treat a reserved word after AS as an alias

cpruijsen:
Cut down to the `AS`-only rule, per the discussion below. The previous version tracked statement
position to also fix the bare alias form; that machinery is gone.

A reserved word used as an alias breaks the formatter. In `SELECT id AS set FROM tbl` the `set`
token is classified as the start of a `SET` clause, so a clause appears where the parser does not
expect one.

Directly after `AS`, a reserved word can only be an alias name. `keywordAliasAfterAs` converts it to
`IDENTIFIER`, and joins the existing `disambiguateTokens` pipeline alongside the five neighbour-local
passes, which are unchanged.

Only the token types that cannot legitimately follow `AS` are converted. `RESERVED_CLAUSE` is
excluded apart from `SET`, so `CREATE TABLE t AS SELECT ...`, `... AS VALUES (...)` and
`PREPARE foo AS UPDATE ...` keep working. Data types are excluded for the same reason.

Not covered: the bare alias form, as in `LEFT JOIN pg_settings set ON set.name = $9`. A bare alias
needs to know where it sits in the statement, which is the positional tracking this version drops.
Anyone hitting that case can add `AS`.

Tests: the `AS` alias in the shared suite, the two PostgreSQL forms from the issue, and a regression
check that `SET` still starts a clause after an alias, as in `UPDATE tbl AS set SET x = 1`.

Refs #801


nene:
Thanks for the pull request.

I have to take some time to digest it, as it's quite a mouthful.

I totally understand why this change might be quite a large one. Well.. I personally had put that one aside as an unfixable one with the current architecture.

An initial glance at the code tells me that it contains lots of code to tackle various specific cases. That in turn leads me to the initial conclusion that this whole thing is likely very fragile. That is, it's not some sort of generic fix for this whole problem, but rather it's a set of patches for the problem in various scenarios that occurred to the author, with lots of other scenarios to be patched up in the future.

In general I'm reluctant to take aboard such a large piece of code which in its essence is one big hack. Don't take this personally, I very much believe that it might not really be possible to fix it in any non-hack way. I'm OK with small hacks to correct some shortcomings in this fundamentally flawed SQL Formatter library. But I'd like to avoid big hacks that can become a burden to maintain. Instead I like directing people into using [prettier-plugin-sql-cst](https://github.com/nene/prettier-plugin-sql-cst), which is built on top of a proper SQL parser, or to use conditional comments to disable the formatter for parts of code it can't handle.

Anyway, these are just my very initial thoughts. I'm planning to have a more proper look at this, and then give my full verdict.

cpruijsen:
Yeah, that's a fair read. It is a pile of positional special cases, and each new
dialect shape probably adds one.

One thing on the size: an `AS`-only rule would be a fraction of this, but it only
covers the first example in #801. The second one,
`LEFT JOIN pg_settings set ON set.name = $9`, has no `AS`, and a bare alias needs
to know where it is in the statement. That's where the context tracking came from.

If you want it smaller I'll cut it to `AS` only and leave the bare-alias form
unfixed, around 30 lines. And if neither version belongs here, just say so and
I'll close it. No need to read the whole thing first.

nene:
Hey, that sounds like a great plan. I think fixing the `AS`-aliases would solve a major chunk of the problem. Plus it will give an easy way for people to tweak their code to make it passable for the formatter.

cpruijsen:
Cut down, pushed as bcd3a42. The PR is now 82 added lines across 3 files, down from 275, and the
title and description are rewritten to match.

The positional tracking is gone entirely. What is left is `keywordAliasAfterAs`, a neighbour-local
pass that sits in the existing `disambiguateTokens` pipeline next to the five already there:

```ts
const keywordAliasAfterAs = (token: Token, i: number, tokens: Token[]): Token => {
  if (canBeAliasAfterAs(token)) {
    const prevToken = prevNonCommentToken(tokens, i);
    if (prevToken && isAsKeyword(prevToken)) {
      return { ...token, type: TokenType.IDENTIFIER, text: token.raw };
    }
  }
  return token;
};
```

No paren stack, no clause tracking, no new helpers. The rest of the 45 source lines is
`canBeAliasAfterAs`, which is a list of the token types that cannot legitimately follow `AS`, plus
the doc comments. `RESERVED_CLAUSE` is excluded apart from `SET`, so `CREATE TABLE t AS SELECT ...`,
`... AS VALUES (...)` and `PREPARE foo AS UPDATE ...` are untouched.

The bare alias form is not covered, as agreed. `LEFT JOIN pg_settings set ON set.name = $9` still
formats badly, and the description says so rather than leaving people to find out. Adding `AS` is
the workaround.

Full suite is green, 5866 passing, plus eslint and prettier. New tests are the `AS` alias in the
shared suite, the two PostgreSQL forms from the issue, and a check that `SET` still starts a clause
after an alias, as in `UPDATE tbl AS set SET x = 1`.

Happy to drop the `canBeAliasAfterAs` list to just `SET` if you would rather the first version of
this be as small as possible.

src/lexer/disambiguateTokens.ts
```diff
@@ -39,6 +43,47 @@ const propertyNameKeywordToIdent = (token: Token, i: number, tokens: Token[]): T
...
+    if (prevToken && isAsKeyword(prevToken)) {
+      return { ...token, type: TokenType.IDENTIFIER, text: token.raw };
+    }
+  }
+  return token;
+};
+
+const isAsKeyword = (token: Token): boolean =>
+  (token.type === TokenType.RESERVED_KEYWORD || token.type === TokenType.RESERVED_KEYWORD_PHRASE) &&
+  token.text === 'AS';
```
nene: Interesting. It shouldn't be the case that `AS` is classified as `RESERVED_KEYWORD_PHRASE` (it's not much of a phrase, it's just a single keyword).

Can you point me at a case where `AS` occurs in a list like:

```ts
const reservedKeywordPhrases = expandPhrases([
  '{ROWS | RANGE | GROUPS} BETWEEN',
  'SIMILAR TO',
  'IS [NOT] DISTINCT FROM',
]);
```

src/lexer/disambiguateTokens.ts
```diff
@@ -12,13 +12,17 @@ import { isReserved, Token, TokenType } from './token.js';
  * When IDENTIFIER or RESERVED_DATA_TYPE token is followed by "["
  * converts it to ARRAY_IDENTIFIER or ARRAY_KEYWORD accordingly.
  *
+ * Converts a reserved word directly after AS to IDENTIFIER, as it can only
+ * be an alias name there.
```
nene: This comment is somewhat misleading and the code actually accounts for cases where e.g. `SELECT` follows `AS`.

test/postgresql.test.ts
```diff
@@ -324,4 +324,30 @@ describe('PostgreSqlFormatter', () => {
...
+    `);
+    expect(format(`SELECT * FROM pg_settings AS set WHERE set.name = $9;`)).toBe(dedent`
+      SELECT
+        *
+      FROM
+        pg_settings AS set
+      WHERE
+        set.name = $9;
+    `);
+  });
```
nene: For some reason we have a test for table aliases for PostgreSQL only and a test for column aliases for other SQL dialect. But really, the table aliases are supported in all dialects. Also, for some reason we test two cases `SELECT .. FROM` and `SELECT .. FROM .. WHERE`. That latter one is really unnecessary as the `set.name` disambiguation is handled separately and not part of issue #801.

I would just merge this into a one simple test, which covers both a column alias and a table alias and is executed for all SQL dialects.

test/postgresql.test.ts
```diff
@@ -324,4 +324,30 @@ describe('PostgreSqlFormatter', () => {
...
+
+  it('keeps SET as a clause after a reserved-word alias', () => {
+    expect(format(`UPDATE tbl AS set SET x = 1;`)).toBe(dedent`
+      UPDATE tbl AS set
+      SET
+        x = 1;
+    `);
+  });
```
nene: This test should also not be PostgreSQL-specific.

nene (commented):
Thanks. Looks good in general.

Added some notes, mostly about the tests.

test/postgresql.test.ts
nene: Unfortunately can't move this to `behavesLikeSqlFormatter()` because SparkSQL annoyingly doesn't have an UPDATE statement. But we have a separate `supportsUpdate()` function we can add this to.

cpruijsen:
No, there isn't one. I checked every dialect's `reservedKeywordPhrases` and `reservedDataTypePhrases`: none declares a bare `AS`. The only AS-bearing phrases are postgresql's `AS MATERIALIZED` and `GENERATED {ALWAYS | BY DEFAULT} AS IDENTITY`, plus redshift's `NULL AS`, and those tokenize with the whole phrase as the token text, so `text === 'AS'` could never match. Dropped the branch.

The other four are in 55c68bb:

- Reworded the `disambiguateTokens` doc comment. It claimed every reserved word after `AS` becomes an identifier, which is not what the code does.
- Merged the two PostgreSQL tests into the shared one, now `SELECT id AS set FROM tbl AS set;` so a column alias and a table alias are covered in one statement, for every dialect. Dropped the `WHERE` case.
- Moved the SET-as-a-clause test to `supportsUpdate()`.

`test/postgresql.test.ts` is back to its original content, so the PR is down to 3 files and 65 added lines. 5883 tests green.

nene:
Thanks.

nene:
This is now out in 15.9.0

